PR #142, opened 2020-02-19
feat: add a new GoogleFont for ENSs

nicosantangelo:
I don't think this is enough, but it's a start. Not really sure if we should close the issue

#140


cazala:
Closes in favor of #149

PR #152, opened 2020-02-20
feat: link to explorer

eordano:


eordano:
@nicosantangelo wdyt?

nicosantangelo:
I like it! we're chatting with mat to see if we can style it a bit before merging

PR #161, opened 2020-02-27
feat: filters

### LINKED ISSUE #28: Market page filters (dropdown)
Depends on #26

### LINKED ISSUE #134: Feedback from community member
Screen Shot 2020-02-17 at 21 38 53

### LINKED ISSUE #154: Add total assets to the NFTList component
We can send two queries when we fetch nfts,  the second one without a limit (or limited at 1000) that only brings back the ids, to count the results. We can then display this number in the nft list component

cazala:
Closes #28
Closes #154 
Closes #134

This PR adds filters for wearables (rarity and gender)and the total results for each query also it adds missing translations (like the dates)

Screen Shot 2020-02-27 at 4 09 17 PM


webapp/src/modules/ui/selectors.ts
```diff
@@ -62,6 +63,36 @@ export const getUIOnlyOnSale = createSelector<
   return onlyOnSale === null ? onlyOnSale : onlyOnSale === 'true'
 })
 
+export const getUIWearableRarities = createSelector<
+  RootState,
+  string,
+  WearableRarity[]
+>(getSearch, search => {
+  const rarities = new URLSearchParams(search).get('rarities')
```
nicosantangelo: Maybe we can move the `as WearableRarity[]` to this line to avoid repeating it below

webapp/src/modules/ui/selectors.ts
```diff
@@ -62,6 +63,36 @@ export const getUIOnlyOnSale = createSelector<
...
+          Object.values(WearableRarity).includes(rarity as WearableRarity)
+        ) as WearableRarity[])
+})
+
+export const getUIWearableGenders = createSelector<
+  RootState,
+  string,
+  WearableGender[]
+>(getSearch, search => {
+  const genders = new URLSearchParams(search).get('genders')
```
nicosantangelo: Maybe we can move the `as WearableGender` to this line to avoid repeating it below

webapp/src/modules/ui/selectors.ts
```diff
@@ -62,6 +63,36 @@ export const getUIOnlyOnSale = createSelector<
...
+export const getUIWearableRarities = createSelector<
+  RootState,
+  string,
+  WearableRarity[]
+>(getSearch, search => {
+  const rarities = new URLSearchParams(search).get('rarities')
+  return rarities === null
+    ? []
+    : (rarities
+        .split('_')
```
nicosantangelo: You think it'd be a good idea to move this to it's own function? maybe to `routing/search.ts`, to keep the separator (`_` in this case) abstracted in one place

We can also move it to it's own `filter` concept/module and move both this and `params.set('rarities', options.wearableRarities.join('_'))` there

cazala:
@nicosantangelo thanks for the review! feedback addressed

PR #184, opened 2020-05-20
feat: rename swanky to rare

nicosantangelo:


webapp/src/modules/translation/locales/zh.json
```diff
@@ -135,7 +135,7 @@
       "mythic": "神话",
       "legendary": "传奇",
       "epic": "史诗",
-      "swanky": "时髦",
+      "rare": "时髦",
```
nachomazzara, line 138: This one is made by someone of the community or should we need to use google translator?

For `rare` is = 罕见 which is pretty similar (if not the same) as `uncommon`. The other suggestion is `稀有`

webapp/src/modules/translation/locales/es.json
```diff
@@ -144,7 +144,7 @@
       "mythic": "Para algunos pocos: máximo 10",
       "legendary": "Ediciión limitada: máximo 100",
```
nachomazzara, line 145: Not for this PR, but typo here.

PR #216, opened 2020-08-16
Fix frontend typo in readme

mikkhimself:


nicosantangelo:
Good catch, thank you for the PR!

nicosantangelo:
Good catch, thank you for the PR!

PR #231, opened 2020-10-18
fix: broken wearables

cazala:
Some wearables were indexed wrong (lots of null data) and it's breaking the fronted. This PR mitigates the issue until we get the subgraph fixed. 

Screen Shot 2020-10-17 at 9 01 33 PM


nicosantangelo (approved):
LGTM

PR #232, opened 2020-10-23
feat: add a unique id for bids

nicosantangelo:


indexer/src/handlers/bid.ts
```diff
@@ -1,11 +1,12 @@
-import { BigInt, Address } from '@graphprotocol/graph-ts'
+import { log, BigInt, Address } from '@graphprotocol/graph-ts'
```
nachomazzara, line 1: Are we using `log` in this file?

nachomazzara (approved):
LGTM!  

Let's merge it on Monday

PR #238, opened 2020-11-27
feat: updated at when bid/order changes

nicosantangelo:


indexer/src/handlers/marketplace.ts
```diff
@@ -73,6 +74,7 @@ export function handleOrderSuccessful(event: OrderSuccessful): void {
     order.save()
 
     nft.owner = event.params.buyer.toHex()
+    nft.updatedAt = event.block.timestamp
     nft = updateNFTOrderProperties(nft!, order!)
```
nachomazzara, line 78: Maybe we can move the update of `updatedAt` to the `updateNFTOrderProperties`. WDYT?

nachomazzara (approved):

PR #277, opened 2021-05-06
Refactor: In scripts simplify fs usage

theqp:


nicosantangelo:
This works! but we prefer keeping the I/O methods async to free cpu in this case. The difference is almost negligible but’s there.

What we could do however, is use the new Node 11 (and onwards) [promisified fs](https://nodejs.org/api/fs.html#promises-api) methods!
Would you like to make that change instead?

Thanks for the PR!

PR #287, opened 2021-05-17
feat: use catayst for images

nachomazzara:


indexer/src/modules/network/index.ts
```diff
@@ -0,0 +1,6 @@
+import { dataSource } from '@graphprotocol/graph-ts'
+
+export function getNetwork(): string {
+  let network = dataSource.network()
+  return network == "mainnet" ? "ethereum" : network
+}
```
nicosantangelo, line 6: I feel like this is more a localized thing of obtaining the URN right? I have two suggestions, tell me what you think but still the code is 💯 so I'll approve and you can skip.
- naming this `getURNNetwork()`, so it more closely reflects what it's doing
- Moving this:
```ts
  let baseURI = getCatalystBase()
  let urn = 'urn:decentraland:' + getNetwork() + ':collections-v1:' + wearable.collection + ':' + wearable.representationId
```
to it's own method `getURN()` and doing this logic there.
- Doing both things

What do you think?

nicosantangelo (approved):
🆗

indexer/src/modules/network/index.ts
nachomazzara: Good call!

Yes, I thought about doing it that way, let me clarify why:

- Why not `getURNNetwork`? I didn't realize that adding the `URN` will help to understand its usage. I can update it.
- Why not `getURN`? because as we are going to support only l1 wearables that method should not be reusable but I can move it anyway.

Thanks

nicosantangelo (approved):

PR #315, opened 2021-05-31
fix: stuff in activity page

### LINKED ISSUE #312: MANA symbol oddities
Some problems I found using the marketplace:

- The Ethereum MANA symbol seems to be a **tiny bit** missaligned on the activity page, the Icon on each TX:

- In the Settings page, you can't really tell which authorization is for Polygon and which for Ethereum. I think is most noticiable in the MANA section for buying

cazala:
Fixes #312 
Also fixes the transaction links (uses the chainId to pick the right block explorer)
Also fixes wrong properties (using `contract.name` instead of `contract.address` to link to block explorer)
Also fixes the `networks.undefined` error in console
Also bumps `decentraland-dapps` which includes fixes in `decentraland-transactions` (new polygon contracts)

### DIFF
--- a/webapp/src/components/ActivityPage/Transaction/TransactionDetail/TransactionDetail.css
+++ b/webapp/src/components/ActivityPage/Transaction/TransactionDetail/TransactionDetail.css
@@ -39,6 +39,7 @@
   width: 41px;
   background-repeat: no-repeat;
   margin-top: -9px;
+  margin-left: 1px;
 }
 
 .TransactionDetail .left .text {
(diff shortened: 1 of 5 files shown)


nachomazzara (approved):

webapp/src/components/ActivityPage/Transaction/TransactionDetail/TransactionDetail.css
```diff
@@ -39,6 +39,7 @@
   width: 41px;
   background-repeat: no-repeat;
   margin-top: -9px;
+  margin-left: 1px;
```
nicosantangelo, line 42: ❤️

nicosantangelo (approved):

PR #317, opened 2021-06-01
feat: add toasts from dapps and handle meta transaction errors

### LINKED ISSUE #288: Show errors when the meta-tx fail or user refuses to sign the message

nicosantangelo:
Closes #288


webapp/src/modules/toast/toasts.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+import { ToastType } from 'decentraland-ui'
+import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
+
+const DISCORD_URL = process.env.REACT_APP_DISCORD_URL
```
cazala, line 5: Maybe we should add this to the `.env.example` and the `.ci` folder vars?

webapp/src/modules/toast/sagas.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function* toastSaga() {
+  yield all([baseToastSaga(), customToastSaga()])
+}
+
+function* customToastSaga() {
+  yield takeEvery(TRANSFER_NFT_FAILURE, handleNFTMetaTransactionFailure)
+  yield takeEvery(CREATE_ORDER_FAILURE, handleNFTMetaTransactionFailure)
+  yield takeEvery(EXECUTE_ORDER_FAILURE, handleNFTMetaTransactionFailure)
+  yield takeEvery(CANCEL_ORDER_FAILURE, handleNFTMetaTransactionFailure)
```
cazala, line 24: We are missing two meta transaction error actions: `GRANT_TOKEN_FAILURE` and `REVOKE_TOKEN_FAILURE`. Those should use a different handler since they don't have an `nft` property.

webapp/src/modules/toast/sagas.ts
```diff
@@ -0,0 +1,39 @@
...
+function* handleNFTMetaTransactionFailure(
+  action:
+    | TransferNFTFailureAction
+    | CreateOrderFailureAction
+    | ExecuteOrderFailureAction
+    | CancelOrderFailureAction
+) {
+  const { nft } = action.payload
+
+  if (nft.network === Network.MATIC) {
```
cazala, line 36: Should we also check and ignore errors that include `User denied message signature`? Otherwise we get these toasts when the users rejects the MetaMask prompt.

cazala (changes requested):

webapp/src/modules/toast/sagas.ts
nicosantangelo: Really good point, I think so, I should also check builder for the same thing. I'll try to come up with a way to do that but I think I'll have to compare strings

webapp/src/modules/toast/toasts.tsx
nicosantangelo: Sorry, yes!

webapp/src/modules/translation/locales/es.json
```diff
@@ -342,5 +342,11 @@
     "manage_account": "Mi Cuenta",
     "content": "Polygon MANA te permitirá hacer transacciones mas rápido y barato.{br}Puedes aprender mas sobre Polygon MANA en nuestro {blog_link}.{br}Puedes gestionar tus tokens desde la {account_link}.{br}Haz click en tu avatar, luego en {manage_account} y comienza a conquistar el Metaverso.",
     "dismiss": "Entendido!"
+  },
+  "toast": {
+    "meta_transaction_failure": {
+      "title": "Meta transacción",
```
cazala, line 348: joder!

cazala (approved):

webapp/src/modules/translation/locales/es.json
nicosantangelo: Meta transacción, saque vestimenta

PR #346, opened 2021-06-16
feat: prevent smart contract wallets

cazala:


nicosantangelo (approved):
LGTM

nicosantangelo:
This makes me think that we should move to using errors as arguments in FAILURE actions instead of strings everywhere 😓

cazala:
I converted this back into draft, because after discussing we decided that it is better to preserve the error as `string` and add a new property `code` to deal with error codes (less disruptive to current codebase).

Also, as a side note, this feat is already included in master (and prod) since it has the latest `decentraland-transactions`, the thing is the error message is not specific without the changes introduced in this PR. Later on I'll make the adjustments to this PR and remove the draft flag.

PR #361, opened 2021-07-09
feat: Part of estate

### LINKED ISSUE #355: Change the Owner of parcels that belong to Estate to say `Part of {ESTATE NAME}`
Right now, we're showing the actual owner, which is the EstateRegistry.
We should also hide other parcel actions like BID or BUY even though the contract allows them to exist, to avoid confusion

Lastly, we should add this feature to Districts as well, showing the name of it instead of the address that owns it

LautaroPetaccio:
This PR does the following:
- Replaces the parcel's owner indicator with "Part of {estate_name}" in any parcel that is part of an estate. The estate name is navigable so the users can see the detail of the estate the parcel belongs to.
- Uses the Coordinates component in the places that where not using it but were showing a similar component.
- Reworks the EstateDetails and ParcelDetails component so the name of the estates or parcels can grow without breaking the styling or shrinking other elements.
- Refactors the JumpIn badge into a component to remove repeated code.

Closes #355 

**Desktop**
part-of-desktop
**Mobile**
part-of-mobile


webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
```diff
@@ -1,21 +1,40 @@
...
-  background-size: 16px 16px;
-  width: 16px;
-  height: 16px;
-  background-repeat: no-repeat;
-  display: inline-block;
-  margin-left: 6px;
+.EstateDetail .left-title {
+  flex: 1 1 auto;
+  display: flex;
+  align-items: flex-start;
```
nicosantangelo, line 4: I think we can try changing this to `center` so the estate name is vertically aligned to the badges:

https://user-images.githubusercontent.com/692077/125319339-0dc69780-e311-11eb-9248-96f48119a124.mov

I tried it without taking every resolution into account, just a suggestion

webapp/src/components/NFTPage/EstateDetail/EstateDetail.tsx
```diff
@@ -36,28 +35,21 @@ const EstateDetail = (props: Props) => {
...
+              </Header>
+              <Badge className="estate-title-land" color="#37333d">
+                {estate.size.toLocaleString()} LAND
+              </Badge>
+              {estate.size > 0 ? (
+                <JumpIn className="estate-title-jump-in" x={x} y={y} />
+              ) : null}
+            </>
           }
+          rightClassName={'right-title'}
```
nicosantangelo, line 52: Seen this before on `<Coordinate className={'coordinates'}` on `Atlas/Popup/Popup.tsx` but didn't comment on it.

Super minor but I think it's a bit cleaner to just do `rightClassName="right-title"`, at the very least, to avoid two chars (`{}`), like [here](https://github.com/decentraland/marketplace/pull/361/files#diff-1c518eb7340846ac926b03d7c9e66bb365edd5f11289f70756bd6aaf97e6695bR62)

What do you think?

webapp/src/components/NFTPage/ParcelDetail/ParcelDetail.css
```diff
@@ -1,31 +1,50 @@
...
-  display: flex;
-  align-items: center;
+.ParcelDetail .parcel-title-name {
+  flex: 0 1 auto;
+  word-break: break-word;
+}
+
+.ParcelDetail .parcel-title-coordinate {
+  margin-left: 16px;
+  margin-top: 0px;
```
nicosantangelo, line 23: Couldn't find the use for this particular rule. If it's really not necessary this selector could be merged with `ParcelDetail .parcel-title-jump-in` below

webapp/src/components/NFTPage/ParcelDetail/ParcelDetail.tsx
```diff
@@ -37,28 +41,47 @@ const ParcelDetail = (props: Props) => {
...
+            isPartOfEstate ? (
+              <div className="estate-information">
+                <T
+                  id="nft_page.part_of_estate"
+                  values={{
+                    estate_name: (
+                      <Link
+                        to={locations.nft(nft.owner, parcel.estate!.tokenId)}
+                      >
+                        <span title={parcel.estate!.name}>
```
nicosantangelo, line 73: I think this `span` is unnecessary, as the title can be placed on the `Link` right?

nicosantangelo (changes requested):
Great work, long time coming this feature.

UI Stuff:

- I think I commented this on the code itself but it'd be great if we could check the vertical alignment of the `Part of xxx` text in regards to the Land name. I tried to exemplify this on the screenshots below, where `Part of` is aligned to the top instead of the center of the dotted lined rectangle
- 
image

vs

image

- Mobile works great for long Estate names, but it looks a bit weird for short ones. Maybe we could align it to the left too?
 
image

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
LautaroPetaccio: The problem with centering the badges is that the estate/parcel name will grow in height, having multiple lines and moving the badges down to be centered. This also moves the badges from being at the same level with the owner badge.

webapp/src/components/NFTPage/EstateDetail/EstateDetail.tsx
LautaroPetaccio: This is a typo! I do prefer using the string whenever it's possible. Thanks for noticing

webapp/src/components/NFTPage/ParcelDetail/ParcelDetail.css
LautaroPetaccio: You're right, great catch!

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
```diff
@@ -1,21 +1,40 @@
...
+  margin-left: 16px;
+  margin-top: 3px;
+  flex: 0 0 auto;
+}
+
+.EstateDetail .estate-title-jump-in {
+  flex: 0 0 auto;
+  margin-top: 3px;
+  margin-left: 16px;
 }
```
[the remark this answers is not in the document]
LautaroPetaccio: Done!

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
LautaroPetaccio: I can remove the margin-top that we're using in the badges, but I added that to making it look a bit closer to the owner badge on the right.

webapp/src/components/NFTPage/ParcelDetail/ParcelDetail.tsx
LautaroPetaccio: You're right, removed!

webapp/src/components/NFTCard/NFTCard.tsx
```diff
@@ -43,7 +43,7 @@ const NFTCard = (props: Props) => {
           ) : null}
         </Card.Header>
         <Card.Meta>{t(`networks.${nft.network.toLowerCase()}`)}</Card.Meta>
-        {parcel ? <ParcelTags nft={nft} /> : null}
+        {parcel ? <ParcelTags className="tags" nft={nft} /> : null}
```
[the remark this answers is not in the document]
LautaroPetaccio: We do! we're using the `tags` class in the `ParcelTags` component [here](https://github.com/decentraland/marketplace/pull/361/files#diff-6aa0af39deaa8828ca1e2a3ce7aaa13ce2d3600b833e650749319b7e36b36c23L10) but since it switched to using CSS modules, we need to set it here.

webapp/src/components/NFTCard/ParcelTags/ParcelTags.module.css
```diff
@@ -0,0 +1,19 @@
+.ParcelTags {
+  display: flex;
+}
+
+.coordinates {
+  font-size: 12px !important;
+  padding: 4px 8px !important;
+  border-radius: 4px !important;
+  color: white !important;
```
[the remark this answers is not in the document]
LautaroPetaccio: You're right, removed it!

webapp/src/components/NFTCard/NFTCard.tsx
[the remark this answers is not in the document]
nicosantangelo: My b 👌

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
nicosantangelo: Hm I see, never mind, it's super minor and I really didn't test it thoroughly

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
```diff
@@ -1,21 +1,34 @@
...
+.EstateDetail .estate-title-badge {
+  margin-left: 16px;
+  margin-top: 3px;
+  flex: 0 0 auto;
 }
 
 @media (max-width: 768px) {
-  .EstateDetail .jump-in-badge {
-    display: none;
+  .EstateDetail .estate-title-jump-in {
```
nicosantangelo, line 26: I think we need to re-add this class even if it's just for this selector to be applied, maybe something like: `<JumpIn className="estate-title-badge estate-title-jump-in" />` kinda painful, but the explorer does not work on mobile and to avoid this:

image


Same [here](https://github.com/decentraland/marketplace/pull/361/files#diff-ee9c6187c80c7e62e602988b5ec31cd58057bd6ff11879d19e781ba9baeae9f0R36)

nicosantangelo (changes requested):
Last thingy

webapp/src/components/NFTPage/EstateDetail/EstateDetail.css
LautaroPetaccio: Oh shoot, you're right, I forgot about that!

nicosantangelo (approved):
daledaledale

PR #362, opened 2021-07-09
fix: Show all parcels in estate using tiles

### LINKED ISSUE #360: Show all parcel coordinates for big Estates
Thegraph's max query size is 100, that makes parcels for estates be capped at that number. See for example https://market.decentraland.net/contracts/0x959e104e1a4db6317fa58f8295f586e1a978c297/tokens/1096

To fix this, instead of having to get all the information from thegraph's server, we can use `tile`s. We're already doing something similar for rendering the Map, and also in the builder for rendering Estates, you can check `src/modules/land/selectors.ts` on the builder (`getCoordsByEstateId` or `getLandTiles`)

LautaroPetaccio:
This PR fixes the issue of receiving a lower amount of parcels for a estate because of limitations in the API by loading the parcels from the pre-loaded tiles.

Closes #360

### DIFF
--- a/webapp/src/components/NFTPage/EstateDetail/ParcelCoordinates/ParcelCoordinates.tsx
+++ b/webapp/src/components/NFTPage/EstateDetail/ParcelCoordinates/ParcelCoordinates.tsx
@@ -1,12 +1,9 @@
 import React from 'react'
-import { useSelector } from 'react-redux'
 import { Header } from 'decentraland-ui'
 import { Link } from 'react-router-dom'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
-import classNames from 'classnames'
 
 import { locations } from '../../../../modules/routing/locations'
-import { getTilesByEstateId } from '../../../../modules/tile/selectors'
 import { Row } from '../../../Layout/Row'
 import Coordinate from '../../../Coordinate/Coordinate'
 import { Collapsible } from '../../../Collapsible'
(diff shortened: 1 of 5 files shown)


webapp/src/components/NFTPage/EstateDetail/ParcelCoordinates/ParcelCoordinates.tsx
```diff
@@ -1,26 +1,30 @@
...
 import Coordinate from '../../../Coordinate/Coordinate'
 import { Collapsible } from '../../../Collapsible'
 import { Props } from './ParcelCoordinates.types'
 import styles from './ParcelCoordinates.module.css'
 
 const ParcelCoordinates = (props: Props) => {
-  const { estate } = props
+  const { estateId } = props
   const coordinatesClasses = classNames(styles.coordinates)
+  const estateAndParcelCoordinates = useSelector(getTilesByEstateId)
```
nicosantangelo, line 19: this works great but to keep consistent with our [standards on file structure](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#2-naming-conventions-and-file-structure) I'd add a .container file to ParcelCoordinates and connect it to the store there, where you can use getTilesByEstateId and pass down ParcelCoordinates to this component

nicosantangelo (changes requested):
Minor comment

nicosantangelo (approved):

PR #364, opened 2021-07-15
feat: new nft server

cazala:
This PR adds support for the new NFT Server: https://github.com/decentraland/nft-server/pull/20


webapp/src/modules/nft/types.ts
```diff
@@ -35,19 +27,12 @@ export type Data<V extends VendorName> = V extends VendorName.DECENTRALAND
...
-  owner: string
-  name: string
-  category: NFTCategory
-  image: string
-  url: string
+export type NFT<V extends VendorName = VendorName.DECENTRALAND> = Omit<
+  BaseNFT,
+  'category' | 'data'
+> & {
+  category: NFTCategory | 'art'
```
nicosantangelo, line 34: I don't want you to change the entire codebase, but would I was thinking, we had two options here: 
- Extend NFTCategory like we did with `NFT` with `ART = 'art'` and import it from here
- Have a ExtraNFTCategories or something of the like (not the best name) just to avoid having to use `'art'` [as a string](https://github.com/decentraland/marketplace/pull/364/files#diff-0dfee1733e8d438adbe789f29f75869ec52327681f0689377b8abd19ace65480R23) [below](https://github.com/decentraland/marketplace/pull/364/files#diff-caabffcbe30dd87b0c61320207d0cb2b3bafde9d373bf10d844d7fff910dbb4fR36)

still, this is not necessary, that `'art'` is typed, but wanted to put it out there regardless

webapp/src/modules/nft/wearable/types.ts
```diff
@@ -1,62 +1,6 @@
...
-  FEMALE = 'BaseFemale'
-}
-
-export type Wearable = {
-  description: string
-  category: WearableCategory
-  rarity: WearableRarity
-  bodyShapes: BodyShape[]
-}
+export type Wearable = NFT['data']['wearable']
```
nicosantangelo, line 3: Given that we're using `NFT['data']['parcel']` and `NFT['data']['estate']` in all other places, maybe we can just delete this type also, right?

webapp/src/modules/nft/estate/utils.ts
```diff
@@ -2,12 +2,12 @@ import { Eth } from 'web3x-es/eth'
 import { Address } from 'web3x-es/address'
 import { getConnectedProvider } from 'decentraland-dapps/dist/lib/eth'
 import { EstateRegistry } from '../../../contracts/EstateRegistry'
-import { Estate } from './types'
-import { NFTCategory } from '../types'
+import { NFT } from '../types'
 import { getContract } from '../../contract/utils'
+import { NFTCategory } from '@dcl/schemas'
```
nicosantangelo, line 7: Import order! 🌚

nicosantangelo (commented):
WOooorrookkkss great!

### General Question!

I'm a bit confused about which `NFT` type we should be using. We sometimes use the base `NFT` from `common-schemas` but we also have the one we export from `vendor/types`. Is there a rule for each I'm not seeing?

webapp/src/components/Vendor/NFTFilters/FiltersMenu/FiltersMenu.tsx
```diff
@@ -42,9 +38,9 @@ const FiltersMenu = (props: Props) => {
   }, [])
 
   const rarityOptions = useMemo(() => {
-    const options = Object.values(WearableRarity)
-      .filter(x => x !== WearableRarity.COMMON && x !== WearableRarity.UNIQUE)
-      .reverse()
+    const options = Object.values(Rarity)
```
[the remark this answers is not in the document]
cazala: yep, maybe just as `Rarity.values` like it has `Rarity.schema`

webapp/src/modules/nft/types.ts
cazala: tbh i left it like this becuase after primary sales the idea would be to get rid of the ART category and the `vendor` abstraction no?

cazala:
@nicosantangelo regarding the use of `NFT` from `@dcl/schemas`, it should only be used to create the `NFT` type of the project, and then all the references to `NFT` should point to the one in the project. This wasn't the case tho, I just fixed it, now the common schema for the `NFT` is used only to generate the project's `NFT` and `DecentralandData` types.

PR #365, opened 2021-07-15
fix: Change environment variable to reflect the correct environment

LautaroPetaccio:
This PR changes the environment variable used for rollbar to another one reflecting the correct environment.


webapp/src/modules/analytics/rollbar.ts
```diff
@@ -7,9 +7,9 @@ let _rollbarConfig = {
   captureUncaught: true,
   captureUnhandledRejections: true,
   payload: {
-    environment: process.env.NODE_ENV
+    environment: process.env.REACT_APP_ENVIRONMENT
   },
-  checkIgnore: function () {
```
nachomazzara, line 12: maybe we can export this variable from `utils/environment.ts` as `export const environment = process.env.REACT_APP_ENVIRONMENT`

Also, at that file we will need to replace the usage of `NODE_ENV`

LautaroPetaccio: I think we should revert the changes done in [this file](https://github.com/decentraland/marketplace/pull/363/files#diff-59062bd3db245b3ecf8e63f0cf25d6d215a8351c093c8f5d9bdeb347046d1f84R23), because there it needs the NODE_ENV to be working as it was before.

webapp/src/modules/analytics/rollbar.ts
```diff
@@ -1,16 +1,18 @@
 /* eslint-disable */
 
-import { isDevelopment } from '../../lib/environment'
+import { environment } from '../../lib/environment'
+
+console.log(process.env.REACT_APP_ROLLBAR_ACCESS_TOKEN)
```
nachomazzara, line 5: Oops

LautaroPetaccio: Removed!

webapp/src/modules/analytics/rollbar.ts
```diff
@@ -1,16 +1,18 @@
...
   captureUncaught: true,
   captureUnhandledRejections: true,
   payload: {
-    environment: process.env.NODE_ENV
+    environment
   },
-  checkIgnore: function () {
-    if (window.location.hostname === 'localhost' || isDevelopment) {
+  checkIgnore: function() {
+    if (window.location.hostname === 'localhost') {
```
nachomazzara, line 13: What do you think of ignoring rollbar also if it is development and/or staging

LautaroPetaccio: That makes sense, we're not using staging extensively as to give value to rollbar there. I've just changed it.

nachomazzara (approved):

PR #398, opened 2021-08-09
feat: show supply in available

cazala:
Show supply next to available in item detail:

Screen Shot 2021-08-09 at 2 08 28 PM


webapp/src/components/AssetPage/ItemDetail/ItemDetail.tsx
```diff
@@ -19,6 +20,7 @@ import { WearableRarity } from '../WearableRarity'
 import { WearableHighlights } from '../WearableHighlights'
 import { WearableCollection } from '../WearableCollection'
 import { Props } from './ItemDetail.types'
+import './ItemDetail.css'
```
nicosantangelo, line 23: we could use a `module` here right? giving that it's a new stylesheet

nicosantangelo (commented):
Tiny comment

webapp/src/components/AssetPage/ItemDetail/ItemDetail.tsx
cazala: done :)

PR #407, opened 2021-08-11
chore: Update dependencies

LautaroPetaccio:
This PR updates the marketplace dependencies.
Most of the dependencies don't have breaking changes that affect the Marketplace but there's a minor exception (not a breaking change) of the React 17 upgrade that makes the `react-script` dependency change the `jsx` property to `react-jsx`.


webapp/src/modules/vendor/MarketplacePrice.ts
```diff
@@ -25,8 +25,8 @@ export class MarketplacePrice {
       )
     }
 
-    this.feePerMillion = toBN(feePerMillion)
-    this.oneMillion = toBN('1000000')
+    this.feePerMillion = new BN(feePerMillion)
+    this.oneMillion = new BN('1000000')
```
nicosantangelo, line 29: Question: we're still using toBN on this file, why are we changing only these calls to it?

webapp/src/components/AssetPage/ItemDetail/ItemDetail.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
```
nicosantangelo, line 1: Should we delete this line here too?

webapp/src/modules/vendor/MarketplacePrice.ts
LautaroPetaccio: There was an issue with the typing, so I changed it for the ones that worked with the rest of the code.
I think that the concept of using `toBN` as a proxy of another library is counterproductive, but I wanted to keep code changes to a minimum, specially because I updated a lot of stuff, so I didn't change the other ones.

webapp/src/components/AssetPage/ItemDetail/ItemDetail.tsx
LautaroPetaccio: We're using `Memo`, so we can delete it :(

nicosantangelo: oh! my bad couldn't see it on the diff

webapp/src/modules/vendor/MarketplacePrice.ts
nicosantangelo: I don't particularly like `toBN` either, but for what I can see the `webapp`  does not have bn as an explicit dependency. I understand that this should work anyways but maybe we should add it?

We were using toBN because it comes bundled with web3-es, which we were already using.

One las thing, maybe we should add a comment/TODO to this file explaining why we're doing this in two different ways for future reference (we can even link this PR)

LautaroPetaccio: It seemed better to update the other parts of the code than adding a comment, so I updated them.
I also added BN as a dependency.

nicosantangelo (approved):
Thanks for the BN update 🚀

PR #412, opened 2021-08-24
feat: Add clear all filters

LautaroPetaccio:
This PR adds a clear filters component button to clear all the filters when searching for items or NFTs in the market.
clear-filters-mobile
clear-filters

### DIFF
--- a/webapp/src/modules/routing/sagas.ts
+++ b/webapp/src/modules/routing/sagas.ts
@@ -1,4 +1,4 @@
-import { takeEvery, put, select } from 'redux-saga/effects'
+import { takeEvery, put, select, call } from 'redux-saga/effects'
 import { push, getLocation } from 'connected-react-router'
 import { NFTCategory } from '@dcl/schemas'
 import { VendorName } from '../vendor/types'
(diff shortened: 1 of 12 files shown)


webapp/src/modules/routing/sagas.ts
```diff
@@ -59,21 +61,43 @@ function* handleFetchAssetsFromRoute(action: FetchAssetsFromRouteAction) {
...
+function* handleClearFilters() {
+  const browseOptions: BrowseOptions = yield call(getCurrentBrowseOptions)
+  const { pathname }: ReturnType<typeof getLocation> = yield select(getLocation)
+
+  // Delete filters and reset the page number.
+  delete browseOptions.wearableRarities
+  delete browseOptions.wearableGenders
+  delete browseOptions.network
+  delete browseOptions.contracts
+  delete browseOptions.page
```
nicosantangelo, line 73: Wouldn't it be better to just create a new object here instead of mutating the one that comes from `getCurrentBrowseOptions`?

Without looking at the code, I can tell this would work anyways because `getCurrentBrowseOptions` is not a selector getting a slice directly from the state, but that could change tomorrow and we could have a hard to track bug here.

You could use `decentraland-commons`'s `utils.omit` if you want

nicosantangelo (changes requested):
ausum

webapp/src/modules/routing/sagas.ts
LautaroPetaccio: You're completely right. I didn't want to add another dependency or code an omit function to be used only here and I didn't know that we had an omit function in `decentraland-common`.
I've added the library and used the omit function.

PR #414, opened 2021-08-26
feat: support native polygon transactions

cazala:
This PR adds support for setting the provider in the polygon network and sending regular transactions on polygon. Some parts of the UI that interact with ethereum will be disabled if the provider is on the polygon network.


LautaroPetaccio (changes requested):
LGTM, the only thing that needs to be added is the `TRANSACTIONS_API_URL`

LautaroPetaccio (approved):

PR #419, opened 2021-09-07
fix: Partners not showing up

LautaroPetaccio:
This PR fixes three issues with the partners feature:
1. The partners were not being displayed in staging and production because the `kyber_network` isn't returning the prices for the MANA token.
2. The partners tab was displaying decentraland items whenever a user got into it for the first time.
3. Any partners tab showed the `ToggleBox` with `Store` as the selected thing to see, but it should only show `Listings`.

To fix 1, `kyber_network` was replaced for `uniswap_v2` in the environment variables.
To fix 2, a fix was made to make some components receive the correct type for `Section` which was before a map of the sections available instead of the set of all possible sections. This fix worked together with changes in the `getSection` and `getAssetType` selectors which now return the correct output based on the vendors.
To fix 3, the logic in `AssetBrowse.tsx` that defined if an item in the `ToggleBox` is active or disabled changed to consider if the vendor is DCL or not.


webapp/src/modules/vendor/routing/types.ts
```diff
@@ -1,12 +1,18 @@
...
+
+export type Section =
+  | decentraland.Section
+  | superRare.Section
+  | makersPlace.Section
+  | knownOrigin.Section
 
 // eslint-disable-next-line @typescript-eslint/no-redeclare -- Intentionally naming the variable the same as the type
-export const Section = {
+export const Sections = {
```
LautaroPetaccio, line 15: This was not working with the older imports. The decentraland section was blank.

cazala (changes requested):
When visiting the `Account` page (either my own assets or seeing other user's assets) I can click on `All Assets` under the Decentraland section and it does load all the asset, but the sidebar doesn't show the "All Assets" section selected, it stays showing Wearables highlighted.

cazala (approved):

PR #429, opened 2021-09-21
chore: Add NFT Service tests

LautaroPetaccio:
This PR does the following:
- Adds tests for the NFT Service.
- Rewrites the fetch method, fixing a bug where the accounts were not computed correctly.

### DIFF
--- a/webapp/src/modules/vendor/decentraland/NFTService.spec.ts
+++ b/webapp/src/modules/vendor/decentraland/NFTService.spec.ts
@@ -70,12 +70,11 @@ describe("Decentraland's NFTService", () => {
     })
 
     describe('when the fetch is successful', () => {
-      let total: number
+      const total = 3
       let nftResults: NFTResult[]
       let anotherNFT: NFT
 
       beforeEach(() => {
-        total = 3
         anotherNFT = { ...nft, owner: 'anotherAddress', id: 'anotherNFTID' }
         nftResults = [
           { nft, order },
@@ -138,10 +137,9 @@ describe("Decentraland's NFTService", () => {
     })
 
     describe('when the fetch is successful', () => {
-      let total: number
+      const total = 1
 
       beforeEach(() => {
-        total = 1
         ;(api.nftAPI.fetch as jest.Mock).mockResolvedValueOnce({ total })
       })
 
--- a/webapp/src/modules/vendor/decentraland/NFTService.spec.ts
+++ b/webapp/src/modules/vendor/decentraland/NFTService.spec.ts
@@ -1,0 +1,344 @@
+import { Network, ChainId } from '@dcl/schemas'
+import { ContractData } from 'decentraland-transactions'
+import * as walletUtils from 'decentraland-dapps/dist/modules/wallet/utils'
+import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
+import ERC721Abi from '../../../contracts/ERC721Abi'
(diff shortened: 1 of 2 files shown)


webapp/src/modules/vendor/decentraland/NFTService.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+            aTxHash
+          )
+        })
+
+        it("should have called send transaction with the erc721's crafted contract using the nft's chain id, the contract address and the ABI", async () => {
+          await nftService.transfer(wallet, anAddress, nft)
+          expect(walletUtils.sendTransaction as jest.Mock).toHaveBeenCalled()
+
+          const firstParameter = (walletUtils.sendTransaction as jest.Mock).mock
+            .calls[0][0] as ContractData
```
fzavalia, line 247: I think there is an assert function for this called `.toHaveBeenCalledWith` that could help you not have to get into the mock data.

fzavalia: You used that method further in the code 👍

webapp/src/modules/vendor/decentraland/NFTService.ts
```diff
@@ -20,21 +20,28 @@ export class NFTService
...
-        account = this.toAccount(address)
-      }
-      account.nftIds.push(result.nft.id)
-      nfts.push({ ...result.nft, vendor: VendorName.DECENTRALAND })
-      if (result.order) {
-        orders.push(result.order)
-      }
-    }
+    const accounts: Account[] = results.reduce(
+      (accumulator: Account[], nftResult: NFTResult) => {
```
fzavalia, line 24: Types are infered by the initial accumulator and the results array, you can remove them

webapp/src/modules/vendor/decentraland/NFTService.spec.ts
LautaroPetaccio: `toHaveBeenCalledWith` can be used only if you know in advance both arguments. The problem with this is that the second parameter of the method is a lambda function and the only way to verify that is does what we think it does it to execute it. There's no equality comparison between function if we're not comparing references.

webapp/src/modules/vendor/decentraland/NFTService.ts
LautaroPetaccio: You're right! I have just changed it.

webapp/src/modules/vendor/decentraland/NFTService.spec.ts
fzavalia: In that case you can use `expect.any(Function)` as the second argument

fzavalia: Or even `expect.anything()` if you don't care for it to a function

LautaroPetaccio: That is great! I completely forgot about those matchers.

fzavalia: Ending with something like:

```
          expect(walletUtils.sendTransaction as jest.Mock).toHaveBeenCalledWith(
            {
              chainId: nft.chainId,
              name: 'ERC721',
              abi: ERC721Abi,
              address: nft.contractAddress,
              version: '1'
            },
            expect.any(Function)
          )
```

fzavalia (approved):
LGTM!
Nice job 👍

PR #451, opened 2021-11-02
feat: index sales

cazala:
This PR adds support for the `Sale` entity and analytics added for `Account` and `NFT` entities, to track the same data as we currently do on the collections' subgraph


nachomazzara (approved):
LGTM

PR #452, opened 2021-11-03
feat: prevent deploying wrong network

cazala:
I replaced the `deploy` command with `deploy:ropsten` and `deploy:mainnet`, which performs a build prior to deploying, to prevent deploying the wrong network to a subgraph.


fzavalia (approved):
LGTM

Issue #447, opened 2021-10-29
Validate addresses at /account


nachomazzara:
We need to validate if the address passed is valid.

https://market.decentraland.io/accounts/pepe should not be valid.

https://market.decentraland.io/accounts/0x4e54d0b254113b381e5fbdb2488570f2cdf0ac87 should be valid

mercush:
Addressing this rn. Will likely have pr by end of day.

juanmahidalgo:
Addressed & deployed to prod.

PR #460, opened 2021-11-13
address validation for accounts

mercush:
Here is my attempt to resolve [this issue](https://github.com/decentraland/marketplace/issues/447). Please forgive me if my code is bad. This is my first attempt at an open-source contribution. Would be very happy to take suggestions :).

### DIFF
--- a/webapp/src/components/AccountPage/AccountPage.tsx
+++ b/webapp/src/components/AccountPage/AccountPage.tsx
@@ -5,7 +5,7 @@ import { Icon } from 'semantic-ui-react'
 import { Profile } from 'decentraland-dapps/dist/containers'
 import { isMobile } from 'decentraland-dapps/dist/lib/utils'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
-import { BaseAPI } from 'decentraland-dapps/dist/lib/api'
+import { checkValidAddress, parseAddress } from '../../modules/account/selectors'
 
 import { View } from '../../modules/ui/types'
 import { useTimer } from '../../lib/timer'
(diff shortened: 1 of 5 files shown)


webapp/src/components/AccountPage/AccountPage.tsx
```diff
@@ -32,14 +33,28 @@ const AccountPage = (props: Props) => {
   const isCurrentAccount =
     address === undefined || (wallet && wallet.address === address)
 
+  const checkValidAddress = async () => {
+    const API_URL = process.env.REACT_APP_MAKERS_PLACE_API_URL!
+    const baseAPI = new BaseAPI(API_URL)
+    const params = {
+      owner_address: address
+    }
+    const response = baseAPI.request('get', '/assets/', params)
```
nachomazzara, line 42: What do you think of just use just web3x or ethers to check if the address is valid or not? also, we should accept `ens` in order to resolve them in the near future https://app.zenhub.com/workspaces/dapps-5ffc44ecec9f8500140e173c/issues/decentraland/marketplace/448

nachomazzara (changes requested):
Thanks for the collaboration! I left a comment

mercush:
I made another commit that supports ens resolution for /account and uses web3 to validate addresses. Let me know if you have any other suggestions!

cazala:
Hey @mercush this is awesome! There's one thing that would make it even better, we are trying to move away from web3x (since it's not maintained anymore) and use ethers instead. There are still some parts of the code that use it (like the Bids module) but everything else has been re-written to ethers and hopefully soon we'll be able to remove the dependency and use only ethers. Would you like to use the the ethers ENS helpers instead of the web3x ones? Here's some docs on how to use it, it's pretty much the same thing as web3x: https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods

mercush:
Hey, I've been busy for the last couple of weeks so I haven't worked on this much. I'm trying to implement ethers right now but I don't see anywhere in the repo where it's been used before. Can you give me an example of a file where ethers has been used instead of web3x?

cazala:
hey @mercush, we usually use a helper `sendTransaction` on the dApps that abstract several things away (like getting the right provider, sending either a transaction or a meta-transaction according to the connected chain, etc), but you can look into that helper to find out how to get a provider and stuff like that: https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/wallet/utils.ts#L86-L138

mercush:
Ok. I think that should do the trick. Let me know if there's anything else I should do to this pr.

webapp/src/modules/account/selectors.ts
```diff
@@ -19,3 +23,23 @@ export const getAddress = createSelector<
...
   string | undefined
 >(accountMatchSelector, match => match?.params.address?.toLowerCase())
+
+export const parseAddress = async (address: string) => {
+  if (address === undefined) {
+    return undefined
+  }
+  var parsedAddress = address
+  const provider = await ethers.getDefaultProvider()
+  await provider.getResolver(address).then((res) => {
```
cazala, line 33: This seems to always return the address of the resolver, but not the address mapped by the ENS. No matter what ENS I use to test this it always redirects me to `0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41` which is the ENS resolver contract.

webapp/src/modules/account/selectors.ts
```diff
@@ -19,3 +23,23 @@ export const getAddress = createSelector<
...
+    console.log('res', res)
```
cazala, line 34: oops this console log slipped thru

webapp/src/modules/account/selectors.ts
```diff
@@ -19,3 +23,23 @@ export const getAddress = createSelector<
...
+    if (res) {
+      parsedAddress = res.address?.toString()!
+      console.log('parsed address', parsedAddress)
```
cazala, line 37: same as above

cazala (changes requested):

nachomazzara:
@mercush are you willing to finish the PR?

mercush:
Yes. I've just been busy for the last couple of weeks.

mercush:
On second thought, I probably won't be able to work on this again for a while so I don't think I'll finish it.

PR #468, opened 2021-11-24
feat: Collection Detail Page

fzavalia:
Created the collection detail page to show the different items a collection contains
Buttons in the detail page will redirect to the builder in this iteration of the feature.
This page can be accessed by clicking on a collection in the collection list view or directly via url.
Added a CollectionProvider that handles the availability of a collection in the store and its items (optionally)

`http://localhost:3000/collections/0xbcf5784c4cfa38ba49253527e80c9e9510e01c67`

image

`https://builder.decentraland.io/?redirectTo=%7B%22type%22%3A%22COLLECTION_DETAIL_BY_CONTRACT_ADDRESS%22%2C%22data%22%3A%7B%22contractAddress%22%3A%220xbcf5784c4cfa38ba49253527e80c9e9510e01c67%22%7D%7D`


webapp/src/components/AssetCard/AssetCard.tsx
```diff
@@ -12,7 +12,7 @@ import { WearableTags } from './WearableTags'
 import { ENSTags } from './ENSTags'
 import { AssetImage } from '../AssetImage'
 import { Props } from './AssetCard.types'
-import listedSvg from '../../images/listed.svg'
```
fzavalia, line 15: Removed the svg because I could use the icon from decentralan-ui instead

webapp/src/components/CollectionList/CollectionList.tsx
```diff
@@ -27,18 +29,6 @@ const CollectionList = ({
     { value: SortBy.SIZE, text: t('filters.size') }
   ])
 
-  // Amount of collections without applying any filters.
```
fzavalia, line 30: The pros of doing this were less than not doing it so I'm removing it.
I was storing the first count which could be used as a total of all collections, however when fetching a collection for the collection detail page and going back to the list page, the total showed as 1.
Some other workaround could be developed but it was not worth the effort for the small problem it solved

webapp/src/components/CollectionPage/CollectionPage.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { formatMANA } from '../../lib/mana'
+import styles from './CollectionPage.module.css'
+import { Rarity } from '@dcl/schemas'
+import { getBuilderCollectionDetailUrl } from './utils'
+
+const CollectionPage = (props: Props) => {
+  const { collection, items, isLoading, onFetchCollection, onBack } = props
+
+  useEffect(() => {
+    onFetchCollection()
```
fzavalia, line 35: Not gonna change it now, but maybe this sort of side effects could be placed inside the location sagas and in this case, whenever the collection page is accessed, the collection could be fetched inside the saga

webapp/src/components/CollectionPage/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { Props } from './CollectionPage.types'
+
+export const getContractAddressFromProps = (props: Props) =>
+  (props.match!.params as any).contractAddress as string
+
+export const getBuilderCollectionDetailUrl = (contractAddress: string) =>
+  `${process.env.REACT_APP_BUILDER_URL}?redirectTo=${encodeURIComponent(
```
fzavalia, line 7: We don't seem to have a place to store constants or config so I'm using the env here directly

webapp/src/components/CollectionPage/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+    JSON.stringify({
+      type: 'COLLECTION_DETAIL_BY_CONTRACT_ADDRESS',
```
fzavalia, line 9: This could be built by some common lib between the builder and the marketplace

webapp/src/components/OnSaleList/OnSaleListElement/OnSaleListElement.module.css
```diff
@@ -1,9 +1,9 @@
-.first-cell {
```
fzavalia, line 1: .module styles autocomplete seem to have a better time with camelCase than the kebab-case.
(styles.firstCell instead of styles['first-cell])

webapp/src/components/CollectionPage/CollectionPage.tsx
cazala: the approach that we usually take for this type of behavior is creating an `AssetProvider` (ie `CollectionProvider`) that you can pass the id of the asset and it will load it, and pass it down to the children. We have this for items and nfts in the marketplace, for lands and other stuff in the builder. So then you don't really care about the routes where it's needed, you just wrap your render with this provider and you can use the asset as if it was already loaded. Same thing can be also created as a hook here since we are using React 17, like `const [collection, isLoading] = useCollection(address)`

webapp/.ci/.env.dev
```diff
@@ -7,6 +7,7 @@ REACT_APP_NFT_SERVER_URL=https://nft-api.decentraland.io/v1
 REACT_APP_ATLAS_SERVER_URL=https://api.decentraland.io
 REACT_APP_PEER_URL=https://peer.decentraland.zone
 REACT_APP_TRANSACTIONS_API_URL=https://transactions-api.decentraland.io/v1
+REACT_APP_BUILDER_URL=https://builder.decentraland.io
```
cazala, line 10: ⚠️ We should add this to the [vercel.json](https://github.com/decentraland/marketplace/blob/master/webapp/vercel.json) otherwise the pull request previews won't work

cazala (changes requested):

webapp/src/components/CollectionPage/CollectionPage.tsx
```diff
@@ -0,0 +1,173 @@
...
+                      </Table.HeaderCell>
+                      <Table.HeaderCell>{t('global.rarity')}</Table.HeaderCell>
+                      <Table.HeaderCell>{t('global.stock')}</Table.HeaderCell>
+                      <Table.HeaderCell>{t('global.price')}</Table.HeaderCell>
+                      <Table.HeaderCell />
+                    </Table.Row>
+                  </Table.Header>
+                  <Table.Body>
+                    {items.map(item => (
+                      <Table.Row key={item.id} className={styles.row}>
```
cazala, line 117: Should we wrap this with a `<Link to={locations.itemDetail(item.id)}>` ?

fzavalia: Absolutely 🙂

webapp/.ci/.env.dev
fzavalia: Added

webapp/src/components/CollectionPage/CollectionPage.tsx
fzavalia: I went for  the CollectionProvider pattern because we currently have no hooks for this and didn't want to change what we already use.
The good thing about the provider is that it does not request everytime the user enters the page as the required data is already in the store.
Thanks for the idea @cazala 😁

webapp/src/components/CollectionPage/CollectionPage.tsx
fzavalia: Done, I wrapped only the image and name cell instead of the whole row. This is because it might bring conflicts with the dropdown, even when preventing bubbling (or propagation, I don't remember the name of what prevents events to travel upwards) as the user might click the row instead and be confused as to how they traveled to the item instead of the builder.

fzavalia: And TBH adding the link to the whole row was messing up the styles 😅

cazala (approved):

PR #475, opened 2021-11-30
feat: Update asset details

fzavalia:
Update NFT and Item detail pages

image
image
image
image
image
image


webapp/src/components/AssetBrowse/Box/Box.types.ts
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 
 export type Props = {
-  header: string
+  header?: string
```
fzavalia, line 4: Made the header optional so it gives me the possibility to have a completely empty box

webapp/src/components/AssetPage/BaseDetail/BaseDetail.module.css
```diff
@@ -0,0 +1,55 @@
+.detail .supply {
```
fzavalia, line 1: I have both a plain css and a module css because on the one hand I need to do some styling overrides only allowed by the plain one, on the other hand it seems that the defacto styling for the market is by using modules. When cases like this come through, I'd rather have only the plain one. Will change if reviewer agrees

webapp/src/components/AssetPage/BaseDetail/BaseDetail.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Container } from 'decentraland-ui'
+import { AssetImage } from '../../AssetImage'
+import { PageHeader } from '../../PageHeader'
+import Title from '../Title'
+import { Box } from '../../AssetBrowse/Box'
+import ListedBadge from '../../ListedBadge'
+import { Props } from './BaseDetail.types'
+import styles from './BaseDetail.module.css'
+import './BaseDetail.css'
+import classNames from 'classnames'
```
fzavalia, line 11: Reorder imports

webapp/src/components/AssetPage/BaseDetail/BaseDetail.tsx
```diff
@@ -0,0 +1,49 @@
...
+  isOnSale,
+  badges,
+  left,
+  box,
+  below
+}: Props) => {
+  return (
+    <div className={classNames('BaseDetail', styles.detail)}>
+      <PageHeader>
+        <AssetImage asset={asset} {...assetImageProps} />
```
fzavalia, line 25: Pass entire component instead of props

webapp/src/components/AssetPage/Expiration/Expiration.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const Expiration = ({ order }: Props) => {
+  if (!order) {
+    return null
+  }
+
+  return (
+    <div className={styles.container}>
+      <Icon name="clock outline" />
+      Expires{' '}
```
fzavalia, line 15: Translate this

webapp/src/components/AssetPage/IconBadge/IconBadge.module.css
```diff
@@ -0,0 +1,21 @@
+.badge {
```
fzavalia, line 1: Use just .css instead of module

webapp/src/components/AssetPage/BaseDetail/BaseDetail.module.css
fzavalia: Using plain css now

webapp/src/components/AssetPage/BaseDetail/BaseDetail.tsx
fzavalia: Reordered

webapp/src/components/AssetPage/Expiration/Expiration.tsx
fzavalia: Translated

webapp/src/components/AssetPage/IconBadge/IconBadge.module.css
fzavalia: Used just .css

cazala (approved):

PR #480, opened 2021-12-07
feat: Added goBack action that defaults to root when there is no history

fzavalia:


fzavalia:
@cazala I tried to implement how it is done in the builder, but there is a flaw with that one because it only works when you open a page and there is no back history, yet if you navigate and then go back until you have no more back history left, the back button will do nothing. It does not happen with this implementation. (We should update the builder too, even abstract this to dapps would be a nice idea)

cazala (approved):

PR #483, opened 2021-12-13
feat: Settings page layout

fzavalia:
Page layout for settings.
Is not connected to anything, just allows a cover image to be picked and inputs to be changed, all done locally.
Will connect when there is somewhere to upload the store settings

image

image


webapp/src/modules/translation/locales/en.json
```diff
@@ -121,6 +121,19 @@
...
+    "see_store_as_guest": "See store as guest",
+    "store_cover": "Store cover",
+    "description": "Description",
+    "website": "Website",
+    "facebook": "Facebook",
+    "twitter": "Twitter",
+    "discord": "Discord",
+    "save": "Save",
+    "revert": "Revert",
+    "add_cover_picture": "Click here to add a cover picture"
```
cazala, line 135: these are missing in the other locales

fzavalia: you are right!

fzavalia: done 😁

cazala (approved):

PR #485, opened 2021-12-15
feat: added support for bids on polygon

### LINKED ISSUE #476: Support Bids in polygon

cazala:
Closes #476


webapp/src/modules/vendor/decentraland/BidService.ts
```diff
@@ -41,72 +42,81 @@ export class BidService
...
 
-    if (fingerprint) {
-      return bids.methods
-        .placeBid(
-          Address.fromString(nft.contractAddress),
-          nft.tokenId,
-          priceInWei,
-          expiresIn,
-          fingerprint
+    console.log(priceInWei, priceInWei.toString())
```
fzavalia, line 52: Remove log

webapp/src/modules/vendor/decentraland/BidService.ts
```diff
@@ -41,72 +42,81 @@ export class BidService
...
-        .send({ from })
-        .getTxHash()
-    } else {
-      return bids.methods
-        .placeBid(
-          Address.fromString(nft.contractAddress),
-          nft.tokenId,
-          priceInWei,
-          expiresIn
+        return sendTransaction(contract, bids =>
```
fzavalia, line 84: Why bids instead of bid like on ethereum?

webapp/src/modules/vendor/decentraland/BidService.ts
```diff
@@ -41,72 +42,81 @@ export class BidService
...
-      return bids.methods
-        .placeBid(
-          Address.fromString(nft.contractAddress),
-          nft.tokenId,
-          priceInWei,
-          expiresIn,
-          fingerprint
+    console.log(priceInWei, priceInWei.toString())
+
+    switch (nft.network) {
```
fzavalia, line 54: Maybe it would be nice to have tests for the place function. At least to see that the function goes to the correct place depending on the nft.

webapp/src/modules/vendor/decentraland/BidService.ts
cazala: i started calling it `bid`, then i coded the `cancel` method, which receives a `bid: Bid` object, so I could not use that name. But it's true that they should not be all named different, I would rather call the contract `bids` in all places to avoid the name collision.

fzavalia (approved):
LGTM ✅

PR #486, opened 2021-12-16
feat: Save store settings changes locally

fzavalia:
+ Add `viewAsGuest` query param to see your store as a visitor
+ Update dcl schemas to the version with Store type
+ Revert button reverts store settings when changes have been made

More functionality will be added to this once the entity can be queried.


webapp/src/components/StoreSettings/StoreSettings.container.ts
```diff
@@ -0,0 +1,32 @@
...
+import StoreSettings from './StoreSettings'
+import { getLocalStore } from '../../modules/store/selectors'
+import { Dispatch } from 'redux'
+import { Store } from '../../modules/store/types'
+import { revertLocalStore, updateLocalStore } from '../../modules/store/actions'
+import { getEmptyLocalStore } from '../../modules/store/utils'
+
+const mapState = (state: RootState): MapStateProps => {
+  const address = getAddress(state)
+  const baseStore = getEmptyLocalStore()
```
fzavalia, line 14: This will change once we have the store entity and will use that as base and the empty when there is none

webapp/src/modules/store/reducer.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  RevertLocalStoreAction,
+  REVERT_LOCAL_STORE,
+  UpdateLocalStoreAction,
+  UPDATE_LOCAL_STORE
+} from './actions'
+import { Store } from './types'
+import { getEmptyLocalStore } from './utils'
+
+export type StoreState = {
```
fzavalia, line 10: This will be extended with the common data, loading, error props all other modules have once the Store entity is integrated. (I wanted to keep the PR simple)

webapp/src/modules/store/types.ts
```diff
@@ -0,0 +1,9 @@
+export type Store = {
```
fzavalia, line 1: Intermediate type to work with in the UI without depending on the one from dcl schemas

webapp/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "marketplace-ui",
   "version": "0.4.0",
   "dependencies": {
-    "@dcl/schemas": "^2.2.0",
+    "@dcl/schemas": "^2.2.2",
```
fzavalia, line 5: For this PR this bump is unnecessary, I can revert it but this version will be used shortly

cazala (approved):

PR #498, opened 2021-12-24
feat: Sales page in My Store

fzavalia:
This PR adds

- Sales page
- Whole Sales module
- AccountMetrics to the account module with sagas and reducers to handle fetching these metrics from the nft server

image


webapp/src/components/OnSaleList/AssetCell/AssetCell.tsx
```diff
@@ -0,0 +1,46 @@
...
+      subtitle = t('global.parcel_count', {
+        count: asset.data.estate!.parcels.length
+      })
+      break
+    case NFTCategory.PARCEL:
+      const { x, y } = asset.data.parcel!
+      subtitle = `${x},${y}`
+  }
+
+  const link = asset.itemId
```
cazala, line 25: Oops, polygon assets always have `itemId` (both items and nfts) so this will always produce links to items. It should be `asset.tokenId ? ...` instead.

cazala (changes requested):

cazala:
I noticed that the mobile view of the sales page has a strange layout: 

Screen Shot 2021-12-27 at 16 53 45

We could do something more like the transaction history:

Screen Shot 2021-12-27 at 16 52 40
7/147504117-e066ee11-ed05-4492-a62a-51325b0afc94.png">

I notice that is also the case with the On Sale page so i think we can just add a ticket to go over the newly added pages and make them more responsive if needed, no need to do it on this PR 👍

webapp/src/components/OnSaleList/AssetCell/AssetCell.tsx
fzavalia: Sure! thanks for the catch

cazala (approved):

PR #502, opened 2021-12-27
feat: Update and View Store entities from the catalysts

fzavalia:
Also add the identity module to sign deployments.
A modal to show that you are exiting decentraland and to be careful with the link.
Link validations as well as only setting handles and ids instead of full links.

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -8,6 +8,7 @@
     "classnames": "^2.3.1",
     "connected-react-router": "^6.9.1",
     "date-fns": "^2.23.0",
+    "dcl-catalyst-client": "^5.1.1",
     "decentraland-dapps": "^12.33.2",
     "decentraland-transactions": "^1.29.1",
     "decentraland-ui": "^3.15.0",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -9,6 +9,8 @@
     "connected-react-router": "^6.9.1",
     "date-fns": "^2.23.0",
     "dcl-catalyst-client": "^5.1.1",
+    "dcl-catalyst-commons": "^4.0.1",
+    "dcl-crypto": "^2.2.0",
     "decentraland-dapps": "^12.33.2",
     "decentraland-transactions": "^1.29.1",
     "decentraland-ui": "^3.15.0",
--- a/webapp/src/components/StoreSettings/StoreSettings.container.ts
+++ b/webapp/src/components/StoreSettings/StoreSettings.container.ts
@@ -5,29 +5,39 @@ import { MapStateProps, MapDispatchProps } from './StoreSettings.types'
 import StoreSettings from './StoreSettings'
 import {
   getData as getStoresByOwner,
-  getLocalStore
+  getLocalStore,
+  getLoading as getStoreLoading
 } from '../../modules/store/selectors'
 import { Dispatch } from 'redux'
 import { Store } from '../../modules/store/types'
 import {
   fetchStoreRequest,
+  FETCH_STORE_REQUEST,
   revertLocalStore,
   updateLocalStore,
-  updateStoreRequest
+  updateStoreRequest,
+  UPDATE_STORE_REQUEST
 } from '../../modules/store/actions'
 import { getEmptyLocalStore } from '../../modules/store/utils'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
 
 const mapState = (state: RootState): MapStateProps => {
   const address = getAddress(state)!
-  const baseStore = getStoresByOwner(state)[address!] || getEmptyLocalStore()
-  const store =
-    getLocalStore(state) || getStoresByOwner(state)[address!] || baseStore
+  const savedStore = getStoresByOwner(state)[address!]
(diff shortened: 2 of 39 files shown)


webapp/src/modules/store/sagas.ts
```diff
@@ -0,0 +1,109 @@
...
+        const response: Response = yield fetch(store.cover)
+        const arrayBuffer: Blob = yield response.arrayBuffer()
+        files.set(`cover/${store.coverName}`, Buffer.from(arrayBuffer))
+      }
+
+      const optionsWithFiles: BuildEntityOptions = { ...baseOptions, files }
+
+      entity = yield client.buildEntity(optionsWithFiles)
+    } else {
+      entity = yield client.buildEntityWithoutNewFiles(baseOptions)
```
cazala, line 63: should we wrap these `yield`s with a `call` for testability?

fzavalia: Yeah, I'll be adding tests now and that would be helpful

webapp/package.json
```diff
@@ -8,6 +8,9 @@
     "classnames": "^2.3.1",
     "connected-react-router": "^6.9.1",
     "date-fns": "^2.23.0",
+    "dcl-catalyst-client": "^5.1.1",
+    "dcl-catalyst-commons": "^4.0.1",
+    "dcl-crypto": "^2.2.0",
```
fzavalia, line 13: Using the same catalyst client version as the builder. The latest one had some issues when importing as, beside targeting ec2020 which is currently not supported by us at the moment. It had some node specific imports not supported by the web. I think that issue has been addressed, but lets keep this version of the client so we don't need to update o do any workarounds on the React app

webapp/src/components/StoreSettings/StoreSettings.container.ts
```diff
@@ -1,31 +1,55 @@
...
 const mapState = (state: RootState): MapStateProps => {
-  const address = getAddress(state)
-  const baseStore = getEmptyLocalStore()
-  const store = getLocalStore(state) || baseStore
+  const address = getAddress(state)!
+  const savedStore = getStoresByOwner(state)[address!]
+  const emptyStore = getEmptyStore()
+  const localStore = getLocalStore(state)
+  const baseStore = savedStore || emptyStore
+  const store = localStore || savedStore || baseStore
```
cazala, line 31: this can be `localStore || baseStore`

webapp/src/modules/identity/sagas.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function* identitySaga() {
+  yield takeLatest(GENERATE_IDENTITY_REQUEST, handleGenerateIdentityRequest)
+}
+
+function* handleGenerateIdentityRequest(action: GenerateIdentityRequestAction) {
+  const address = action.payload.address.toLowerCase()
+
+  try {
+    const eth: Eth = yield call(getEth)
```
cazala, line 30: lets change this for `ethers` instead of `web3x`

webapp/src/modules/identity/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!address) {
+    yield put(connectWalletRequest())
+
+    const { success }: { success: ConnectWalletSuccessAction } = yield race({
+      success: take(CONNECT_WALLET_SUCCESS),
+      failure: take(CONNECT_WALLET_FAILURE)
+    })
+
+    if (success) {
+      return yield call(getIdentity)
```
cazala, line 48: We could do the following to avoid the `unknown` return value
```
const identity: Identity = yield call(getIdentity)
return identity
```

webapp/src/modules/store/utils.ts
```diff
@@ -1,11 +1,182 @@
...
+  pushLink(LinkType.WEBSITE)
+  pushLink(LinkType.FACEBOOK)
+  pushLink(LinkType.TWITTER)
+  pushLink(LinkType.DISCORD)
+
+  const images: StoreEntityMetadata['images'] = []
+
+  if (store.cover && store.coverName) {
+    hasDifferentCover
+      ? images.push({ name: 'cover', file: `cover/${store.coverName}` })
```
cazala, line 90: could this `cover/` be added already to the `store` beforehand instead of using `hasDifferentCover`?

webapp/src/components/StoreSettings/StoreSettings.container.ts
fzavalia: Updated

webapp/src/modules/identity/utils.ts
fzavalia: This solution did not work, ended up adding `Generator<any, AuthIdentity, any>` as type

webapp/src/modules/store/utils.ts
fzavalia: Updated

webapp/src/modules/identity/sagas.ts
fzavalia: I think this could be done in a new PR as web3x is also being used in other places.

cazala (approved):

PR #506, opened 2021-12-29
feat: minor changes

cazala:
This PR adds a few changes:

- The `Get Started` button in the home page takes you to `Collectibles` instead of `Land`
- The `buy` verb was replaced with `mint` for primary sales
- Added the "Recently Sold" sorting option for items and nfts in the `AssetBrowse`  component, also added a `Latest sales` slideshow to the homepage with the recently sold items
- Prevent doing an unnecessary fetch on some sections in My Store


webapp/src/modules/nft/hooks.ts
```diff
@@ -34,54 +31,3 @@ export const useFingerprint = (nft: NFT | null) => {
 
   return [fingerprint, isLoading] as const
 }
-
-export const useComputedPrice = (nft: NFT, order: Order | null) => {
```
cazala, line 38: these utils were used for partners, that's why i got rid of them

webapp/src/modules/translation/locales/es.json
```diff
@@ -311,11 +314,17 @@
     "title": "Comprar {category}",
     "subtitle": "Estás por comprar {name} por {amount}.",
     "not_for_sale": "{name} no esta en venta.",
-    "price_too_high": "Dados los precios de cambio actuales, comprar {category} te costará un {percentageIncrease}% más de su precio normal",
-    "please_wait": "Por favor espere a que los precios se normalicen nuevamente",
-    "actual_price": "El precio arriba es una aproximación, tomando los precios de cambio actuales terminarás pagando al rededor de {computedPrice}",
     "buy": "Comprar",
-    "proceed_anyways": "Proceder",
```
cazala, line 318: all the copies above were only for partners

webapp/src/components/BuyPage/BuyNFTModal/BuyNFTModal.tsx
```diff
@@ -119,39 +109,6 @@ const BuyNFTModal = (props: Props) => {
         }}
       />
     )
-  } else if (isPartner(nft.vendor) && computedPrice) {
```
cazala, line 122: the following block was only for partners, so i removed it

webapp/src/components/BuyPage/MintItemModal/MintItemModal.tsx
```diff
@@ -82,13 +83,31 @@ const BuyItemModal = (props: Props) => {
...
   if (!item.isOnSale) {
-    subtitle = <T id={'buy_page.not_for_sale'} values={{ name }} />
+    subtitle = (
+      <T
+        id={'mint_page.not_for_sale'}
+        values={{
+          name,
+          secondary_market_link: (
+            <Link
+              to={locations.browse({
```
fzavalia, line 93: There is an issue here with the back buttons, I would suggest just dispatching a `goBack` here because if you direct to the item page from here, the back button on the item page will send you here and you will be stuck

fzavalia (approved):
Nice changes!

PR #507, opened 2021-12-30
fix: user stores responsiveness + transation history

### LINKED ISSUE #504: Make tables in store pages responsive
Make tables in store pages responsive

Make sidebar in mystore accessible (currently it disappears and there's no way to navigate)

cazala:
Closes #504

This PR:

- Fixes the responsiveness of the `Sales` and `On Sale` sections of `My Store`. 
- Since the `CurrentAccountSidebar` was completely removed on mobile, it was not possible to navigate the `My Store` page, so I added some `Tabs` to allow navigation on this view on mobile.
- Added the sale `type` to the sales information in `Sales`, `On Sale`, and `TransactionHistory`.
- I also took the chance to get rid of the deprecated `Responsive` component and replace it with `Mobile` and `NotMobile`. 
- Finally, I modified the `TransactionHistory` component to use the new `/sales` API, added pagination to it, and made it so it support both `NFT` and `Item`, so I added it to the `ItemDetail` as well:

My Store responsive tables:
Screen Shot 2021-12-30 at 12 31 43

ItemDetail with latest sales:
Screen Shot 2021-12-30 at 18 39 20

### DIFF
--- a/webapp/src/components/AssetPage/TransactionHistory/TransactionHistory.tsx
+++ b/webapp/src/components/AssetPage/TransactionHistory/TransactionHistory.tsx
@@ -1,21 +1,30 @@
 import { Link } from 'react-router-dom'
 import React, { useState, useEffect } from 'react'
-import { Bid, ListingStatus, Order } from '@dcl/schemas'
-import { Header, Table, Responsive } from 'decentraland-ui'
+import { Sale } from '@dcl/schemas'
+import {
+  Header,
+  Table,
+  Mobile,
+  NotMobile,
+  Pagination,
+  Loader,
+  Row
+} from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Profile } from 'decentraland-dapps/dist/containers'
 import dateFnsFormat from 'date-fns/format'
 
 import { Mana } from '../../Mana'
 import { locations } from '../../../modules/routing/locations'
-import { VendorFactory } from '../../../modules/vendor'
+import { saleAPI } from '../../../modules/vendor/decentraland'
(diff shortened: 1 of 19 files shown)


fzavalia:
image

When viewAsGuest is present you should not show the tabs for the current user. It should just show things as if it were a different user than the one connected.

fzavalia:
Seems to be also appearing when not in the my store page. For example, I'm in collectibles and can still see bids, settings and collections which should only be available in my store

image

cazala:
fixed @fzavalia 👍

webapp/src/components/AssetPage/TransactionHistory/TransactionHistory.tsx
```diff
@@ -28,56 +37,49 @@ const formatDateTitle = (updatedAt: number) => {
...
+      if ('tokenId' in asset) {
+        params.tokenId = asset.tokenId
+      } else {
+        params.itemId = asset.itemId
+      }
+      saleAPI
+        .fetch(params)
+        .then(response => {
+          setSales(response.data)
+          setTotalPages((response.total / ROWS_PER_PAGE) | 0)
```
fzavalia, line 66: Why the `| 0`?

fzavalia (approved):
LGTM!

webapp/src/components/AssetPage/TransactionHistory/TransactionHistory.tsx
cazala: it’s like Math.floor(), converts float to int (which removes the decimal part).

`2.5 | 0 // —> 2`

fzavalia: cool

PR #518, opened 2022-01-12
Deselect 'my store' tab when seeing a collection

### LINKED ISSUE #517: Deselect the `my store` tab when seeing a collection
By using this link https://market.decentraland.org/collections/0xb96697fa4a3361ba35b774a42c58daccaad1b8e1 we should deselect the `My store` tab by either do not select any tab or by selecting the collectibles one.

nishit-g:
Problem : 
By using this link https://market.decentraland.org/collections/0xb96697fa4a3361ba35b774a42c58daccaad1b8e1 we should deselect the My store tab by either do not select any tab or by selecting the collectibles one.

Solution : Selects the collectibles one.

closes #517

### DIFF
--- a/webapp/src/components/CollectionPage/CollectionPage.tsx
+++ b/webapp/src/components/CollectionPage/CollectionPage.tsx
@@ -41,7 +41,7 @@ const CollectionPage = (props: Props) => {
   return (
     <div>
       <Navbar isFullscreen />
-      <Navigation activeTab={NavigationTab.MY_STORE} />
+      <Navigation activeTab={NavigationTab.COLLECTIBLES} />
       <Page className={styles.page}>
         <CollectionProvider contractAddress={contractAddress} withItems>
           {({ collection, items, isLoading }) => {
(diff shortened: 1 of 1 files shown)


webapp/src/components/CollectionPage/CollectionPage.tsx
```diff
@@ -41,7 +41,7 @@ const CollectionPage = (props: Props) => {
   return (
     <div>
       <Navbar isFullscreen />
-      <Navigation activeTab={NavigationTab.MY_STORE} />
+      <Navigation activeTab={NavigationTab.COLLECTIBLES} />
```
cazala, line 44: Hi @nishit-g! The collection detail can be accessed from different parts of the app (like from an item detail, or from the collections section of My Store). This change in the PR makes it consistent when you navigate from an item detail to a collection detail (in both cases the Collectibles tab is active), but now it is inconsistent when you navigate from your collections' list in My Store (it switches back and forth between My Store and Collectibles).

I think the best solution is to turn this tab off in the collection detail rather than keeping Collectibles or My Store active.

```suggestion
      <Navigation />
```

cazala (approved):

cazala (changes requested):
@nishit-g there seems to be an error that's making the build fail: 

```
'NavigationTab' is declared but its value is never read. 
```

nishit-g:
@cazala sorry my bad, fixed it!

PR #531, opened 2022-01-21
feat: Add new Manage button to the name detail page

juanmahidalgo:
## Description
Adds a new button so the user can navigate to the builder /names page.

## Screenshot

image

### DIFF
--- a/webapp/src/components/AssetPage/Actions/Actions.tsx
+++ b/webapp/src/components/AssetPage/Actions/Actions.tsx
@@ -9,6 +9,7 @@ import { AssetType } from '../../../modules/asset/types'
 import { VendorFactory } from '../../../modules/vendor'
 import { Props } from './Actions.types'
 import styles from './Actions.module.css'
+import { builderUrl } from '../../../lib/environment'
 
 const Actions = (props: Props) => {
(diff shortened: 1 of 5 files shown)


webapp/src/components/AssetPage/Actions/Actions.tsx
```diff
@@ -105,6 +106,15 @@ const Actions = (props: Props) => {
           {t('asset_page.actions.transfer')}
         </Button>
       ) : null}
+      {isOwner && isENSName && (
+        <Button
+          as="a"
+          href={`${process.env.REACT_APP_BUILDER_URL}/names`}
```
fzavalia, line 112: Not an issue here, but maybe we should create some sort of config file in the market to handle environment variables instead of using them directly. We should create an issue for that. @cazala

fzavalia: we could store them in the `environment.ts` file

juanmahidalgo: Good idea @fzavalia! I added `builderUrl` to the `environment.ts` file :)

cazala (approved):

PR #537, opened 2022-01-24
feat: Prompt user when leaving My Store Settings page with unsaved changes

juanmahidalgo:
## Description

Uses react-router `Prompt` to prevent the router navigation and `onbeforeunload` to prevent closing/refreshing the tab when there are unsaved changes

https://user-images.githubusercontent.com/8763687/150797359-0bc9da04-37b9-4592-b210-abab9c2ae863.mov

### DIFF
--- a/webapp/src/modules/store/sagas.ts
+++ b/webapp/src/modules/store/sagas.ts
@@ -21,6 +21,8 @@ import {
   fetchStoreEntity,
   getStoreFromEntity
 } from './utils'
+import { Store } from './types'
+import { getIsLocalStoreDirty } from './selectors'
 
 export function* storeSaga(client: CatalystClient) {
   yield takeLatest(FETCH_STORE_REQUEST, handleFetchStoreRequest)
@@ -30,12 +32,16 @@ export function* storeSaga(client: CatalystClient) {
   function* handleLocationChange({
     payload: { location }
   }: LocationChangeAction) {
+    const isLocalStoreDirty: Store | null = yield select(getIsLocalStoreDirty)
+    if (!isLocalStoreDirty) {
+      return
+    }
+
     const address: string = yield select(getAddress)
-    if (
-      !location.search.includes('section=store_settings&') &&
-      !location.search.includes('viewAsGuest=true')
-    )
+    const allowed = ['section=store_settings&', 'viewAsGuest=true']
+    if (!allowed.some(value => location.search.includes(value))) {
       yield put(revertLocalStore(address))
+    }
   }
 
   function* handleFetchStoreRequest({
--- a/webapp/src/modules/store/sagas.ts
+++ b/webapp/src/modules/store/sagas.ts
@@ -21,7 +21,6 @@ import {
   fetchStoreEntity,
   getStoreFromEntity
 } from './utils'
-import { Store } from './types'
 import { getIsLocalStoreDirty } from './selectors'
 
 export function* storeSaga(client: CatalystClient) {
@@ -32,7 +31,7 @@ export function* storeSaga(client: CatalystClient) {
   function* handleLocationChange({
     payload: { location }
   }: LocationChangeAction) {
-    const isLocalStoreDirty: Store | null = yield select(getIsLocalStoreDirty)
+    const isLocalStoreDirty: boolean = yield select(getIsLocalStoreDirty)
     if (!isLocalStoreDirty) {
       return
     }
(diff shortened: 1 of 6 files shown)


cazala (approved):

fzavalia (changes requested):
Is it possible to reset the `localStore` property when the user leaves the settings? So this way when the user returns the changes are not there anymore

fzavalia:
> Is it possible to reset the `localStore` property when the user leaves the settings? So this way when the user returns the changes are not there anymore

Maybe listen to the LOCATION_CHANGE action of the connected react router on the store module sagas and if you are moving out of the settings page somewhere else than the view as guest page, dispatch the `REVERT_LOCAL_STORE` action

juanmahidalgo:
> > Is it possible to reset the `localStore` property when the user leaves the settings? So this way when the user returns the changes are not there anymore
> 
> Maybe listen to the LOCATION_CHANGE action of the connected react router on the store module sagas and if you are moving out of the settings page somewhere else than the view as guest page, dispatch the `REVERT_LOCAL_STORE` action

I tried what you suggested and the issue with that is that the `LOCATION_CHANGE` payload doesn't provide the previous location. 
The payload looks like: ```
 payload: {
    location,
    action,
    isFirstRendering,
  }
```
where `location` is the new one. That said, we'll be triggering the `LOCATION_CHANGE` on all the navigations across the page (not just the action of leaving my store settings) and we can't tell where the user came from. 
I'll explore trying to call the `onRevert` and discard the changes when the SettingsPage is unmounting and going to a route that is not the `viewAsGuest=true` one.

**Comment by juanmahidalgo** [comment]
> > > Is it possible to reset the `localStore` property when the user leaves the settings? So this way when the user returns the changes are not there anymore
> > 
> > 
> > Maybe listen to the LOCATION_CHANGE action of the connected react router on the store module sagas and if you are moving out of the settings page somewhere else than the view as guest page, dispatch the `REVERT_LOCAL_STORE` action
> 
> I tried what you suggested and the issue with that is that the `LOCATION_CHANGE` payload doesn't provide the previous location. The payload looks like:
> 
> ```
>  payload: {
>     location,
>     action,
>     isFirstRendering,
>   }
> ```
> 
> where `location` is the new one. That said, we'll be triggering the `LOCATION_CHANGE` on all the navigations across the page (not just the action of leaving my store settings) and we can't tell where the user came from. I'll explore trying to call the `onRevert` and discard the changes when the SettingsPage is unmounting and going to a route that is not the `viewAsGuest=true` one.

@fzavalia @cazala Handler added :)

File: webapp/src/modules/store/sagas.ts [hunk 1]
```diff
@@ -22,6 +25,18 @@ import {
...
+  function* handleLocationChange({
+    payload: { location }
+  }: LocationChangeAction) {
+    const address: string = yield select(getAddress)
+    if (
+      !location.search.includes('section=store_settings&') &&
+      !location.search.includes('viewAsGuest=true')
+    )
+      yield put(revertLocalStore(address))
+  }
```
fzavalia, line 39: In order to avoid the revert local store action to be logged everytime you could just check if the local store is not null

```suggestion
  function* handleLocationChange({
    payload: { location }
  }: LocationChangeAction) {
    const localStore: Store | null = yield select(getLocalStore)

    if (!localStore) {
      return
    }

    const address: string = yield select(getAddress)
    const allowed = ['section=store_settings&', 'viewAsGuest=true']

    if (!allowed.some(value => location.search.includes(value))) {
      yield put(revertLocalStore(address))
    }
  }
```

cazala: @fzavalia mm wouldn't it be possible that the localStore is assigned a previous store instead of a null one? according to the reducer:  https://github.com/decentraland/marketplace/blob/master/webapp/src/modules/store/reducer.ts#L103-L111

```
case REVERT_LOCAL_STORE: {
  const { address } = action.payload
  const previous = state.data[address]

  return {
    ...state,
    localStore: previous ? { ...previous } : null
  }
}
```

Tho we could create a selector to know if the local store is different to null and different to all the stores in state:

```ts
export const isLocalStoreDirty = createSelector<
  RootState,
  Record<string, Store>,
  Store | null,
  boolean
>(getData, getLocalStore, (storeData, localStore) => {
  return (
    !!localStore &&
    !Object.values(storeData).some(store => store === localStore)
  )
})
```

And we can use that from within the saga to exit before dispatching the action as you suggest

fzavalia: image

PR #551, opened 2022-01-27
feat: added skin category

cazala:
Added `skin` category to subgraph + webapp


webapp/src/modules/translation/locales/es.json
```diff
@@ -88,6 +88,7 @@
     "wearables_mask": "Mascaras",
     "wearables_tiara": "Tiara",
     "wearables_top_head": "Tope de la cabeza",
+    "wearables_skin": "Skin",
```
LautaroPetaccio, line 91: It seems that you're missing the category translation of skin.

webapp/src/components/HomePage/HomePage.tsx
```diff
@@ -47,7 +47,7 @@ const HomePage = (props: Props) => {
 
   const sort: Partial<Record<View, SortBy>> = useMemo(
     () => ({
-      [View.HOME_NEW_ITEMS]: SortBy.RECENTLY_REVIEWED,
+      [View.HOME_NEW_ITEMS]: SortBy.RECENTLY_LISTED,
```
LautaroPetaccio, line 50: Would you mind explaining the reason behind this change?

webapp/.ci/.env.stg
```diff
@@ -8,6 +8,7 @@ REACT_APP_ATLAS_SERVER_URL=https://api.decentraland.net
 REACT_APP_PEER_URL=https://peer-lb.decentraland.org
 REACT_APP_TRANSACTIONS_API_URL=https://transactions-api.decentraland.net/v1
 REACT_APP_BUILDER_URL=https://builder.decentraland.net
+REACT_APP_FF_SKINS=
```
LautaroPetaccio, line 11: Would you mind adding the FF in Vercel as well?

LautaroPetaccio (changes requested):

webapp/src/components/HomePage/HomePage.tsx
cazala: That's not related to this PR, it something that is currently broken in prod, if you click on the View More of the first slideshow, it takes you to the Collectibles page, with `sortBy=recently_reviewed` which is currently not a valid option for that page, so what happens is that the saga defaults to "recently_listed", but the dropdown in that page shows "Cheapest" as the selected option for some reason. With this change, the parameter is set to `sortBy=recently_listed` and the dropdown shows the right value "Recently Listed"

webapp/.ci/.env.stg
cazala: Good catch! I added it ✅

LautaroPetaccio (approved):

PR #554, opened 2022-01-31
feat: Add set item price and beneficiary

juanmahidalgo:
## Description

Replicates the same logic used in the `builder`  

### Collection list when it's not for sale

image

### `CollectionPage` page modal

image

### `ItemDetail` page

image


webapp/src/components/EditPriceAndBeneficiaryModal/EditPriceAndBeneficiaryModal.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useState } from 'react'
+// import { Address } from 'web3x/address'
+import { fromWei, toWei } from 'web3x/utils'
```
cazala, line 3: Could we remove the usage of web3x from here and instead use `ethers`? 

Instead of `toWei(amount)` you can do `utils.parseEther(amount).toString()`, and instead of `fromWei(amount)` you can do `utils.formatEther(amount)`

webapp/src/components/EditPriceAndBeneficiaryModal/EditPriceAndBeneficiaryModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const handleIsNotForSaleToggle = () => {
+    setItemProps(prevItemProps => ({
+      beneficiary: prevItemProps.isFree ? undefined : item.creator,
+      price: prevItemProps.isFree ? undefined : '0',
+      isFree: !prevItemProps.isFree
+    }))
+  }
+
+  const isGift = itemProps.beneficiary !== item.creator
```
cazala, line 60: Should we rename this to `isNotOnSale` or something more representative? Or maybe `isOnSale` and then use it with a `!` below.

webapp/src/components/EditPriceAndBeneficiaryModal/EditPriceAndBeneficiaryModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+      Number(numberPrice) > 0 ||
+      (numberPrice === 0 && beneficiary === item.creator)
+    )
+  }
+
+  const isButtonDisabled = () => {
+    return !isValidPrice() || !isValidBeneficiary() || isLoading
+  }
+
+  const { price, isFree, beneficiary } = itemProps
```
cazala, line 101: Same as above, the `isFree` might need to be renamed, maybe `isDisabled`?

webapp/src/modules/item/sagas.ts
```diff
@@ -17,13 +17,64 @@ import {
...
+    const metadata = getMetadata(newItem)
+    const chainId: ChainId = yield call(getChainIdByNetwork, Network.MATIC)
+    const contract = {
+      ...getContract(ContractName.ERC721CollectionV2, chainId),
+      address: item.contractAddress!
+    }
+    const newPrice = newItem.price === '0' ? constants.MaxUint256.sub(1) : newItem.price!;
+    const txHash: string = yield call(sendTransaction, contract, collection =>
+      collection.editItemsData(
+        [newItem.itemId!],
```
cazala, line 65: i think this `!` is not needed

webapp/src/modules/item/sagas.ts
```diff
@@ -17,13 +17,64 @@ import {
...
+
+    const newItem = { ...item, price, beneficiary, updatedAt: Date.now() }
+
+    const metadata = getMetadata(newItem)
+    const chainId: ChainId = yield call(getChainIdByNetwork, Network.MATIC)
+    const contract = {
+      ...getContract(ContractName.ERC721CollectionV2, chainId),
+      address: item.contractAddress!
+    }
+    const newPrice = newItem.price === '0' ? constants.MaxUint256.sub(1) : newItem.price!;
```
cazala, line 62: Is this is this `.sub(1)` necessary? I tried this locally without it and it worked fine. Also, the server is checking against `2 ** 256 - 1`, and `MaxUint256` should already be that number (it already includes the `-1` so shouldn't need that extra `.sub(1)`)

webapp/src/components/AssetPage/ItemDetail/ItemDetail.tsx
```diff
@@ -30,77 +32,87 @@ const ItemDetail = ({ item, wallet }: Props) => {
...
+              <Network asset={item} />
+            </div>
+            {isOwner ? (
+              <div className={styles.ownerButtons}>
+                <Button fluid onClick={() => setShowEditPriceModal(true)}>
+                  {t('asset_page.actions.edit_price')}
+                </Button>
+                <Button as="a" href={builderCollectionUrl} fluid>
+                  {t('asset_page.actions.change_beneficiary')}
+                </Button>
```
cazala, line 84: Should we remove this `change beneficiary` button? It currently takes you to the builder, but I think the edit price button above is enough and we can remove this one.

cazala (changes requested):
A few minor comments but it's looking lit 🔥

webapp/src/components/EditPriceAndBeneficiaryModal/EditPriceAndBeneficiaryModal.tsx
cazala: nvm i confused this variable with the `isFree` one

cazala (approved):
🤸

PR #571, opened 2022-03-01
fix: Avatars not rendering correctly

fzavalia:
Includes fix to display newly created avatars after catalyst changes


nachomazzara (approved):
LGTM

PR #619, opened 2022-04-27
feat: change button for link in pagination

braianj:
In order to perform this task, we took out three saga (buildBrowseURL, getCurrentBrowseOptions, getAddress) and put them in utils (buildBrowseURL) and the rest in selectors. 
As part of the process we created selectors that group other selectors, mainly the ones that get parameters from the url (getPaginationUrlParams, getAssetsUrlParams, getLandsUrlParams, getWearablesUrlParams), so we can use them in the getCurrentBrowseOptions selector.
Once this process was finished, the sagas that use these new selectors were updated.
Then we added in the AssetList.container.ts file the urlNext property to send the url to use in the href of the anchor tag.

### DIFF
--- a/webapp/src/components/AssetList/AssetList.container.ts
+++ b/webapp/src/components/AssetList/AssetList.container.ts
@@ -17,19 +17,26 @@ import { MapStateProps, MapDispatch, MapDispatchProps } from './AssetList.types'
 import AssetList from './AssetList'
 import { FETCH_ITEMS_REQUEST } from '../../modules/item/actions'
 import { buildBrowseURL } from '../../modules/routing/utils'
+import { getLocation } from 'connected-react-router'
 
-const mapState = (state: RootState): MapStateProps => ({
-  vendor: getVendor(state),
-  assetType: getAssetType(state),
-  nfts: getNFTs(state),
-  items: getItems(state),
-  page: getPage(state),
-  count: getCount(state),
-  isLoading:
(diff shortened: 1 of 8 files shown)


webapp/src/components/AssetList/AssetList.container.ts
```diff
@@ -25,7 +27,8 @@ const mapState = (state: RootState): MapStateProps => ({
   count: getCount(state),
   isLoading:
     isLoadingType(getLoadingNFTs(state), FETCH_NFTS_REQUEST) ||
-    isLoadingType(getLoadingItems(state), FETCH_ITEMS_REQUEST)
+    isLoadingType(getLoadingItems(state), FETCH_ITEMS_REQUEST),
+  urlNext: buildBrowseURL(state.router.location.pathname, { ...getCurrentBrowseOptions(state), page: (getPage(state) + 1)})
```
2fd, line 31: There are too many thing in only one line, try to move some functions call into variables to make it a little more readable, and don't repeat the call to `getPage`.

And access the `state` variable directly can be consider a bad practice, it could change at any given time and the only place expected to change how is access is in selectors, use the `getLocation` (from `connected-react-router` dependency) to access the location

2fd (changes requested):

2fd (approved):

cazala (approved):

cazala:
Awesome 👏 👏 👏

PR #623, opened 2022-05-06
fix: min sale value as an env var

nicosantangelo:
Min sale value as an ENV variable:
`RaaEACT_APP_MIN_SALE_VALUE_IN_WEI`


webapp/src/components/BuyPage/BuyPage.tsx
```diff
@@ -34,6 +35,17 @@ const BuyPage = (props: Props) => {
           {wallet => (
             <AssetProviderPage type={type}>
               {(asset, order) => {
+                if (type === AssetType.ITEM) {
+                  const { price } = asset as Item
+                  const minValueInWei = Number(
+                    process.env.REACT_APP_MIN_SALE_VALUE_IN_WEI
```
fzavalia, line 41: Should we have a default value for this?

fzavalia (approved):

webapp/src/components/BuyPage/BuyPage.tsx
```diff
@@ -34,6 +35,17 @@ const BuyPage = (props: Props) => {
           {wallet => (
             <AssetProviderPage type={type}>
               {(asset, order) => {
+                if (type === AssetType.ITEM) {
+                  const { price } = asset as Item
+                  const minValueInWei = Number(
```
fzavalia, line 40: This could be directly be converted to a BigNumber instead of being converted to Number first

fzavalia (approved):

PR #641, opened 2022-05-10
feat: price too low for items

nicosantangelo:
https://user-images.githubusercontent.com/692077/167705844-336e957d-803d-4c6a-b95f-c12c57722181.mov

Related: https://github.com/decentraland/dapps-issues/issues/23


webapp/src/modules/translation/locales/en.json
```diff
@@ -30,6 +30,7 @@
     "sale_type": "Sale type",
     "sell_price": "Sell price",
     "no_results": "No results",
+    "learn_more": "Learn More",
```
LautaroPetaccio, line 33: This change seems to not have been replicated in the other languages.

webapp/src/components/BuyPage/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Checks against the min value of costless item sales
+ */
+export function isPriceTooLow(type: AssetType, asset: Asset) {
+  if (type === AssetType.ITEM) {
+    const { price } = asset as Item
+    const minSaleValue = getMinSaleValueInWei()
+    if (minSaleValue) {
+      return toBN(price).gte(toBN(minSaleValue))
```
LautaroPetaccio, line 13: Isn't it lower than?
```suggestion
      return toBN(price).lt(toBN(minSaleValue))
```

webapp/src/components/BuyPage/utils.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    beforeEach(() => {
+      process.env.REACT_APP_MIN_SALE_VALUE_IN_WEI = ''
+    })
+
+    it('should return true', () => {
+      expect(isPriceTooLow(AssetType.ITEM, {} as Item)).toBe(true)
+    })
+  })
+
+  describe('when there a min sale value', () => {
```
LautaroPetaccio, line 41:
```suggestion
  describe('when there is a min sale value', () => {
```

webapp/src/components/BuyPage/utils.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    beforeEach(() => {
+      process.env.REACT_APP_MIN_SALE_VALUE_IN_WEI = minSaleValue
+    })
+
+    describe('and the asset type is NFT', () => {
+      it('should return true', () => {
+        expect(isPriceTooLow(AssetType.NFT, {} as Item)).toBe(true)
+      })
+    })
```
LautaroPetaccio, line 52: Are we just covering minting of new items with this change?

webapp/src/components/BuyPage/utils.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    describe('and the asset type is ITEM', () => {
+      it('should return true if the price is greater than the minimum', () => {
+        expect(
+          isPriceTooLow(AssetType.ITEM, {
```
LautaroPetaccio, line 57: The name might be a bit weird, if the price is too low, why is it returning true for a min price that is lower than the price? It seems that the price is higher than the minimum amount.

webapp/src/components/BuyPage/MintItemModal/MintItemModal.tsx
```diff
@@ -77,6 +79,9 @@ const MintItemModal = (props: Props) => {
     setShowAuthorizationModal
   ])
 
+  const isInDifferentChainId = wallet.chainId !== item.chainId
+  const hasLowPrice =
+    !isPriceTooLow(AssetType.ITEM, item) && isInDifferentChainId
```
LautaroPetaccio, line 84: Isn't the condition `isPriceTooLow(AssetType.ITEM, item) && !isInDifferentChainId`?

LautaroPetaccio (changes requested):

webapp/src/modules/translation/locales/en.json
nicosantangelo: Good eye!

webapp/src/components/BuyPage/utils.ts
nicosantangelo: Correct, the function is actually fine, the refactored name was the problem. Changing tho!

webapp/src/components/BuyPage/MintItemModal/MintItemModal.tsx
nicosantangelo: fixed, related to the rename discussed above

webapp/src/components/BuyPage/utils.spec.ts
nicosantangelo: fixed, related to the rename discussed above

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #644, opened 2022-05-12
Feat/slider update

### LINKED ISSUE #627: Update Slideshow Component
We need to update the current Slideshow component so it more closely resembles the new design.

To complete this issue:
- Add arrows on hover to move the slider horizontally to the left or to the right
- Add the pages, highlighting the current one, at the bottom of the component
- Make it loop when reaching the end

juanmahidalgo:
Closes #627

### New look

https://user-images.githubusercontent.com/8763687/168136238-38875f7d-d123-441d-8300-04677267c415.mov

### DIFF
--- a/webapp/src/components/HomePage/Slideshow/Slideshow.tsx
+++ b/webapp/src/components/HomePage/Slideshow/Slideshow.tsx
@@ -11,7 +11,7 @@ const INITIAL_PAGE = 1
 const Slideshow = (props: Props) => {
   const slideRef = useRef<HTMLDivElement>(null)
   const { title, assets, isSubHeader, isLoading, onViewAll } = props
-  const [showArrows, setShowArrows] = useState(true)
+  const [showArrows, setShowArrows] = useState(false)
   const [currentPage, setCurrentPage] = useState(INITIAL_PAGE)
   const [assetsToRender, setAssetsToRender] = useState(
     assets.slice(0, PAGE_SIZE)
@@ -45,15 +45,20 @@ const Slideshow = (props: Props) => {
     setCurrentPage(currentPage - 1 === 0 ? totalPages : currentPage - 1)
   }
 
-  const showArrowsHandlers = {
-    onMouseEnter: () => setShowArrows(true),
-    onMouseLeave: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
-      const contained = slideRef.current?.contains?.(
-        event.relatedTarget as Node
-      )
-      !contained && setShowArrows(false)
-    }
-  }
+  const showArrowsHandlers = useMemo(
+    () => ({
+      onMouseEnter: () => setShowArrows(true),
+      onMouseLeave: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+        const contained =
+          event.relatedTarget instanceof Node &&
+          slideRef.current?.contains(event.relatedTarget as Node)
+        if (!contained) {
+          setShowArrows(false)
+        }
+      }
+    }),
+    []
+  )
 
   return (
     <div className="Slideshow" ref={slideRef} {...showArrowsHandlers}>
--- a/webapp/src/components/HomePage/Slideshow/Slideshow.tsx
+++ b/webapp/src/components/HomePage/Slideshow/Slideshow.tsx
@@ -45,20 +45,20 @@ const Slideshow = (props: Props) => {
     setCurrentPage(currentPage - 1 === 0 ? totalPages : currentPage - 1)
   }
 
-  const showArrowsHandlers = useMemo(
-    () => ({
-      onMouseEnter: () => setShowArrows(true),
-      onMouseLeave: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
-        const contained =
-          event.relatedTarget instanceof Node &&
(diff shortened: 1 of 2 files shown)


webapp/src/components/Navigation/Navigation.tsx
```diff
@@ -13,6 +13,11 @@ const Navigation = (props: Props) => {
   return (
     <Tabs isFullscreen={isFullscreen}>
       <Tabs.Left>
+        <Link to={locations.root()}>
+          <Tabs.Tab active={activeTab === NavigationTab.LANDS}>
```
nicosantangelo, line 17: careful when rebasing here

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
```diff
@@ -1,18 +1,62 @@
...
+
+  const showArrowsHandlers = {
+    onMouseEnter: () => setShowArrows(true),
+    onMouseLeave: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      const contained = slideRef.current?.contains?.(
+        event.relatedTarget as Node
+      )
+      !contained && setShowArrows(false)
+    }
+  }
```
nicosantangelo, line 56: Shouldn't all these use `useCallback` ?

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
```diff
@@ -1,18 +1,62 @@
...
+    setCurrentPage(currentPage - 1 === 0 ? totalPages : currentPage - 1)
+  }
+
+  const showArrowsHandlers = {
+    onMouseEnter: () => setShowArrows(true),
+    onMouseLeave: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      const contained = slideRef.current?.contains?.(
+        event.relatedTarget as Node
+      )
+      !contained && setShowArrows(false)
```
nicosantangelo, line 54: maybe we could just write an if here to be more explicit?

```ts
if(!isContained) {
  setShowArrows(false)
}
```

nicosantangelo (commented):
Two things:

- Alt-tabing in and out the browser and hovering around yield me this error:
image

- The arrows should not be visible in the first load. It works correctly after the first hover

- You think we could add an animation to the change? I know that, to do that, we should move the right/left position attributes instead of changing the current "page", so we can just create an issue for that

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
```diff
@@ -1,18 +1,62 @@
...
+  ])
+
+  useEffect(() => {
+    const currentPosition = (currentPage - INITIAL_PAGE) * PAGE_SIZE
+    setAssetsToRender(
+      assets.slice(currentPosition, currentPosition + PAGE_SIZE)
+    )
+  }, [currentPage, assets, setAssetsToRender])
+
+  const renderNfts = useCallback(
```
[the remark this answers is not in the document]
juanmahidalgo: Could be both.. as it was a function before, I use `useCallback` to keep it that way. Do you prefer `useMemo` in this case?

webapp/src/components/Navigation/Navigation.tsx
juanmahidalgo: done :)

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
juanmahidalgo: mhmm.. I can't add `useCallback` inside the object. I could move them to the base level but I wanted to destructure them in the React components as I'm doing atm. I wrapped the `showArrowsHandlers` with a `useMemo` so it's created just once during the mounting. I think that's pretty close to what you said

juanmahidalgo:
> Two things:
> 
> * Alt-tabing in and out the browser and hovering around yield me this error:
> 
> image
> 
> * The arrows should not be visible in the first load. It works correctly after the first hover
> * You think we could add an animation to the change? I know that, to do that, we should move the right/left position attributes instead of changing the current "page", so we can just create an issue for that

Ohhh, good catch, I didn't see that error. I added some code to fix it. I also hide the arrows during the first render.
Agreed with the animation, let's add it in a follow up PR!

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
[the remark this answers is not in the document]
nicosantangelo: I don't have a preference really! I asked because I associate useCallbacks with functions that act as a handler (or a callback heh) of an action that could be triggered later
And seeing this as a map that changes with the list, I thought it looked like a memoized value

webapp/src/components/HomePage/Slideshow/Slideshow.tsx
nicosantangelo: Totally! in my mind it could look like this too:
```ts
  const onMouseEnter = useCallback(() => {}, [])
  const onMouseLeave = useCallback(() => {}, [])
  const handlers = { onMouseEnter, onMouseLeave }

  return (
    <div className="Slideshow" ref={slideRef} {...handlers}>
```

but both are effectively the same I think

nicosantangelo (approved):

PR #657, opened 2022-05-24
feat: Add handlers to track VolumeDayData

juanmahidalgo:


indexer/src/handlers/marketplace.ts
```diff
@@ -87,13 +88,19 @@ export function handleOrderSuccessful(event: OrderSuccessful): void {
   nft = updateNFTOrderProperties(nft!, order!)
   nft.save()
 
+  // Bind contract
+  let marketplaceContract = Marketplace.bind(event.address)
+  // On ropsten we hace a different version of the marketplace where the method is ownerCutPercentage
```
nicosantangelo, line 93:
```suggestion
  // On ropsten we have a different version of the marketplace where the method is ownerCutPercentage
```

😅

indexer/src/handlers/marketplace.ts
```diff
@@ -87,13 +88,19 @@ export function handleOrderSuccessful(event: OrderSuccessful): void {
...
+  let ownerCutPerMillion = marketplaceContract.try_ownerCutPerMillion()
```
nicosantangelo, line 94: What if we use something [like this](https://github.com/decentraland/collections-graph/blob/master/src/modules/Network/index.ts) to get the current network (the ternary is not necessary) and use it to call the correct implementation ?
is that possible?

indexer/src/modules/analytics/index.ts
```diff
@@ -59,4 +61,38 @@ export function trackSale(
...
+  let volumeDayData = updateVolumeDayData(sale, timestamp, feesCollectorCut)
+  volumeDayData.save()
+}
+
+function getOrCreateAnalyticsDayData(blockTimestamp: BigInt): AnalyticsDayData {
+  let timestamp = blockTimestamp.toI32()
+  let dayID = timestamp / 86400 // unix timestamp for start of day / 86400 giving a unique day index
+  let dayStartTimestamp = dayID * 86400
+  let volumeDayData = AnalyticsDayData.load(dayID.toString())
+  if (volumeDayData === null) {
```
nicosantangelo, line 74: I think this works but because we're using the old AssemblyScript verision ([0.0.3 on .yaml](https://thegraph.com/docs/en/developer/assemblyscript-migration-guide/#nullability)) I think we should keep `==`

We can [update it later](https://thegraph.com/docs/en/developer/assemblyscript-migration-guide/#how-to-upgrade)

nicosantangelo (commented):
Awesome

indexer/src/handlers/marketplace.ts
juanmahidalgo: I tried what you suggested but couldn't get around the fact that the right method is not typed in any of the generated types and we can't just forced it to be `any` :( 
image

indexer/src/modules/analytics/index.ts
juanmahidalgo: ohh good catch, thanks!

cazala (approved):

PR #761, opened 2022-07-22
feat: Round up sales per day stat

### LINKED ISSUE #760: Round up the sales per day stat
## 🎉 Description

Let's round up the Number of sales per day introduced in the Volume component that was released with Rankings.

* [X] 🐛 This is a bug report.

juanmahidalgo:
image

Closes #760

### DIFF
--- a/webapp/src/components/AnalyticsVolumeDayData/utils.ts
+++ b/webapp/src/components/AnalyticsVolumeDayData/utils.ts
@@ -14,9 +14,9 @@ export function formatDailySales(sales: number, timeframe: AnalyticsTimeframe) {
   if (!sales) return '0'
   switch (timeframe) {
     case AnalyticsTimeframe.WEEK:
-      return `${Number((sales / 7).toFixed(2))}/day`
+      return `${Math.round(Number(sales / 7))}/day`
     case AnalyticsTimeframe.MONTH:
-      return `${Number(sales / 30).toFixed(2)}/day`
+      return `${Math.round(Number(sales / 30))}/day`
     default:
       return ''
   }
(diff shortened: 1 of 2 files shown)


webapp/src/components/AnalyticsVolumeDayData/utils.ts
```diff
@@ -14,9 +14,9 @@ export function formatDailySales(sales: number, timeframe: AnalyticsTimeframe) {
   if (!sales) return '0'
   switch (timeframe) {
     case AnalyticsTimeframe.WEEK:
-      return `${Number((sales / 7).toFixed(2))}/day`
+      return `${Math.round(Number(sales / 7))}/day`
```
fzavalia, line 17: This will round up if the decimals is > .5. Is that intended?
If not use `Math.trunc` or `Number(sales / 7) << 0`

fzavalia, line 17: No need to `Number()` if you are not using `toFixed`

fzavalia, line 17: This will round up decimals > .5
Is that ok?
If not you can use `Math.trunc` or `<< 0` to remove the decimals

juanmahidalgo: good call! updated 👍

juanmahidalgo: yeah, I believe we want to round up to the closest integer

fzavalia (approved):

PR #768, opened 2022-08-03
feat: fat finger for buys

nicosantangelo:


nicosantangelo:
Closed in favor of a different route (for now)

PR #769, opened 2022-08-11
feat: create rental

### LINKED ISSUE #776: Rental configuration step for the List for rent modal
## 🎉 Description

To complete this task, the following must be done:
- Create the rental configuration step as described in the pictures below
- Create the sagas, reducers and other methods to support signing a rental.
- Create the mechanisms to authorize the rental contract to operate with LAND.

Modal requirements:
- The user has to input a price in MANA per day.
- The user has to select any of the rent period options available. This is a multiple selection.
- The user can select the expiration of the listing but a default of 1 month from now will be selected automatically.
- Once all input fields are completed by the user, the list for rent button is enabled.
- Clicking the list for rent button will take you to the third view.

cazala:
Closes #776 

This PR adds the `CreateRentalModal`. This modal has 2 views, one to approve the Rental contract if necessary, and the second one to create the rental itself. I added a button to the ParcelDetail to be able to open the rental modal and test it until we have the new management page for land. This button is hidden behind a feature flag.

Screen Shot 2022-08-18 at 17 07 30

Screen Shot 2022-08-18 at 17 01 37


webapp/src/components/AssetPage/ParcelDetail/ParcelDetail.tsx
```diff
@@ -73,6 +77,16 @@ const ParcelDetail = ({ nft }: Props) => {
         <>
           <BidList nft={nft} />
           <TransactionHistory asset={nft} />
+          {isRentalsEnabled ? (
+            <Button onClick={() => setShowCreateRentalModal(true)}>
+              Create Rental
```
cazala, line 82: I added this button here to be able to test this modal from the Parcel detail page.

webapp/src/index.css
```diff
@@ -33,3 +33,8 @@ code {
     margin-left: 0px;
   }
 }
+
+body
+  > iframe[style='position: fixed; top: 0px; left: 0px; width: 100%; height: 100%; border: none; z-index: 2147483647;'] {
```
cazala, line 38: for some reason, every time I save a file and the marketplace is rebuilt, I get a `process is not defined` error in the console and then an invisible iframe blocks the whole screen.. i could figure out a way to fix the error, but this rule at least gets rid of that invisible iframe so I can keep working without having to reload the page

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+    if (shouldAdd) {
+      setPeriodOptions([...periodOptions, periodOption])
+    } else {
+      setPeriodOptions(periodOptions.filter(option => option !== periodOption))
+    }
+  }
+
+  // validation
+  const isInvalidPrice = parseMANANumber(pricePerDayInput) <= 0
+  const isInvalidDate = new Date(expiresAt).getTime() < Date.now()
```
LautaroPetaccio, line 93: What do you think about adding `expiration` to this variable name?
```suggestion
  const isInvalidExpirationDate = new Date(expiresAt).getTime() < Date.now()
```

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+              label={t('create_rental_modal.price')}
+              type="text"
+              placeholder={1000}
+              network={nft.network}
+              value={pricePerDayInput}
+              focus={true}
+              error={pricePerDayInput !== '' && isInvalidPrice}
+              onChange={(_event, props) => {
+                setPricePerDayInput(toFixedMANAValue(props.value))
+              }}
```
LautaroPetaccio, line 117: What if we define a handler for these functions so we don't have to re-render the field when a re-render is triggered?

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+    onCreate,
+    nft,
+    pricePerDayInput,
+    periodOptions,
+    expiresAt,
+    onGrant,
+    authorization
+  ])
+
+  const createOptionHanlder = (periodOption: PeriodOption) => () => {
```
LautaroPetaccio, line 82: Would you mind changing this function to:
```suggestion
  const handlePeriodsOptionChange = (periodOption: PeriodOption) => () => {
```
so it follows the dApps standards.

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+        {isAuthorized ? (
+          <>
+            <ManaField
+              label={t('create_rental_modal.price')}
+              type="text"
+              placeholder={1000}
+              network={nft.network}
+              value={pricePerDayInput}
+              focus={true}
+              error={pricePerDayInput !== '' && isInvalidPrice}
```
LautaroPetaccio, line 114: There's a message in case an invalid date is set, but, there's no message for an invalid price, what do you think about setting one?

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+              <div className="period-options">
+                {Object.values(PeriodOption).map(option => (
+                  <Radio
+                    label={t(`create_rental_modal.period_options.${option}`)}
+                    checked={periodOptions.includes(option)}
+                    onClick={createOptionHanlder(option)}
+                  />
+                ))}
+              </div>
```
LautaroPetaccio, line 153: Should we add a kind of error message saying that the user needs to set at least one period?

LautaroPetaccio (changes requested):

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: the thing is that the function itself is not a handler, is currying a handler with the period option, and return the handler itself.

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: but should that error be shown all the time until the user selects a period? currently the 'put for rent' button is disabled if there's no period selected, so the user can't click it

webapp/src/index.css
juanmahidalgo: @cazala try running `npm install react-error-overlay@6.0.9 --save-dev`. If it gets fixed, feel free to push the `-lock` here :)

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
LautaroPetaccio: I think it might be a bit annoying, you're right. I was trying to find a solution to the problem of "disabled buttons that don't show why", but this might be overkill.

webapp/src/index.css
cazala: it worked! thanks :)

webapp/src/modules/rental/types.ts
```diff
@@ -0,0 +1,8 @@
+export enum PeriodOption {
+  ONE_WEEK = 'one_week',
+  ONE_MONTH = 'one_month',
+  TWO_MONTHS = 'two_months',
+  QUARTER_YEAR = 'quarter_year',
+  HALF_YEAR = 'half_year',
+  ONE_YEAR = 'one_year'
```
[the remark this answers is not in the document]
cazala: As discussed offline, didn't take this approach because using numbers as values for enums causes them to have 2 entries per value in the final record (typescript implementation), which makes it more difficult to get a list of all the values

webapp/src/modules/rental/sagas.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const signature: string = yield call(
+      getSignature,
+      nft.chainId,
+      nft.contractAddress,
+      nft.tokenId,
+      nonces,
+      periods,
+      expiresAt
+    )
```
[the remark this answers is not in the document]
cazala: done 👍

webapp/src/modules/rental/reducer.ts
```diff
@@ -0,0 +1,66 @@
...
+  | CreateRentalRequestAction
+  | CreateRentalSuccessAction
+  | CreateRentalFailureAction
+
+export function rentalReducer(
+  state = INITIAL_STATE,
+  action: RentalReducerAction
+): RentalState {
+  switch (action.type) {
+    case CREATE_RENTAL_REQUEST: {
```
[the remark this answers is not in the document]
cazala: done 👍

webapp/src/modules/rental/utils.ts
```diff
@@ -0,0 +1,139 @@
...
+  return nonce.toString()
+}
+
+export async function getSignerNonce(chainId: ChainId, signerAddress: string) {
+  const rentals = await getRentalsContractInstance(chainId)
+  const nonce: BigNumber = await rentals.signerNonce(signerAddress)
+  return nonce.toString()
+}
+
+export async function getContractNonce(chainId: ChainId) {
```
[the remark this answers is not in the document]
cazala: done 👍

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: done 👍

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: done 👍

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: done 👍

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
cazala: mm it has a `value` but how would you use the same handler to set the values of different Radio components?

webapp/src/components/CreateRentalModal/CreateRentalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+import { ContractName, getContract } from 'decentraland-transactions'
+import { TransactionLink } from 'decentraland-dapps/dist/containers'
+import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { toFixedMANAValue } from 'decentraland-dapps/dist/lib/mana'
+import { PeriodOption } from '../../modules/rental/types'
+import { ManaField } from '../ManaField'
+import { parseMANANumber } from '../../lib/mana'
+import { getDefaultExpirationDate } from '../../modules/order/utils'
+import { Props } from './CreateRentalModal.types'
+import './CreateRentalModal.css'
```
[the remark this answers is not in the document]
cazala: AFAIK it's not possible to theme components (like the Radios) using css modules, since it will screw up all the class names of the decentraland-ui components i'm trying to theme.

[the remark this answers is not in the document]
cazala: for example: https://github.com/decentraland/marketplace/pull/769/files#diff-33b52f92e09f2ffd9b051dde3817f0f24e6965eb91c1b7794f13014e487cb9e7R38-R52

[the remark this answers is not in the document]
cazala: nvm, i forgot about `:global` for these cases, done 👍

webapp/src/modules/rental/contract.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+  describe('and the provider is connected', () => {
+    beforeEach(async () => {
+      getConnectedProviderMock.mockResolvedValueOnce(providerMock)
+    })
+    it('should return an instance', () => {
+      expect(
+        getRentalsContractInstance(ChainId.ETHEREUM_GOERLI)
+      ).resolves.toBeTruthy()
+    })
```
LautaroPetaccio, line 35: There's a problem with this test, it checks for a truthy value, but the method could anything and it will still pass. You would have to mock ethers or directly remove the test if it's too much of a problem as this test could be a bit tricky to understand.

webapp/src/modules/rental/reducer.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      error: null
+    }
+    const newState = rentalReducer(state, createRentalSuccess(nft, rental))
+    expect(newState.data).toEqual({
+      [rental.id]: rental
+    })
+  })
+})
+
+describe('when reducing a CREATE_RENTAL_SUCCESS action', () => {
```
LautaroPetaccio, line 76:
```suggestion
describe('when reducing a CREATE_RENTAL_FAILURE action', () => {
```

webapp/src/modules/rental/reducer.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  it('should remove the loading action', () => {
+    const state: RentalState = {
+      loading: [
+        createRentalRequest(nft, 100, [PeriodOption.ONE_WEEK], 1976562675847)
+      ],
+      data: {},
+      error: null
+    }
```
LautaroPetaccio, line 50: You can abstract all state definitions by adding a `beforeEach` on top of the describe scope with the following:
```
let state
beforeEach(() => {
  state = {
        loading: [
        createRentalRequest(nft, 100, [PeriodOption.ONE_WEEK], 1976562675847)
      ],
      data: {},
      error: 'some error'
  }
})
```
This state should be functional to all `its` down below.

webapp/src/modules/rental/utils.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+  beforeEach(() => {
+    getRentalsContractInstanceMock.mockResolvedValue(rentalsMock)
+  })
+  afterEach(() => {
+    getRentalsContractInstanceMock.mockReset()
+  })
+  it('should return the contract nonce, signer nonce and asset nonce using the correct chain id, contract address, token id and signer address', async () => {
+    rentalsMock.contractNonce.mockResolvedValueOnce(BigNumber.from(0))
+    rentalsMock.signerNonce.mockResolvedValueOnce(BigNumber.from(1))
+    rentalsMock.assetNonce.mockResolvedValueOnce(BigNumber.from(2))
```
LautaroPetaccio, line 184: Would you mind moving these mocks to the `beforeEach`? The idea behind this is to build the context on it so the it is used to define expectations.

LautaroPetaccio (approved):

webapp/src/modules/rental/reducer.spec.ts
cazala: thanks <3

webapp/src/modules/rental/contract.spec.ts
cazala: I changed it to expect the resolved value to be an instance of `ethers.Contract`

webapp/src/modules/rental/reducer.spec.ts
cazala: done 👍

webapp/src/modules/rental/utils.spec.ts
cazala: done 👍

PR #772, opened 2022-08-16
feat: Add rental price to land card

### LINKED ISSUE #755: Add price per day for listed lands/estates in their cards

LautaroPetaccio:
Closes #755

### DIFF
--- a/webapp/src/components/AssetCard/AssetCard.container.ts
+++ b/webapp/src/components/AssetCard/AssetCard.container.ts
@@ -1,26 +1,39 @@
 import { connect } from 'react-redux'
 import { RootState } from '../../modules/reducer'
 import { getData } from '../../modules/order/selectors'
+import { isLand } from '../../modules/nft/utils'
 import { getActiveOrder } from '../../modules/order/utils'
 import { MapStateProps, OwnProps, MapDispatchProps } from './AssetCard.types'
 import AssetCard from './AssetCard'
 import { getView } from '../../modules/ui/browse/selectors'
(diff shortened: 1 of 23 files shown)


webapp/src/components/AssetCard/AssetCard.container.ts
```diff
@@ -1,26 +1,38 @@
...
   }
 
-  const view = getView(state)
-  const price = getAssetPrice(asset, order)
+  let rentalPricePerDay: string | null = null
+  const openRentalId = getOpenRentalId(asset)
+
+  if (!rental && openRentalId && isLand(asset)) {
+    // The price per day is the same in this version of rentals
+    rentalPricePerDay = getRentalById(state, openRentalId)!.periods[0]
```
cazala, line 29: im thinking.. even tho it will not be possible through the UI, what if I create a land with a pricePerDay of 0 (or a very small number if cero is not valid) for a 7 day period, and a very expensive price of the 30 day period. In the UI this is going to show that the price per day is 0 or a very small number, and when I go ahead to rent it, I will be able to do it for 7 or 30 days, and it will show the same price for both periods (the cheap price). Then I go ahead and send the tx and it I'll pay the expensive price.. I would validate here that all the periods have the same price and ignore it otherwise as if the rental was `null`.

LautaroPetaccio: I've changed the code to get the maximum price of the list of periods so if there's a period with a greater price, it will show it instead of the price of the first one.

cazala (approved):

PR #797, opened 2022-09-05
fix: Add browse tests

LautaroPetaccio:
This PR adds tests for the browse sagas when setting the `onlyOnRent` or `onlyOnSell` filters.


juanmahidalgo (approved):
🚀

PR #812, opened 2022-09-14
feat: Add emote categories to originals sidebar

### LINKED ISSUE #808: Add Emote section and emote categories to Originals Sidebar
## 🎉 Description

When seeing Originals created by a user, let's add the Emote section and its categories in the marketplace originals Sidebar so the users can browse between them.

* [ ] 🐛 This is a bug report.

cyaiox:
This PR adds the emote section and categories to the `Originals` sidebar.

image

Closes #808

### DIFF
--- a/webapp/src/modules/routing/sagas.ts
+++ b/webapp/src/modules/routing/sagas.ts
@@ -120,10 +120,11 @@ export function* routingSaga() {
 }
 
 function* handleFetchAssetsFromRoute(action: FetchAssetsFromRouteAction) {
-  const newOptions: BrowseOptions = yield getNewBrowseOptions(
+  const newOptions: BrowseOptions = yield call(
+    getNewBrowseOptions,
     action.payload.options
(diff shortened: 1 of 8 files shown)


webapp/src/modules/routing/sagas.ts
```diff
@@ -236,6 +237,11 @@ export function* fetchAssetsFromRoute(options: BrowseOptions) {
           ? getSearchWearableCategory(section)
           : undefined
 
+        const emoteCategory =
```
fzavalia, line 240: Can this be tested if it not too complicated?

fzavalia (changes requested):
🙏

webapp/src/modules/routing/sagas.ts
cyaiox: sure!

fzavalia (approved):

PR #819, opened 2022-09-16
feat: Add emote controls in the try-on screen

### LINKED ISSUE #736: [Try on] Add emote controls
## 🎉 Description

Nowadays, the preview is already separated. 
Add the following:
- Play Emote button. If it's not looped, go back to initial state when it finishes.
- Zoom controls

Implementation notes:
To be sure if it's a loop/1 time play, we need to fetch the data first from the catalyst to know it (the graph doesn't know)
An alternative, could be to ask the controller for the 'isPlaying' flag (maybe add an interval and poll that)

* [ ] 🐛 This is a bug report.

cyaiox:
This PR adds zoom in/out and play/stop controls to emotes on the try-on screen.

image

image

https://user-images.githubusercontent.com/3170051/190712374-f24b58f3-7bf4-44b4-8103-b01ea004d6b9.mov

Closes #736

### DIFF
--- a/webapp/src/modules/ui/preview/reducer.spec.ts
+++ b/webapp/src/modules/ui/preview/reducer.spec.ts
@@ -1,0 +1,45 @@
+import { IPreviewController } from '@dcl/schemas'
+import { setWearablePreviewController, setEmotePlaying } from './actions'
+import { previewReducer, INITIAL_STATE, PreviewState } from './reducer'
+
+let state: PreviewState
+
+beforeEach(() => {
+  state = { ...INITIAL_STATE }
+})
+
+describe('when reducing the successful action of setting the wearable preview', () => {
+  const controller = {} as IPreviewController
+
+  it('should return a state with the created wearable preview', () => {
+    expect(
+      previewReducer(state, setWearablePreviewController(controller))
+    ).toEqual({
+      ...INITIAL_STATE,
+      wearablePreviewController: controller
+    })
+  })
+
+  it('should return a state with the unloaded wearable preview', () => {
+    expect(previewReducer(state, setWearablePreviewController(null))).toEqual({
+      ...INITIAL_STATE,
+      wearablePreviewController: null
+    })
+  })
+})
+
+describe('when reducing the successful action of setting the playing emote state', () => {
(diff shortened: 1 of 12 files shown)


webapp/src/modules/ui/preview/reducer.ts
```diff
@@ -21,6 +34,18 @@ export function previewReducer(
...
+        ...state,
+        wearablePreviewController: action.payload.controller
+      }
+    }
+    case SET_EMOTE_PLAYING: {
+      return {
+        ...state,
+        isPlayingEmote: action.payload.isPlayingEmote
+      }
+    }
```
LautaroPetaccio, line 48: Would you mind adding tests for the reducer?

cyaiox: doing!

webapp/src/modules/ui/preview/reducer.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    })
+  })
+
+  it('should return a state with the unloaded wearable preview', () => {
+    expect(previewReducer(state, setWearablePreviewController(null))).toEqual({
+      ...INITIAL_STATE,
+      wearablePreviewController: null
+    })
+  })
+})
```
LautaroPetaccio, line 29: The `success` keyword is used when there's an action called success, here you can leave it behind.
As the method is executed with a different context, that is, the controller changes, a different describe is needed:
```suggestion
describe('when reducing action of setting the wearable preview', () => {
  const controller = {} as IPreviewController
  it('should return a state with the created wearable preview', () => {
    expect(
      previewReducer(state, setWearablePreviewController(controller))
    ).toEqual({
      ...INITIAL_STATE,
      wearablePreviewController: controller
    })
  })
})
describe('when reducing action of setting the wearable preview to null', () => {
  it('should return a state with the unloaded wearable preview', () => {
    expect(previewReducer(state, setWearablePreviewController(null))).toEqual({
      ...INITIAL_STATE,
      wearablePreviewController: null
    })
  })
})
```
The same applies for the other test.

LautaroPetaccio (approved):

PR #837, opened 2022-09-22
feat: Derive parcels in estate for performance

### LINKED ISSUE #838: Optimize parcels inside the estate entity
Should use the @derivedFrom tool instead of adding them manually

fzavalia:
Closes #838

### DIFF
--- a/indexer/package.json
+++ b/indexer/package.json
@@ -11,7 +11,8 @@
     "deploy-studio:mainnet": "graph deploy --node https://api.studio.thegraph.com/deploy/ --ipfs https://api.thegraph.com/ipfs/ decentraland-marketplace-ethereum-mainnet",
     "deploy:ropsten": "ts-node ./scripts/buildData.ts --network ropsten && ts-node ./scripts/deploy.ts --network ropsten",
     "deploy:mainnet": "ts-node ./scripts/buildData.ts --network mainnet && ts-node ./scripts/deploy.ts --network mainnet",
-    "deploy:goerli": "ts-node ./scripts/buildData.ts --network goerli && ts-node ./scripts/deploy.ts --network goerli"
+    "deploy:goerli": "ts-node ./scripts/buildData.ts --network goerli && ts-node ./scripts/deploy.ts --network goerli",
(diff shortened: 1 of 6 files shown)


indexer/schema.graphql
```diff
@@ -91,7 +91,7 @@ type Estate @entity {
   id: ID!
   tokenId: BigInt!
   owner: Account!
-  parcels: [Parcel!]
+  parcels: [Parcel!]! @derivedFrom(field: "estate")
```
LautaroPetaccio, line 94: I think that it's also possible to derive the `estate` in the `Parcel` type, right?

cazala: mm don't you need to set at least one of the two manually, so you can derive the other? It makes more sense to set the estate prop manually and derive the parcels on the other end.

indexer/package.json
```diff
@@ -11,7 +11,8 @@
     "deploy-studio:mainnet": "graph deploy --node https://api.studio.thegraph.com/deploy/ --ipfs https://api.thegraph.com/ipfs/ decentraland-marketplace-ethereum-mainnet",
     "deploy:ropsten": "ts-node ./scripts/buildData.ts --network ropsten && ts-node ./scripts/deploy.ts --network ropsten",
     "deploy:mainnet": "ts-node ./scripts/buildData.ts --network mainnet && ts-node ./scripts/deploy.ts --network mainnet",
-    "deploy:goerli": "ts-node ./scripts/buildData.ts --network goerli && ts-node ./scripts/deploy.ts --network goerli"
+    "deploy:goerli": "ts-node ./scripts/buildData.ts --network goerli && ts-node ./scripts/deploy.ts --network goerli",
+    "deploy:temp": "ts-node ./scripts/buildData.ts --network mainnet && ts-node ./scripts/deploy.ts --network temp"
```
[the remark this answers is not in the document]
fzavalia: We have a new {graphname}-temp graph as a backup on various subgraphs. This is just a helper

indexer/schema.graphql
fzavalia: As @cazala says, besides, the estate inside the parcel is a 1 to 1 relation so its easy a light to set manually.

Estates on the pther hand might hundreds or thousands of parcels. Making it a really heavy computation if set manually.

cazala (approved):

PR #856, opened 2022-10-05
feat: Introduce MVMF badge & MVMF Page

### LINKED ISSUE #853: Create the MVMF page
## 🎉 Description

Tasks: 
- [ ] Add Metaverse Festival tab to the Navigation component
- [ ] Create new route that renders the AssetBrowse component. On mount fetches the addresses from the new builder-server endpoint by using the tag "MVMF" tag and uses them to filter by those contract addresses.
- [ ] Add support to the AssetBrowse for multiple addresses

* [ ] 🐛 This is a bug report.

### LINKED ISSUE #855: Add MVMF tag in the detail pages
## 🎉 Description

Tasks
- [ ] Create a MVMF Badge component that fetches the new builder-server endpoint and renders the badge in case the contract address passed is part of the response.
- [ ] The MVMF Badge has to redirect to the new route on the onClick.
- [ ] In the detail pages (NFT detail & Item detail) components add the MVMF badge.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
image

image

image

Closes #853 
Closes #855

### DIFF
--- a/webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
+++ b/webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
@@ -19,10 +19,14 @@ const MVMFBadge = ({ contract, isMVMFEnabled }: Props) => {
 
   useEffect(() => {
     ;(async () => {
-      const addresses = await builderAPI.fetchAddressesByTag([MVMFTag])
-      setContracts(addresses)
-      if (isMVMFEnabled && addresses.includes(contract)) {
-        setShowBadge(true)
+      try {
+        const addresses = await builderAPI.fetchAddressesByTag([MVMFTag])
+        setContracts(addresses)
+        if (isMVMFEnabled && addresses.includes(contract)) {
+          setShowBadge(true)
+        }
+      } catch (error) {
+        console.error(error)
       }
     })()
   }, [contract, isMVMFEnabled])
--- a/webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
+++ b/webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
@@ -1,7 +1,11 @@
 import React, { useEffect, useMemo, useState } from 'react'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import * as decentraland from '../../../modules/vendor/decentraland'
 import { locations } from '../../../modules/routing/locations'
 import { builderAPI } from '../../../modules/vendor/decentraland/builder/api'
+import { VendorName } from '../../../modules/vendor'
+import { SortBy } from '../../../modules/routing/types'
+import { AssetType } from '../../../modules/asset/types'
 import IconBadge from '../IconBadge'
 import { Props } from './MVMFBadge.types'
 
@@ -12,7 +16,13 @@ const MVMFBadge = ({ contract, isMVMFEnabled }: Props) => {
   const [contracts, setContracts] = useState<string[]>()
 
   const href = useMemo(() => {
-    return locations.browse({
+    return locations.MVMF22({
+      section: decentraland.Section.WEARABLES,
+      vendor: VendorName.DECENTRALAND,
+      page: 1,
+      sortBy: SortBy.RECENTLY_LISTED,
+      onlyOnSale: true,
+      assetType: AssetType.ITEM,
       contracts
     })
   }, [contracts])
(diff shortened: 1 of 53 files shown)


webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useEffect, useMemo, useState } from 'react'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { locations } from '../../../modules/routing/locations'
+import { builderAPI } from '../../../modules/vendor/decentraland/builder/api'
+import IconBadge from '../IconBadge'
+import { Props } from './MVMFBadge.types'
+
+const MVMFTag = 'MVMF 2022'
```
cazala, line 8: According to notion the tag is `MVMF22`

webapp/src/modules/routing/types.ts
```diff
@@ -42,6 +42,7 @@ export type BrowseOptions = {
   search?: string
   contracts?: string[]
   address?: string
+  contractAddresses?: string[]
```
cazala, line 45: I believe we already have a `contracts` parameter that serves the same purpose as this. This is because the NFTs already supported filtering by multiple contract addresses. We should use that one instead of duplicating it.

webapp/src/modules/vendor/decentraland/item/types.ts
```diff
@@ -29,6 +29,7 @@ export type ItemFilters = {
   rarities?: Rarity[]
   wearableGenders?: WearableGender[]
   contractAddress?: string
+  contractAddresses?: string[]
```
cazala, line 32: There's a single `contractAddress` filter for items, we should refactor it to be an array instead of supporting both single and multiple params. Also it is called `contracts` in the nft filters, I think we should name it the same way for consistency: https://github.com/decentraland/marketplace/blob/master/webapp/src/modules/vendor/decentraland/nft/types.ts#L26

webapp/src/modules/routing/sagas.ts
```diff
@@ -181,7 +181,8 @@ export function* fetchAssetsFromRoute(options: BrowseOptions) {
     onlyOnRent,
     onlySmart,
     isMap,
-    contracts
+    contracts,
+    contractAddresses
```
cazala, line 185: here we have both `contracts` and `contractAddresses` that serve the same purpose

cazala (changes requested):

webapp/src/modules/routing/types.ts
juanmahidalgo: Absolutely right, I completely missed the `contracts` param. Updated to reuse it!

webapp/src/modules/vendor/decentraland/item/types.ts
juanmahidalgo: Agreed! Updated :)

webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const href = useMemo(() => {
+    return locations.browse({
+      contracts
+    })
+  }, [contracts])
+
+  useEffect(() => {
+    ;(async () => {
+      const addresses = await builderAPI.fetchAddressesByTag([MVMFTag])
```
cazala, line 22: should we wrap this in a try/catch?

webapp/src/components/MVMFPage/MVMFPage.types.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '../../modules/event/actions'
+import { Section } from '../../modules/vendor/routing/types'
+import { VendorName } from '../../modules/vendor/types'
+
+export type Props = {
+  vendor: VendorName
+  assetType: AssetType
+  section: Section
+  isFullscreen?: boolean
+  fetchEventContracts: typeof fetchEventRequest
```
cazala, line 15: Could we rename this to `onFetchEventContracts` to keep it consistent with standards?

cazala (changes requested):
When I click on the `Metaverse Festival` tab, it takes me to the `Listings` section first, instead of the `Store`:

Screen Shot 2022-10-07 at 13 26 50

webapp/src/components/MVMFPage/MVMFPage.types.ts
juanmahidalgo: done!

webapp/src/components/AssetPage/MVMFBadge/MVMFBadge.tsx
juanmahidalgo: sure! updated

cazala (approved):

PR #858, opened 2022-10-07
feat: Introduce MVMFBanner and adds it to homepage and mvmf page

juanmahidalgo:
Creates the component to render the banner. 
Note: The images are still missing, waiting to the product team to create them

### DIFF
--- a/webapp/src/components/MVMFPage/MVMFBanner/MVMFBanner.tsx
+++ b/webapp/src/components/MVMFPage/MVMFBanner/MVMFBanner.tsx
@@ -1,0 +1,14 @@
+import React from 'react'
+import classNames from 'classnames'
+import { Props } from './MVMFBanner.types'
+import './MVMFBanner.css'
+
+const MVMFBanner = ({ type }: Props) => {
+  return (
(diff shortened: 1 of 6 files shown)


webapp/src/components/MVMFPage/MVMFBanner/MVMFBanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Button, Container } from 'decentraland-ui'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import * as decentraland from '../../../modules/vendor/decentraland'
+import { locations } from '../../../modules/routing/locations'
+import { VendorName } from '../../../modules/vendor'
+import { SortBy } from '../../../modules/routing/types'
+import { Props } from './MVMFBanner.types'
+import './MVMFBanner.css'
+
+const MVMFBanner = ({ type, onNavigate }: Props) => {
```
cazala, line 12: Could we add the feature flag to this component and make it invisible/null if when the feature flag is not on?

cazala (changes requested):
Every

cazala (approved):

PR #860, opened 2022-10-07
feat: use modals module for Rental Modals

juanmahidalgo:


webapp/src/components/Modals/index.ts
```diff
@@ -1,3 +1,4 @@
 export { default as TestModal } from './TestModal'
 export { ClaimLandModal } from './ClaimLandModal'
 export { RemoveRentalModal } from './RemoveRentalModal'
+export { CreateRentalModal } from '../CreateRentalModal'
```
cazala, line 4: Could we move the `CreateRentalModal` within this directory to keep it more tidy? Also I noticed that the directory is called `CreateRentalModal` and the component inside is called `RentalModal`, following the standards the name of the component and the directory should match, although I don't think neither `CreateRentalModal` nor `RentalModal` are goods names to be honest. That's my bad, I gave them those names but `CreateRentalModal` is wrong now that it can also be used to edit a listing, and just `RentalModal` is too generic. I think I would take the chance to rename both the directory and the component to something like `RentalListingModal` (since it can be used both for creating and editing a listing). What do you think?

cazala (changes requested):

webapp/src/components/Modals/index.ts
juanmahidalgo: Absolutely agree :) Updated!

cazala (approved):

Issue #873, opened 2022-10-12
Users can exploit mana prices below 0.01 to make the UI show that the price is 0


cazala:
There are a few users that have updated the price of their items to be 0.001 MANA (via smart contract, because the UI doesn't let you)

Like this: https://market.decentraland.org/contracts/0x05c67bea23cc0879b07e2d74534395869f314013/items/0

Or that: https://market.decentraland.org/contracts/0x396345293d73a4acd1c77eabe4bfda11881fd5a1/items/0

So the UI shows that the price is 0 and users buy it, but they pay a little bit of mana. Also the indexer ignores the sales where the price is 0, but since this is not 0, it is computing them as sales, making these items show up all the time in the trending items, and counting as sales for the creators, making them go up the rankings.

Tasks:

- [ ] We need to change the `parseMANANumber` and `formatWeiMANA` so if the number is so small that it becomes 0 after using `toFixed`, then we show the minimum number that can be using `MAXIMUM_FRACTION_DIGITS`. So for example the price `0.001` should be shown as `0.01` instead of `0`.

PR #874, opened 2022-10-12
fix: show minimum price instead of zero

cazala:
Fixes: #873

With this, the prices and sales that are smaller than 0 now show the minimum price:

Screen Shot 2022-10-12 at 14 38 28

But the items that are truly free still show 0:

Screen Shot 2022-10-12 at 14 38 59


juanmahidalgo (approved):
LGTM 👏

PR #879, opened 2022-10-14
fix: zoom button style

cazala:
This PR centers the icons of the zoom buttons in the emote preview, and also makes the corners round as the rest of the buttons.

Before:
Screen Shot 2022-10-14 at 13 04 55

After:
Screen Shot 2022-10-14 at 13 08 58

### DIFF
--- a/webapp/src/components/AssetImage/AssetImage.css
+++ b/webapp/src/components/AssetImage/AssetImage.css
@@ -167,13 +167,16 @@
 
 .AssetImage .zoom-controls > .ui.button.zoom-control i.icon {
   margin: 0 !important;
+  width: 16px;
+  height: 16px;
 }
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/AssetImage/AssetImage.css
```diff
@@ -167,13 +167,16 @@
...
   margin: 0 !important;
+  width: 16px;
+  height: 16px;
 }
 
 .AssetImage .zoom-controls > .ui.button.zoom-in-control {
-  width: 32px;
-  height: 32px;
-  padding: 0;
-  min-width: unset;
```
cazala, line 176: I deleted these styles because they were repeated in the `.AssetImage .zoom-controls > .ui.button.zoom-control` rule a few lines above

cyaiox (approved):

PR #903, opened 2022-10-25
feat: disable most of the Rentals accions in mobile version

juanmahidalgo:
image

image

image


LautaroPetaccio (commented):
Looks great!
According to the figma, we shouldn't see the Owner component in mobile, would you mind removing it?
Also, does the highlight component behave correctly when having multiple highlights in the `ManagementAssetPage` and the Item detail page?
Screen Shot 2022-10-25 at 20 06 39

juanmahidalgo:
> Looks great! According to the figma, we shouldn't see the Owner component in mobile, would you mind removing it? Also, does the highlight component behave correctly when having multiple highlights in the `ManagementAssetPage` and the Item detail page? Screen Shot 2022-10-25 at 20 06 39

Sure! I've added some CSS for the highlights section so it looks like the design:

Item detail page:
image

Manage page:

image


Also, owner section removed!

LautaroPetaccio (approved):

PR #907, opened 2022-10-25
feat: Retrieve different statuses

LautaroPetaccio:
This PR changes the way we're requesting NFTs to be able to retrieve NFTs according to their rental status as well.

### DIFF
--- a/webapp/src/modules/nft/actions.ts
+++ b/webapp/src/modules/nft/actions.ts
@@ -1,5 +1,5 @@
 import { action } from 'typesafe-actions'
-import { Order, RentalListing } from '@dcl/schemas'
+import { Order, RentalListing, RentalStatus } from '@dcl/schemas'
 import { buildTransactionPayload } from 'decentraland-dapps/dist/modules/transaction/utils'
 import { ErrorCode } from 'decentraland-transactions'
 
@@ -7,6 +7,7 @@ import { SortDirection } from '../routing/types'
 import { Account } from '../account/types'
 import { getAssetName } from '../asset/utils'
 import { NFT, NFTSortBy, NFTsFetchOptions, NFTsFetchParams } from './types'
+import { FetchOneOptions } from '../vendor'
 
 // Fetch NFTs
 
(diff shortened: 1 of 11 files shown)


webapp/src/modules/nft/actions.ts
```diff
@@ -7,6 +7,7 @@ import { SortDirection } from '../routing/types'
 import { Account } from '../account/types'
 import { getAssetName } from '../asset/utils'
 import { NFT, NFTSortBy, NFTsFetchOptions, NFTsFetchParams } from './types'
+import { FetchOneOptions } from '../vendor'
```
juanmahidalgo, line 10: can we move this line one up? :)

webapp/src/modules/vendor/decentraland/nft/api.ts
```diff
@@ -19,10 +20,15 @@ class NFTAPI extends BaseAPI {
...
+  async fetchOne(
+    contractAddress: string,
+    tokenId: string,
+    options?: FetchOneOptions
+  ): Promise<NFTResult> {
     const response: NFTResponse = await this.request('get', '/nfts', {
       contractAddress,
-      tokenId
+      tokenId,
+      status: options?.status
```
juanmahidalgo, line 31: do we want to spread the options here? we're receiving an `options` object but just using the `status` prop, it may make sense to spread the entire object so we can pass other params later on and we don't have to change it here. WDYT?

webapp/src/modules/nft/sagas.ts
```diff
@@ -94,7 +94,8 @@ function* handleFetchNFTRequest(action: FetchNFTRequestAction) {
     ]: AwaitFn<typeof vendor.nftService.fetchOne> = yield call(
       [vendor.nftService, 'fetchOne'],
       contractAddress,
-      tokenId
+      tokenId,
+      { status: options?.status }
```
juanmahidalgo, line 98:
```suggestion
    options
```
can we use the whole options as parameter here? I believe it looks a bit cleaner and we avoid the `{ status: undefined }` case. WDYT?

juanmahidalgo (approved):
a few small suggestions that are not mandatory to apply. Looks great 👏

webapp/src/modules/vendor/decentraland/nft/api.ts
LautaroPetaccio: Changed it as well!

webapp/src/modules/nft/sagas.ts
LautaroPetaccio: It's better, you're right, changing it.

webapp/src/modules/nft/actions.ts
LautaroPetaccio: Done!

juanmahidalgo (approved):
🚀

PR #960, opened 2022-11-10
feat: add clamining back state on rent page

### LINKED ISSUE #909: Add the claiming back land state to the on rent page
## 🎉 Description

Use the current mechanisms that we have in the marketplace to detect a claiming back transaction in process and show the "Claiming back status" in the on rent page.

* [ ] 🐛 This is a bug report.

flobarreto:
closes #909


webapp/src/modules/ui/browse/selectors.ts
```diff
@@ -90,3 +92,21 @@ export const getOnSaleElements = createSelector<
   ReturnType<typeof getOnSaleNFTs>,
   OnSaleElement[]
 >(getOnSaleItems, getOnSaleNFTs, (items, nfts) => [...items, ...nfts])
+
+
+export const getClaimingBackState = (state: RootState, nft: NFT): boolean => {
+  const userAddress = getAddress(state);
+
+  const transactionsClainLand = getTransactionsByType(state, userAddress!, CLAIM_LAND_TRANSACTION_SUBMITTED)
```
LautaroPetaccio, line 100: Following the name of the transaction:
```suggestion
  const claimLandTransactions = getTransactionsByType(state, userAddress!, CLAIM_LAND_TRANSACTION_SUBMITTED)
```

LautaroPetaccio: We must consider the case where the `userAddress` is undefined or null, as it could happen that the user is not logged in.
We can return false for this case.

webapp/src/modules/ui/browse/selectors.ts
```diff
@@ -90,3 +92,21 @@ export const getOnSaleElements = createSelector<
...
+    element.payload.tokenId === nft.tokenId &&
+    element.payload.contractAddress === nft.contractAddress).sort((a, b) => {
+      if (a.timestamp > b.timestamp) return -1
+      else if (a.timestamp < b.timestamp) return 1
+      return 0
+    })
+
+  console.log('flo llega aca', userAddress, transactionsClainLand)
+  return transactions[0] ? !!transactions[0].status : false
+}
```
LautaroPetaccio, line 112: Would you mind adding a new line here?

webapp/src/components/OnSaleOrRentList/OnRentListElement/OnRentListElement.container.ts
```diff
@@ -0,0 +1,12 @@
+import { connect } from 'react-redux'
+import OnRentListElement from './OnRentListElement'
+import { RootState } from '../../../modules/reducer'
+import { MapStateProps, OwnProps } from './OnRentListElement.types'
+import { getClaimingBackState } from '../../../modules/ui/browse/selectors'
```
LautaroPetaccio, line 5: Would you mind ordering these imports?

webapp/src/components/OnSaleOrRentList/OnRentListElement/OnRentListElement.tsx
```diff
@@ -39,13 +44,18 @@ const OnRentListElement = ({ nft, rental }: Props) => {
...
-                  end_date: format(endDate, 'MMM dd')
-                })
-              : null}
+            {rental.status === RentalStatus.OPEN ? (
+              t('on_rent_list.listed_for_rent')
+            ) : endDate ? (
+              t('on_rent_list.rented_until', {
+                end_date: format(endDate, 'MMM dd')
+              })
+            ) : isClaimingBackLandTransactionPending ? (
```
LautaroPetaccio, line 53: The message won't ever be shown if the LAND was rented, as rented LAND always have an `endDate`. You should move the condition up so whenever the transaction to claim back the land is pending it will be shown.

LautaroPetaccio (approved):

PR #961, opened 2022-11-10
feat: Render empty state component in home page section slideshows

### LINKED ISSUE #742: Show error / empty state if the Trendings endpoint fails
## 🎉 Description

Right now, if the endpoint fails, the carousel is rendered but without any content. Define with product if we should show an empty state or not show the section at all

* [ ] 🐛 This is a bug report.

Design:

image

kevinszuchet:
Closes #742 

For Trending Wearables:
image

For the rest of the sections (the message is configurable):
image

### DIFF
--- a/webapp/src/components/RankingsTable/RankingsTable.container.ts
+++ b/webapp/src/components/RankingsTable/RankingsTable.container.ts
@@ -11,7 +11,7 @@ import {
   fetchRankingsRequest,
   FETCH_RANKINGS_REQUEST
 } from '../../modules/analytics/actions'
-import AnalyticsVolumeDayData from './RankingsTable'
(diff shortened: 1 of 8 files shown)


webapp/src/components/RankingsTable/RankingsTable.container.ts
```diff
@@ -11,7 +11,7 @@ import {
   fetchRankingsRequest,
   FETCH_RANKINGS_REQUEST
 } from '../../modules/analytics/actions'
-import AnalyticsVolumeDayData from './RankingsTable'
+import RankingsTable from './RankingsTable'
```
juanmahidalgo, line 14: Thanks for renaming this old variable ❤️

PR #975, opened 2022-11-15
feat: show rented lands on the Atlas component

### LINKED ISSUE #974: Show the Rented lands with a different color in the Atlas component
## 🎉 Description

Show the rented lands + add a different color so the user knows at a glance which ones are on sale.
Also, update the Modal to show the real tile owner instead of the rentals contract address.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #974

image

### DIFF
--- a/webapp/src/components/Atlas/Atlas.tsx
+++ b/webapp/src/components/Atlas/Atlas.tsx
@@ -92,7 +92,7 @@ const Atlas: React.FC<Props> = (props: Props) => {
       nftsOnRent
         .map(([nft]) => nft)
         .reduce(
-          (lands, nft) => setLand(lands, nft, Color.CANDY_PURPLE),
+          (lands, nft) => setLand(lands, nft, Color.SUNISH),
           new Map<string, ReturnType<Layer>>()
         ),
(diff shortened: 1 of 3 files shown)


webapp/src/components/Atlas/Atlas.tsx
```diff
@@ -46,36 +47,55 @@ const Atlas: React.FC<Props> = (props: Props) => {
...
+
+  const userRentedTiles = useMemo(
+    () =>
+      nftsOnRent
+        .map(([nft]) => nft)
+        .reduce(
+          (lands, nft) => setLand(lands, nft, Color.MILLENIAL_ORANGE),
+          new Map<string, ReturnType<Layer>>()
+        ),
+    [nftsOnRent, setLand]
```
LautaroPetaccio, line 98: As `nftsOnRent` has the list of NFTs which the owner is a lessor, the color should be different than orange as we're using orange for the NFTs that the user rented (the user is a tenant of that NFT).

LautaroPetaccio (approved):

PR #976, opened 2022-11-15
fix: Various rentals fixes

LautaroPetaccio:
This PR does the following:
- In `AssetCard.container.ts` we weren't considering when the rental is not null. This was changed to consider it.
- Changed `ownedBy` to receive a rental in order to check for the lessor to ensure that the user is the owner of the asset.
- Fixes the owner check in the `SaleRentActionBox.tsx` so it considers the owner as the lessor.
- Blocks the `Rent` button when the asset is on rent (when a rental exists).
- Changes the `AssetProviderPage` to have a memoized version of the status array to prevent multiple requests.
- Fixes the styling between the message and the rental data in the `Rent.tsx` component.

### DIFF
--- a/webapp/src/components/AssetCard/AssetCard.container.ts
+++ b/webapp/src/components/AssetCard/AssetCard.container.ts
@@ -1,12 +1,10 @@
 import { connect } from 'react-redux'
 import { getLocation } from 'connected-react-router'
+import { RentalStatus } from '@dcl/schemas'
 import { RootState } from '../../modules/reducer'
 import { getData } from '../../modules/order/selectors'
-import { isLand } from '../../modules/nft/utils'
 import { getActiveOrder } from '../../modules/order/utils'
 import { isClaimingBackLandTransactionPending } from '../../modules/ui/browse/selectors'
-import { MapStateProps, OwnProps, MapDispatchProps } from './AssetCard.types'
-import AssetCard from './AssetCard'
 import { getView } from '../../modules/ui/browse/selectors'
 import { getAssetPrice, isNFT } from '../../modules/asset/utils'
 import { locations } from '../../modules/routing/locations'
@@ -16,6 +14,8 @@ import {
   getOpenRentalId
 } from '../../modules/rental/utils'
 import { getRentalById } from '../../modules/rental/selectors'
+import { MapStateProps, OwnProps, MapDispatchProps } from './AssetCard.types'
+import AssetCard from './AssetCard'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
   const { order, asset, rental } = ownProps
@@ -27,14 +27,15 @@ const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
     price = getAssetPrice(asset, getActiveOrder(asset, orders) || undefined)
   }
 
-  let rentalPricePerDay: string | null = null
   const openRentalId = getOpenRentalId(asset)
-
-  if (!rental && openRentalId && isLand(asset)) {
-    // The price per day is the same in this version of rentals
-    rentalPricePerDay = getMaxPriceOfPeriods(
-      getRentalById(state, openRentalId)!
-    )
+  let rentalPricePerDay: string | null = null
+  if (rental && rental.status === RentalStatus.OPEN) {
+    rentalPricePerDay = getMaxPriceOfPeriods(rental)
+  } else if (!rental && openRentalId) {
+    const rentalOfNFT = getRentalById(state, openRentalId)
+    if (rentalOfNFT && rentalOfNFT.status === RentalStatus.OPEN) {
+      rentalPricePerDay = getMaxPriceOfPeriods(rentalOfNFT)
+    }
   }
 
   return {
--- a/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.container.ts
+++ b/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.container.ts
@@ -15,13 +15,18 @@ import SaleRentActionBox from './SaleRentActionBox'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
   const wallet = getWallet(state)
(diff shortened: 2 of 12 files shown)


webapp/src/components/AssetCard/AssetCard.container.ts
```diff
@@ -27,14 +27,15 @@ const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
...
   const openRentalId = getOpenRentalId(asset)
-
-  if (!rental && openRentalId && isLand(asset)) {
-    // The price per day is the same in this version of rentals
-    rentalPricePerDay = getMaxPriceOfPeriods(
-      getRentalById(state, openRentalId)!
-    )
+  let rentalPricePerDay: string | null = null
+  if (rental && rental.status === RentalStatus.OPEN) {
+    rentalPricePerDay = getMaxPriceOfPeriods(rental)
```
juanmahidalgo, line 33: Quick question here, why are we showing the `max` price of the periods in the card? Shouldn't we show the `min` one? Like showing the starting price

webapp/src/components/ManageAssetPage/ManageAssetPage.tsx
```diff
@@ -31,6 +31,7 @@ import { Details } from './Details'
 import { Sell } from './Sell'
 import { Rent } from './Rent'
 import { Map } from './Map'
+import { isOwnedBy } from '../../modules/asset/utils'
```
juanmahidalgo, line 34: Would you mind moving this import up a few lines?

webapp/src/components/ManageAssetPage/Rent/Rent.container.ts
```diff
@@ -3,16 +3,25 @@ import { Dispatch } from 'redux'
...
 import { MapStateProps, MapDispatchProps, OwnProps } from './Rent.types'
 import { RentalModalMetadata } from '../../Modals/RentalListingModal/RentalListingModal.types'
 import { VendorName } from '../../../modules/vendor'
 import { NFT } from '../../../modules/nft/types'
 import { Rent } from './Rent'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
-  isClaimingBackLandTransactionPending: isClaimingBackLandTransactionPending(state, ownProps.nft),
-  claimingBackLandTransaction: getLastTransactionForClaimingBackLand(state, ownProps.nft)
+  isClaimingBackLandTransactionPending: isClaimingBackLandTransactionPending(
```
juanmahidalgo, line 17: Offtopic, I think we should enforce the prettier step either in a pre-commit step or the CI as a check so we avoid the noise of introducing unrelated prettier changes in PRs (no changes needed, just a thought)

webapp/src/modules/asset/utils.ts
```diff
@@ -65,9 +65,16 @@ export function getAssetPrice(asset: Asset, order?: Order) {
     : null
 }
 
-export function isOwnedBy(asset: Asset, wallet: Wallet | null) {
+export function isOwnedBy(
+  asset: Asset,
+  wallet: Wallet | null,
```
juanmahidalgo, line 70: not introduced in the PR but I think receiving the whole `wallet` object it's a bit too much for this method since it just uses the `address` field. It also forced you to wire the `wallet` in the container in the `webapp/src/components/ManageAssetPage/ManageAssetPage.container.ts`. If this one receives the `userAddress` field instead, then the wallet is not needed in the `ManageAssetPage.container.ts` and you can use the userAddress as it was before.

webapp/src/components/ManageAssetPage/ManageAssetPage.types.ts
```diff
@@ -1,8 +1,11 @@
+import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
+
 export type Props = {
   userAddress?: string
```
juanmahidalgo, line 4: If we're keeping the `wallet` prop, I guess we should get rid of this one

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.container.ts
```diff
@@ -21,20 +21,18 @@ const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
     userHasAlreadyBidsOnNft: wallet
       ? getNFTBids(state).some(bid => bid.bidder === wallet.address)
       : false,
-    isOwner: isOwnedBy(ownProps.nft, wallet)
+    isOwner: isOwnedBy(
+      ownProps.nft,
+      wallet,
+      ownProps.rental ? ownProps.rental : undefined
```
juanmahidalgo, line 27: nit pick: 
```suggestion
      ownProps.rental ?? undefined
```
so it's used as the [one below](https://github.com/decentraland/marketplace/pull/976/files#diff-ee8c189daceb7fc88075336992dfe533703e2db7f881963d684458252eddd6c3R90)

webapp/src/components/AssetCard/AssetCard.container.ts
LautaroPetaccio: The prices of the different periods will be the same for the first milestone of the feature, but, to prevent any issues with people manually inserting rentals in our systems with different prices, we'll always show up the most expensive one as it's the lowest risk scenario we can have.

webapp/src/components/ManageAssetPage/ManageAssetPage.tsx
LautaroPetaccio: Done!

webapp/src/modules/asset/utils.ts
LautaroPetaccio: It's definitely not the best to receive the whole wallet, but I didn't want to change it through the whole app. As you said, the component should not know about the wallet, its only concern is the user address.

webapp/src/components/ManageAssetPage/ManageAssetPage.types.ts
LautaroPetaccio: Removed it!

PR #981, opened 2022-11-16
feat: add canBeClaimed util

### LINKED ISSUE #980: Create a util to compute if an asset can be claimed
An asset can have a rental in state OPEN or CANCELLED and it could still be hold by the rentals contract (by an old rental). We need to update the logic to check this as well.

juanmahidalgo:
Closes #980


webapp/src/components/ManageAssetPage/ManageAssetPage.tsx
```diff
@@ -157,7 +164,10 @@ export const ManageAssetPage = (props: Props) => {
                                 </section>
                                 <Sell
                                   nft={asset}
-                                  isBeingRented={isBeingRented(rental)}
+                                  hasNotBeenClaimed={
```
juanmahidalgo, line 167: Don't we need the `isBeingRented(rental)` too? I'm not sure why It's not here.
The sell should be disabled both for ongoings rentals and also for those that has not been claimed (either executed and finished and the open/cancelled but not claimed)

LautaroPetaccio: You're right, I missed that, adding it!

LautaroPetaccio (approved):

PR #983, opened 2022-11-16
fix: Rental history not showing days and price correctly

LautaroPetaccio:
This PR fixes an issue where placeholders where left for the RentalHistory component, making the real data to not be shown.


webapp/src/components/AssetPage/RentalHistory/RentalHistory.tsx
```diff
@@ -149,13 +139,11 @@ const RentalHistory = (props: Props) => {
                     <T
                       id="rental_history.mobile_price"
                       values={{
-                        days:
-                          rental.periods[rental.selected_period ?? 0].maxDays,
+                        days: getRentalChosenPeriod(rental).maxDays,
```
juanmahidalgo, line 142: Should we use `rental.rentedDays` here?

PR #985, opened 2022-11-17
fix: Avoid blocking the homepage render with the fetch contracts

### LINKED ISSUE #883: Review /contracts endpoint blocking the rendering
Evaluate if we can lazy load the /contracts fetch instead of blocking the UI.
I just tried it out and it took 4s,  blocking the rendering until it was resolved.

kevinszuchet:
Closes #883


juanmahidalgo (approved):
great job @kevinszuchet 🚀 
Approving but let's wait for @cazala I want to get his eyes on this in case we missed something.

cazala (approved):

PR #987, opened 2022-11-17
fix: Don't show cancelled rentals

LautaroPetaccio:


webapp/src/components/ManageAssetPage/Rent/Rent.tsx
```diff
@@ -114,7 +116,7 @@ export const Rent = (props: Props) => {
     wallet && rental && canBeClaimed(wallet.address, rental, nft)
 
   const rentButton = useMemo(() => {
-    if (!rental) {
+    if (!rental || canCreateANewRental(rental)) {
```
juanmahidalgo, line 119: I believe the !rental is not needed here since it's already included in canCreateANewRental, right?

webapp/src/modules/rental/utils.ts
```diff
@@ -192,6 +192,24 @@ export function isRentalListingOpen(rental: RentalListing | null) {
...
+ * Checks wether a new rental listing can be created on an asset.
+ * @param rental - A rental listing or null if it doesn't have one.
+ * @returns true if the rental listing exists and is open, false otherwise
+ */
+export function canCreateANewRental(rental: RentalListing | null) {
+  return (
+    rental === null ||
+    (rental !== null &&
+      (isRentalListingCancelled(rental) ||
+        (isBeingRented(rental) && hasRentalEnded(rental))))
```
juanmahidalgo, line 209: Do you think we can rename the isBeingRented method to something more related to the fact of being in status executed? The current one is a bit misleading since it can be executed but ended but by the name sounds that is still rented. Wdyt?

juanmahidalgo (approved):
Left a few comments but LGTM

PR #1001, opened 2022-11-18
fix: Change not listed for sale

### LINKED ISSUE #994: Not available for sell is confusing on the SaleRentActionBox
## 🎉 Description

When there's no open order on a LAND, the text says "Not available for sell" but something like "Currently not listed for sale" is way better.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
This PR changes the text `Not available for Sale` for `Currently not listed for sale` as its better than the old one.
Closes #994

### DIFF
--- a/webapp/src/modules/translation/locales/en.json
+++ b/webapp/src/modules/translation/locales/en.json
@@ -379,7 +379,7 @@
     },
     "sales_rent_action_box": {
       "until": "From {from_date} to {to_date}",
-      "not_for_sale": "Not available for Sale",
+      "not_for_sale": "Currently not listed for sale",
       "parcel_belongs_to_estate_bid": "It's not possible to bid on a parcel that belongs to an Estate, please consider going into the Estate page.",
       "parcel_belongs_to_estate_rent": "It's not possible to rent a parcel that belongs to an Estate, please consider going into the Estate page.",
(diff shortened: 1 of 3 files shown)


webapp/src/modules/translation/locales/en.json
```diff
@@ -379,7 +379,7 @@
     },
     "sales_rent_action_box": {
       "until": "From {from_date} to {to_date}",
-      "not_for_sale": "Not available for Sale",
+      "not_for_sale": "Currently not listed for sale",
```
juanmahidalgo, line 382: Should we update the `es.json` file too?

LautaroPetaccio: The spanish one seemed correct, as `No disponible para comprar` can be a pretty good translation for `Currently not listed for sale`. We can use something like `Actualmente no disponible para rentar` what do you think?

juanmahidalgo: I like the `Actualmente` or maybe appending `en este momento` could be a good option too. I'm fine with both :)

LautaroPetaccio: Done! Used `Actualmente no disponible para comprar`

cyaiox (approved):

juanmahidalgo (approved):

PR #1002, opened 2022-11-18
fix: Edit confirmation step buttons not disabled when loading

### LINKED ISSUE #992: Buttons are not disabled when editing a listing
## 🎉 Description

When editing a listing, both of the buttons, the one to bump the asset nonce and the one that creates the new listing are enabled during loading. This shouldn't happen as it might result in multiple requests.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
Closes #992


juanmahidalgo (approved):
LGTM

PR #1008, opened 2022-11-21
fix: View the LAND available to buy as default

### LINKED ISSUE #993: When accessing the LANDs view for the first time, a non selectable filter is selected
## 🎉 Description

When clicking in the LAND section of the Marketplace, the selected filter is the "recently_listed" one, but the "recently_sold" is the one displayed, making the filtering confusing.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
When getting for the first time into the LAND section of the Marketplace, the user has to unselect the map to go to the list view. Clicking the list button makes the user use a not available filter for `All Land`, so now whenever a user comes in for the first time, they'll load the `Available to buy` land first.
Closes #993

### DIFF
--- a/webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
+++ b/webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
@@ -69,7 +69,15 @@ const NFTFilters = (props: Props) => {
     ]
   } else {
     orderByDropdownOptions = [
+      {
+        value: SortBy.RECENTLY_LISTED,
+        text: t('filters.recently_listed')
+      },
       { value: SortBy.RECENTLY_SOLD, text: t('filters.recently_sold') },
+      {
+        value: SortBy.CHEAPEST,
+        text: t('filters.cheapest')
+      },
       { value: SortBy.NEWEST, text: t('filters.newest') },
       { value: SortBy.NAME, text: t('filters.name') }
     ]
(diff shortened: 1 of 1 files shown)


webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
```diff
@@ -147,9 +150,14 @@ const NFTFilters = (props: Props) => {
 
   const handleIsMapChange = useCallback(
     (isMap: boolean) => {
-      onBrowse({ isMap, isFullscreen: isMap, search: '' })
+      onBrowse({
+        isMap,
+        isFullscreen: isMap,
+        search: '',
+        onlyOnSale: (!onlyOnSale && !onlyOnRent) || onlyOnSale
```
juanmahidalgo, line 157: with this change, the `onlyOnSale` will always be present, either as `true` or `false`. In the past, it was send as `undefined`. Is there any problems with that?

juanmahidalgo (approved):
Left a question but LGTM!

webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
LautaroPetaccio: That's a great question. There's no functional change between returning undefined or false here as it's used as a boolean through all the NFT filter component.

PR #1010, opened 2022-11-22
feat: Cancel open listings on nft transfer

### LINKED ISSUE #1003: Call the refresh endpoint for a listing when transfering a LAND/Estate
As we did when accepting a bid, we need to call the refresh endpoint so the listing changes its status to `CANCELLED` when transfering a LAND/Estate

juanmahidalgo:
Closes #1003

### DIFF
--- a/webapp/src/modules/nft/sagas.spec.ts
+++ b/webapp/src/modules/nft/sagas.spec.ts
@@ -5,7 +5,7 @@ import * as matchers from 'redux-saga-test-plan/matchers'
 import { NFTCategory, Order, RentalListing, RentalStatus } from '@dcl/schemas'
 import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
 import { waitForTx } from 'decentraland-dapps/dist/modules/transaction/utils'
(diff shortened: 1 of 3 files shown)


webapp/src/modules/nft/sagas.spec.ts
```diff
@@ -336,10 +339,15 @@ describe('when handling the transfer NFT request action', () => {
   })
 
   describe('when the transfer is successful', () => {
-    it('should dispatch an action signaling the success of the action handling', () => {
+    it('should dispatch an action signaling the success of the action handling and cancel an existing rental listing', () => {
```
LautaroPetaccio, line 342: Would you mind adding a successful test where the NFT doesn't have a rental?

LautaroPetaccio (changes requested):

webapp/src/modules/nft/sagas.spec.ts
juanmahidalgo: for sure, added!

LautaroPetaccio (approved):

PR #1012, opened 2022-11-22
fix: Wait for rental to be cancelled when buying

### LINKED ISSUE #995: The rental is not refreshed nor waited to be cancelled when buying a LAND
When buying a LAND that has rental in an open state, after buying it, refresh it until the rental listing is cancelled.

LautaroPetaccio:
This PR makes the user wait for the rental to be cancelled when buying a LAND that has a rental.
Closes #995

### DIFF
--- a/webapp/src/modules/order/sagas.spec.ts
+++ b/webapp/src/modules/order/sagas.spec.ts
@@ -36,7 +36,7 @@ beforeEach(() => {
     name: 'aName',
     contractAddress: 'aContractAddress',
     tokenId: 'aTokenId',
-    activeOrderId: null,
+    openRentalId: null,
     vendor: VendorName.DECENTRALAND,
     chainId: ChainId.MATIC_MAINNET
   } as NFT
--- a/webapp/src/modules/order/sagas.spec.ts
+++ b/webapp/src/modules/order/sagas.spec.ts
@@ -1,1 +1,206 @@
-describe('when ', () => {})
+import {
+  ChainId,
+  Network,
+  Order,
+  RentalListing,
+  RentalStatus
+} from '@dcl/schemas'
+import {
+  ProviderType,
+  Wallet
+} from 'decentraland-dapps/dist/modules/wallet/types'
+import { ErrorCode } from 'decentraland-transactions'
+import { expectSaga } from 'redux-saga-test-plan'
+import { throwError } from 'redux-saga-test-plan/providers'
+import { call, select } from 'redux-saga/effects'
+import { NFT } from '../nft/types'
+import { getRentalById } from '../rental/selectors'
+import { waitUntilRentalChangesStatus } from '../rental/utils'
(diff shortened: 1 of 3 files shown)


webapp/src/modules/order/sagas.ts
```diff
@@ -57,17 +64,33 @@ function* handleExecuteOrderRequest(action: ExecuteOrderRequestAction) {
   const { order, nft, fingerprint } = action.payload
   try {
     if (
-      nft.contractAddress !== order.contractAddress &&
+      nft.contractAddress !== order.contractAddress ||
```
juanmahidalgo, line 67: good catch here!

webapp/src/modules/order/sagas.ts
```diff
@@ -18,6 +24,7 @@ import {
   cancelOrderSuccess,
   cancelOrderFailure
 } from './actions'
+import { VendorName } from '../vendor'
```
juanmahidalgo, line 27: can we move this import a little bit up?

webapp/src/modules/order/sagas.ts
```diff
@@ -57,17 +64,33 @@ function* handleExecuteOrderRequest(action: ExecuteOrderRequestAction) {
...
-    const txHash: string = yield call(() =>
-      orderService.execute(wallet, nft, order, fingerprint)
+    const txHash: string = yield call(
+      [orderService, 'execute'],
+      wallet,
+      nft,
+      order,
+      fingerprint
     )
+    if (nft.openRentalId) {
```
juanmahidalgo, line 85: Before start cancelling the rental listing, I think we should wait until the tx is confirmed. The execute only triggers the tx so this would inmediately start trying to cancel the listing even is the tx is not confirmed yet. Wdyt? 

```suggestion
    yield call(waitForTx, txHash)
    if (nft.openRentalId) {
```

webapp/src/modules/order/sagas.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+          status: RentalStatus.CANCELLED
+        } as RentalListing
+      })
+
+      it('should wait for the rental to change its status to cancelled and put the success action', () => {
+        return expectSaga(orderSaga)
+          .provide([
+            [select(getWallet), wallet],
+            [call([VendorFactory, 'build'], nft.vendor), vendor],
+            [select(getRentalById, nft.activeOrderId!), rentalListing],
```
juanmahidalgo, line 157: I think you intented to put these here:
```suggestion
            [select(getRentalById, nft.openRentalId!), rentalListing],
```

webapp/src/modules/order/sagas.ts
LautaroPetaccio: It would be great to use the `waitForTx` helper but the transaction is dispatched to the state in the `executeOrderSuccess` so we would have to make more changes in the code to make it work.

PR #1016, opened 2022-11-22
fix: Show rentals price up when there's no order

### LINKED ISSUE #1015: Rental price doesn't fill the order price space when there's no order
When there's no order, the rental price is on the line below, leaving a blank space.
To fix this, render the rental price in the order's price place when there's no order.

LautaroPetaccio:
Closes #1015


juanmahidalgo (approved):
LGTM

PR #1021, opened 2022-11-23
fix: Re-use the asset provider in the AssetPage

### LINKED ISSUE #1020: AssetPage is triggering 2 actions of type "Fetch NFT" instead of one
## 🎉 Description

When opening an NFT detail page like /contracts/0x25b6b4bac4adb582a0abd475439da6730777fbf7/tokens/115792089237316195423570985008687907852929702298719625575994209400481361428515, we can observe in the requests that 2 of the same type are triggered. This is due to the `AssetProvider` component being duplicated in the `AssetPage.tsx` page. Both of them will dispatch the fetch nft action on the mounting phase. 

* [X] 🐛 This is a bug report.

### LINKED ISSUE #1022: There's a race condition between the feature flags request and the Fetch NFT one
## 🎉 Description

The fetch nft request dispatched by the the AssetProvider, depends on the feature flags values (specifically of the rentals feature). So if the FF request is not resolved yet, the AssetProvider will fetch with the incorrect params.

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #1020
Closes #1022

### DIFF
--- a/webapp/src/components/AssetProvider/AssetProvider.tsx
+++ b/webapp/src/components/AssetProvider/AssetProvider.tsx
@@ -43,7 +43,16 @@ const AssetProvider = (props: Props) => {
     isLoadingFeatureFlags
   ])
 
-  return <>{children(asset, order, rental, isLoading)}</>
+  return (
+    <>
+      {children(
+        asset,
+        order,
+        rental,
+        isLoading || (isLoadingFeatureFlags && type === AssetType.NFT)
+      )}
+    </>
+  )
 }
 
 export default React.memo(AssetProvider)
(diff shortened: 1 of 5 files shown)


webapp/src/components/AssetProvider/AssetProvider.tsx
```diff
@@ -30,7 +33,15 @@ const AssetProvider = (props: Props) => {
...
+    contractAddress,
+    tokenId,
+    type,
+    onFetchNFT,
+    onFetchItem,
+    rentalStatus,
+    isLoadingFeatureFlags
+  ])
 
   return <>{children(asset, order, rental, isLoading)}</>
```
LautaroPetaccio, line 46: We might want to add to the `isLoading` flag that is passed down, the loading of the feature flags:
```suggestion
  return <>{children(asset, order, rental, isLoading || (isLoadingFeatureFlags && type === AssetType.NFT))}</>
```

juanmahidalgo: good idea! added

LautaroPetaccio (approved):

PR #1049, opened 2022-11-28
feat: add assetType param to the tracking events for rentals

juanmahidalgo:


webapp/src/modules/analytics/track.ts
```diff
@@ -246,8 +248,9 @@ track<ClaimAssetSuccessAction>(
 track<AcceptRentalListingSuccessAction>(
   ACCEPT_RENTAL_LISTING_SUCCESS,
   'Rent Land',
-  ({ payload: { periodIndexChosen, rental } }) => ({
+  ({ payload: { periodIndexChosen, rental, nft } }) => ({
     nftId: rental.nftId,
+    assetType: isParcel(nft) ? 'Parcel' : 'Estate',
```
LautaroPetaccio, line 253: What if we use the `NFTCategory` type to name the asset types?
```suggestion
    assetType: isParcel(nft) ? NFTCategory.PARCEL : NFTCategory.ESTATE,
```

LautaroPetaccio (approved):
Look great to me. I've left a minor comment to see what you thought about it.

PR #1051, opened 2022-11-28
fix: add a flag to save the first loading status of the FFs

### LINKED ISSUE #1050: The AssetProvider is fetching the Asset data in every feature flags request
With the latest change introduced, where the `AssetProvider` now depends on the loading status of the feature flags, we introduced an undesired behavior of fetching the asset on every FF request. It should only depend on the status of the very first feature flag request.

juanmahidalgo:
Closes #1050 

Right now, the `useEffect` depends on `isLoadingFeatureFlags`, which toggles between `false` and `true` on every single request (one per minute). We just need to know the first time if loads.

### DIFF
--- a/webapp/src/components/AssetProvider/AssetProvider.tsx
+++ b/webapp/src/components/AssetProvider/AssetProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { AssetType } from '../../modules/asset/types'
 import { Props } from './AssetProvider.types'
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/AssetProvider/AssetProvider.tsx
```diff
@@ -18,11 +18,18 @@ const AssetProvider = (props: Props) => {
     isLoadingFeatureFlags
   } = props
 
+  const [hasLoadedInitialFlags, setHasLoadedInitialFlags] = useState(false)
+  useEffect(() => {
+    if (!isLoadingFeatureFlags) {
+      setHasLoadedInitialFlags(true)
+    }
+  }, [isLoadingFeatureFlags])
```
LautaroPetaccio, line 26: I know that it's a bit more work, but, what if we move this logic to the features flag reducer so we can eventually use it somewhere else in the future? That is, when the FF are fetched successful for the first time, we can store a flag in the feature flags state and retrieve it later with a selector.

LautaroPetaccio: Will be implemented when [#290](https://github.com/decentraland/decentraland-dapps/issues/290) is implemented

PR #1053, opened 2022-11-28
fix: Prevent rental modals from closing

### LINKED ISSUE #1046: Some rental modals are closable when performing transactions or waiting
## 🎉 Description

Review all the modals and check for their closing capabilities. They might be closable even after clicking outside of them.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
Closes #1046


webapp/src/components/Modals/RentalListingModal/RentalListingModal.tsx
```diff
@@ -77,7 +77,7 @@ const RentalListingModal = (props: Props) => {
         styles.modal,
         isConfirmingEditingStep && styles.editingModal
       )}
-      onClose={onClose}
+      onClose={undefined}
```
juanmahidalgo, line 80:
```suggestion
      
```
let's remove the prop entirely here

juanmahidalgo (approved):
LGTM, just a small suggestion left

webapp/src/components/Modals/RentalListingModal/RentalListingModal.tsx
LautaroPetaccio: Although it seems unintuitive, onClose changes it's functionality if it's undefined as it seems to overwrite the default functionality. I'll change it to `() => undefined` though so there's no confusion when someone reads it.

PR #1060, opened 2022-11-28
feat: Replace locked banner with popup

### LINKED ISSUE #1059: Remove the "paused" message on the management page in favor of using popups
## 🎉 Description

According to a UX decision, remove the "paused" black messages and use tooltips in the buttons.

* [ ] 🐛 This is a bug report.

LautaroPetaccio:
Screen Shot 2022-11-28 at 20 48 16

Closes #1059


webapp/src/components/LandLockedPopup/LandLockedPopup.tsx
```diff
@@ -0,0 +1,21 @@
+import { Popup, useMobileMediaQuery } from 'decentraland-ui'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { isParcel } from '../../modules/nft/utils'
+import { isLandLocked } from '../../modules/rental/utils'
+
+export const LandLockedPopup = (props: any) => {
```
juanmahidalgo, line 6: Can we type this `any` from the `props` parameter?

webapp/src/components/ManageAssetPage/ManageAssetPage.tsx
```diff
@@ -38,6 +32,7 @@ import { Details } from './Details'
 import { Sell } from './Sell'
 import { Rent } from './Rent'
 import { Map } from './Map'
+import { LandLockedPopup } from '../LandLockedPopup'
```
juanmahidalgo, line 35: can we move this import a few lines up?

webapp/src/components/ManageAssetPage/ManageAssetPage.tsx
```diff
@@ -172,51 +144,43 @@ export const ManageAssetPage = (props: Props) => {
...
+                                        asset={asset}
+                                        rental={rental}
+                                        address={wallet.address}
                                       >
-                                        {t('manage_asset_page.transfer')}
-                                      </Button>
+                                        <Button
+                                          className={styles.transfer}
+                                          as={Link}
+                                          disabled={
```
juanmahidalgo, line 155: it would be nice if the component `LandLockedPopup` would also diasble the button if needed, so we don't have to replicate this check here. Wdyt? We could clone the child element and override the `disabled` prop inside the `LandLockedPopup` component

LautaroPetaccio: I have though of that, but the problem is that the `LandLockedPopup` is not limited to buttons, it could be anything, so it might be a bit confusing to force props on its children, what do you think about that?

juanmahidalgo (approved):
LGTM

PR #1067, opened 2022-11-29
fix: Block buy and rent button if there's no enough MANA

### LINKED ISSUE #1066: The buy and rent button are enabled when there's no enough MANA to rent our buy
## 🎉 Description

The `SaleRentActionBox.tsx` component allows users to buy and rent LAND even if they don't have enough MANA to do so. To prevent his, disable the corresponding button and add a message.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
Screen Shot 2022-11-29 at 17 41 13

Screen Shot 2022-11-29 at 17 56 41

Closes #1066

### DIFF
--- a/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
+++ b/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
@@ -112,19 +112,22 @@ const SaleRentActionBox = ({
     () =>
       !!rental &&
       !!currentMana &&
-      currentMana >=
-        +ethers.utils.formatEther(
+      ethers.BigNumber.from(currentMana).gte(
+        ethers.utils.formatEther(
           ethers.BigNumber.from(
             rental.periods[selectedRentalPeriodIndex].pricePerDay
           ).mul(rental.periods[selectedRentalPeriodIndex].maxDays)
-        ),
+        )
+      ),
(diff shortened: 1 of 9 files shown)


webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -106,6 +108,26 @@ const SaleRentActionBox = ({
     ? getRentalEndDate(rental!)!.getTime()
     : 0
   const rentalHasEnded = isCurrentlyRented && hasRentalEnded(rental!)
+  const hasEnoughManaToRent = useMemo(
+    () =>
+      !!rental &&
+      !!currentMana &&
+      currentMana >=
+        +ethers.utils.formatEther(
```
juanmahidalgo, line 116: I think that in the past we tried to advocate for the `Number` usage instead of the `+` just to be more verbose.

juanmahidalgo (approved):
a small suggestion but LGTM!

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
LautaroPetaccio: You're 100% right, changed it to a big numbers comparison so we can be sure that it will always work.

PR #1086, opened 2022-12-01
fix: fix logic of waiting for transfer transaction to finish

### LINKED ISSUE #1083: Transactions are not stored in some sagas when awaiting for a transaction to finish fails
There are actions that wait for transactions to be completed, like [transferring a LAND](https://github.com/decentraland/marketplace/blob/e5f7627bf174a349ac55b7008793a771a94f7f4f/webapp/src/modules/nft/sagas.ts#L146) when there's an open rental on them that, as the waiting is done before the transaction is submitted to the state, the waiting never finished as `waitForTx` uses sagas and the state to keep track of the transaction statuses.

To fix this, review all the operations that take into consideration the rental or that wait for a transaction to be completed to check if this bug is present.

juanmahidalgo:
Closes #1083

### DIFF
--- a/webapp/src/modules/nft/actions.ts
+++ b/webapp/src/modules/nft/actions.ts
@@ -88,6 +88,8 @@ export type FetchNFTFailureAction = ReturnType<typeof fetchNFTFailure>
 export const TRANSFER_NFT_REQUEST = '[Request] Transfer NFT'
 export const TRANSFER_NFT_SUCCESS = '[Success] Transfer NFT'
 export const TRANSFER_NFT_FAILURE = '[Failure] Transfer NFT'
+export const TRANSFER_NFT_TRANSACTION_SUBMITTED =
+  '[Submitted transaction] Transfer NFT'
 
 export const transferNFTRequest = (nft: NFT, address: string) =>
   action(TRANSFER_NFT_REQUEST, { nft, address })
@@ -108,6 +110,15 @@ export const transferNFTFailure = (
   error: string,
   errorCode?: ErrorCode
 ) => action(TRANSFER_NFT_FAILURE, { nft, address, error, errorCode })
+export const transferNFTransactionSubmitted = (nft: NFT, txHash: string) =>
+  action(
+    TRANSFER_NFT_TRANSACTION_SUBMITTED,
+    buildTransactionPayload(nft.chainId, txHash, {
+      tokenId: nft.tokenId,
+      contractAddress: nft.contractAddress,
+      chainId: nft.chainId
+    })
+  )
 
 export type TransferNFTRequestAction = ReturnType<typeof transferNFTRequest>
 export type TransferNFTSuccessAction = ReturnType<typeof transferNFTSuccess>
--- a/webapp/src/modules/nft/actions.ts
+++ b/webapp/src/modules/nft/actions.ts
@@ -5,7 +5,6 @@ import { ErrorCode } from 'decentraland-transactions'
 
 import { SortDirection } from '../routing/types'
 import { Account } from '../account/types'
-import { getAssetName } from '../asset/utils'
 import { FetchOneOptions } from '../vendor'
 import { NFT, NFTSortBy, NFTsFetchOptions, NFTsFetchParams } from './types'
 
@@ -93,16 +92,10 @@ export const TRANSFER_NFT_TRANSACTION_SUBMITTED =
 
 export const transferNFTRequest = (nft: NFT, address: string) =>
   action(TRANSFER_NFT_REQUEST, { nft, address })
-export const transferNFTSuccess = (nft: NFT, address: string, txHash: string) =>
+export const transferNFTSuccess = (nft: NFT, address: string) =>
   action(TRANSFER_NFT_SUCCESS, {
     nft,
-    address,
-    ...buildTransactionPayload(nft.chainId, txHash, {
-      tokenId: nft.tokenId,
-      contractAddress: nft.contractAddress,
-      name: getAssetName(nft),
-      address
-    })
+    address
   })
 export const transferNFTFailure = (
   nft: NFT,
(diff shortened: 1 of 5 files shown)


webapp/src/modules/nft/actions.ts
```diff
@@ -88,6 +88,8 @@ export type FetchNFTFailureAction = ReturnType<typeof fetchNFTFailure>
 export const TRANSFER_NFT_REQUEST = '[Request] Transfer NFT'
 export const TRANSFER_NFT_SUCCESS = '[Success] Transfer NFT'
 export const TRANSFER_NFT_FAILURE = '[Failure] Transfer NFT'
+export const TRANSFER_NFT_TRANSACTION_SUBMITTED =
+  '[Submitted transaction] Transfer NFT'
 
 export const transferNFTRequest = (nft: NFT, address: string) =>
   action(TRANSFER_NFT_REQUEST, { nft, address })
```
LautaroPetaccio, line 95: Would you mind making the `transferNFTSuccess` a simple action? That is, an action without the transaction payload, as the transaction is submitted to the state by the `transferNFTransactionSubmitted` action.

webapp/src/modules/nft/sagas.spec.ts
```diff
@@ -353,39 +353,58 @@ describe('when handling the transfer NFT request action', () => {
...
+                call([vendor.nftService, 'transfer'], wallet, address, nft),
+                Promise.resolve(txHash)
+              ],
+              [call(waitForTx, txHash), Promise.resolve()],
+              [
+                call(waitUntilRentalChangesStatus, nft, RentalStatus.CANCELLED),
+                Promise.resolve()
+              ]
+            ])
+            .put(transferNFTSuccess(nft, address, txHash))
```
LautaroPetaccio, line 382: If I'm not wrong, we're missing the put assertion here, right?
```suggestion
            .put(transferNFTransactionSubmitted(nft, txHash))
            .put(transferNFTSuccess(nft, address, txHash))
```

webapp/src/modules/nft/sagas.spec.ts
```diff
@@ -353,39 +353,58 @@ describe('when handling the transfer NFT request action', () => {
...
+              [select(getWallet), wallet],
+              [
+                call([vendor.nftService, 'transfer'], wallet, address, nft),
+                Promise.resolve(txHash)
+              ],
+              [call(waitForTx, txHash), Promise.resolve()]
+            ])
+            .put(transferNFTSuccess(nft, address, txHash))
+            .dispatch(transferNFTRequest(nft, address))
+            .run({ silenceTimeout: true })
```
LautaroPetaccio, line 402: Would you mind adding the `transferNFTransactionSubmitted` put assertion here?
```suggestion
            .put(transferNFTransactionSubmitted(nft, txHash))
            .put(transferNFTSuccess(nft, address, txHash))
            .dispatch(transferNFTRequest(nft, address))
            .run({ silenceTimeout: true })
```

webapp/src/modules/nft/sagas.spec.ts
```diff
@@ -394,9 +413,9 @@ describe('when handling the transfer NFT request action', () => {
               call([vendor.nftService, 'transfer'], wallet, address, nft),
               Promise.resolve(txHash)
             ],
-            [call(waitForTx, txHash), Promise.resolve()]
+            [call(waitForTx, txHash), Promise.reject(new Error('anError'))]
           ])
-          .put(transferNFTSuccess(nft, address, txHash))
+          .put(transferNFTFailure(nft, address, 'anError'))
```
LautaroPetaccio, line 418:
```suggestion
          .put(transferNFTransactionSubmitted(nft, txHash))
          .put(transferNFTFailure(nft, address, 'anError'))
```

webapp/src/modules/nft/sagas.ts
```diff
@@ -142,7 +143,7 @@ function* handleTransferNFTRequest(action: TransferNFTRequestAction) {
       address,
       nft
     )
-
+    yield put(transferNFTransactionSubmitted(nft, txHash))
```
LautaroPetaccio, line 146: This is for the line 147. Taking into consideration that after transferring an NFT, the user is redirected to the Activity page, can we move the waitForTx so it gets performed only when there's an open rental? I think we kind of broke the current experience with that change.

LautaroPetaccio (changes requested):
Left some comments, but it looks good!

webapp/src/modules/nft/actions.ts
```diff
@@ -88,26 +87,31 @@ export type FetchNFTFailureAction = ReturnType<typeof fetchNFTFailure>
...
   errorCode?: ErrorCode
 ) => action(TRANSFER_NFT_FAILURE, { nft, address, error, errorCode })
+export const transferNFTransactionSubmitted = (nft: NFT, txHash: string) =>
+  action(
+    TRANSFER_NFT_TRANSACTION_SUBMITTED,
+    buildTransactionPayload(nft.chainId, txHash, {
+      tokenId: nft.tokenId,
+      contractAddress: nft.contractAddress,
+      chainId: nft.chainId
+    })
```
LautaroPetaccio, line 113: This action is missing the `nft` and the `address` to later use it in [here](https://github.com/decentraland/marketplace/blob/c22d455a182d363d14aa9b1d42d7591cfd068e1d/webapp/src/components/ActivityPage/Transaction/Transaction.tsx#L210) by changing the action it expects to `TRANSFER_NFT_TRANSACTION_SUBMITTED` and in [here](https://github.com/decentraland/marketplace/blob/c22d455a182d363d14aa9b1d42d7591cfd068e1d/webapp/src/modules/analytics/track.ts#L115) by changing the action as well.

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):
It goes out with 🍟

PR #1100, opened 2022-12-02
feat: add waitForTx to the accept bid flow

### LINKED ISSUE #1093: Wait for the transaction to be completed when accepting a bid of a LAND with a rental listing
## 🎉 Description

Following this PR, when the user accepts a bid and there's an open rental, wait for the transaction to be completed to later wait for the rental listing to be cancelled.
    if (nft?.openRentalId) {
      yield call(waitForTx, txHash)
      const rental: RentalListing | null = yield select(
        getRentalById,
        nft.openRentalId
      )
      if (isRentalListingOpen(rental)) {
        yield call(waitUntilRentalChangesStatus, nft, RentalStatus.CANCELLED)
      }
    }
Remember to create a new action to send the transaction to the state before waiting for it and to change the track and Transaction files to use the new action.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #1093

### DIFF
--- a/webapp/src/modules/analytics/track.ts
+++ b/webapp/src/modules/analytics/track.ts
@@ -42,10 +42,10 @@ import {
   ARCHIVE_BID,
   UNARCHIVE_BID,
   PlaceBidSuccessAction,
-  AcceptBidSuccessAction,
   CancelBidSuccessAction,
   ArchiveBidAction,
-  UnarchiveBidAction
+  UnarchiveBidAction,
+  AcceptBidTransactionSubmittedAction
 } from '../bid/actions'
 import {
   BuyItemSuccessAction,
@@ -152,7 +152,7 @@ track<PlaceBidSuccessAction>(
   })
 )
 
-track<AcceptBidSuccessAction>(
+track<AcceptBidTransactionSubmittedAction>(
   ACCEPT_BID_TRANSACTION_SUBMITTED,
   'Accept bid',
   ({ payload }) => ({
--- a/webapp/src/modules/analytics/track.ts
+++ b/webapp/src/modules/analytics/track.ts
@@ -37,7 +37,7 @@ import {
 } from '../nft/actions'
 import {
   PLACE_BID_SUCCESS,
-  ACCEPT_BID_SUCCESS,
+  ACCEPT_BID_TRANSACTION_SUBMITTED,
   CANCEL_BID_SUCCESS,
   ARCHIVE_BID,
   UNARCHIVE_BID,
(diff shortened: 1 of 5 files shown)


webapp/src/modules/bid/actions.ts
```diff
@@ -84,9 +86,22 @@ export const acceptBidSuccess = (bid: Bid, txHash: string) =>
 export const acceptBidFailure = (bid: Bid, error: string) =>
   action(ACCEPT_BID_FAILURE, { bid, error })
 
+export const acceptBidtransactionSubmitted = (bid: Bid, txHash: string) =>
```
LautaroPetaccio, line 89: Would you mind removing the transaction data from `acceptBidSuccess`?

webapp/src/modules/analytics/track.ts
```diff
@@ -153,7 +153,7 @@ track<PlaceBidSuccessAction>(
 )
 
 track<AcceptBidSuccessAction>(
-  ACCEPT_BID_SUCCESS,
+  ACCEPT_BID_TRANSACTION_SUBMITTED,
```
LautaroPetaccio, line 156: When tracking this action, the payload is built using bid, but `ACCEPT_BID_TRANSACTION_SUBMITTED` is not receiving a bid when it's created.

webapp/src/modules/bid/actions.ts
```diff
@@ -84,9 +86,22 @@ export const acceptBidSuccess = (bid: Bid, txHash: string) =>
...
+  action(
+    ACCEPT_BID_TRANSACTION_SUBMITTED,
```
LautaroPetaccio, line 91: We should add bid here to use it later in the track file:
```suggestion
    ACCEPT_BID_TRANSACTION_SUBMITTED,
    bid,
```

webapp/src/modules/bid/sagas.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    })
+  })
+
+  describe('when accepting a bid request fails', () => {
+    it('should dispatch an action signaling the failure of the action handling', () => {
+      const bid = {
+        contractAddress: '0x123'
+      } as Bid
+      const address = 'anAddress'
+      const wallet = { address } as Wallet
```
LautaroPetaccio, line 80: This is repeated in some of the tests, if you use a beforeEach on an upper statement, you can re-use them if you want.

LautaroPetaccio (changes requested):
Beside these comments, the PR looks great! Thanks for adding the whole set of tests for the sagas, they look great!

webapp/src/modules/analytics/track.ts
juanmahidalgo: ooh true, good catch! added!

webapp/src/modules/bid/sagas.spec.ts
juanmahidalgo: for sure, updated!

webapp/src/modules/bid/actions.ts
juanmahidalgo: sure, updated!

LautaroPetaccio (approved):

PR #1116, opened 2022-12-06
fix: Add accept rental transaction to activity

### LINKED ISSUE #1115: The transaction to accept rentals is not shown in the activity page
## 🎉 Description

Whenever a user accepts a rental, the transaction to accept the rental should be shown in the activity page but it's not being shown. Add the mechanism to show this transaction.

* [x] 🐛 This is a bug report.

LautaroPetaccio:
This PR adds to the activity the accept rental transaction item that was missing.

Closes #1115

### DIFF
--- a/webapp/src/components/ActivityPage/Transaction/Transaction.tsx
+++ b/webapp/src/components/ActivityPage/Transaction/Transaction.tsx
@@ -24,6 +24,7 @@ import {
 } from '../../../modules/bid/actions'
 import { locations } from '../../../modules/routing/locations'
 import {
+  ACCEPT_RENTAL_LISTING_TRANSACTION_SUBMITTED,
   CLAIM_ASSET_TRANSACTION_SUBMITTED,
   REMOVE_RENTAL_TRANSACTION_SUBMITTED
 } from '../../../modules/rental/actions'
@@ -33,6 +34,7 @@ import { AssetProvider } from '../../AssetProvider'
 import { Mana } from '../../Mana'
 import { TransactionDetail } from './TransactionDetail'
 import { Props } from './Transaction.types'
+import { ethers } from 'ethers'
 
 const Transaction = (props: Props) => {
   const { tx, getContract } = props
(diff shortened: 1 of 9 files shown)


webapp/src/components/ActivityPage/Transaction/Transaction.tsx
```diff
@@ -421,6 +423,44 @@ const Transaction = (props: Props) => {
...
+                <T
+                  id="transaction.detail.accept_rental"
+                  values={{
+                    name: (
+                      <Link to={locations.manage(contractAddress, tokenId)}>
+                        {nft ? getAssetName(nft) : ''}
+                      </Link>
+                    ),
+                    pricePerDay: (
+                      <Mana network={nft?.network} inline>
```
fzavalia, line 447: What happens when network is undefined?

webapp/src/components/ActivityPage/Transaction/Transaction.tsx
```diff
@@ -421,6 +423,44 @@ const Transaction = (props: Props) => {
...
+                        {/* As this there might be already registered transactions and the price information is new, consider it optional */}
+                        {pricePerDay
+                          ? ethers.utils.formatEther(pricePerDay)
```
fzavalia, line 450: I assume pricePerDay is in Wei here

LautaroPetaccio: It is, yes.

webapp/src/components/ActivityPage/Transaction/Transaction.tsx
LautaroPetaccio: It is defaulted to `Ethereum` which is the same network that is used for the rental procedures.
[Here's the link to the code that does that](https://github.com/decentraland/ui/blob/60448893d0c4df40a3231aeba52ff8494a7a310b/src/components/Mana/Mana.tsx#L17).

PR #1138, opened 2022-12-12
feat: Add cheapest filter for rentals

LautaroPetaccio:
This PR adds the cheapest sort under the `MAX_RENTAL_PRICE`.
As of today, the maximum value of the rental price on all periods is the same as the minimum one, but we're using the maximum value everywhere due to the chance of someone registering the values manually.

### DIFF
--- a/webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
+++ b/webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
@@ -68,7 +68,8 @@ const NFTFilters = (props: Props) => {
         text: t('filters.recently_listed_for_rent')
       },
       { value: SortBy.NAME, text: t('filters.name') },
-      { value: SortBy.NEWEST, text: t('filters.newest') }
+      { value: SortBy.NEWEST, text: t('filters.newest') },
+      { value: SortBy.MAX_RENTAL_PRICE, text: t('filters.cheapest') }
     ]
   } else {
     orderByDropdownOptions = [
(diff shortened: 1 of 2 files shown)


webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
```diff
@@ -68,7 +68,8 @@ const NFTFilters = (props: Props) => {
         text: t('filters.recently_listed_for_rent')
       },
       { value: SortBy.NAME, text: t('filters.name') },
-      { value: SortBy.NEWEST, text: t('filters.newest') }
+      { value: SortBy.NEWEST, text: t('filters.newest') },
+      { value: SortBy.MAX_RENTAL_PRICE, text: t('filters.cheapest') }
```
fzavalia, line 72: The filters.cheapest key exists on localization files right?

webapp/src/modules/routing/search.ts
```diff
@@ -295,7 +295,7 @@ export function getAssetOrderBy(sortBy: SortBy) {
     }
     case SortBy.MIN_RENTAL_PRICE: {
       orderBy = NFTSortBy.MIN_RENTAL_PRICE
-      orderDirection = SortDirection.DESC
+      orderDirection = SortDirection.ASC
```
fzavalia, line 298: Good catch 🪨

fzavalia (approved):
🚀

webapp/src/components/Vendor/decentraland/NFTFilters/NFTFilters.tsx
LautaroPetaccio: It doest! it's used down below for the orders as well.

webapp/src/modules/routing/search.ts
LautaroPetaccio: I don't really know why we have this method in the UI because we don't send the order direction to the nft-server and the nft-server doesn't use it 🤷 .

PR #1139, opened 2022-12-13
feat: Add link to builder if tenant

### LINKED ISSUE #1121: Connect tenants viewing their rented land with the builder
As of today, tenants don't have the capability of going into the builder from the Marketplace where they rented a LAND. Use the provided mocks to create this link between the two sites.

LautaroPetaccio:
Screen Shot 2022-12-15 at 17 07 04
Screen Shot 2022-12-15 at 17 02 58
Screen Shot 2022-12-15 at 17 02 00
Screen Shot 2022-12-15 at 16 57 41

Closes #1121

### DIFF
--- a/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
+++ b/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
@@ -1,4 +1,4 @@
-import { memo, useCallback, useMemo, useState } from 'react'
+import React, { memo, useCallback, useMemo, useState } from 'react'
 import { NFTCategory } from '@dcl/schemas'
 import { ethers } from 'ethers'
 import intlFormat from 'date-fns/intlFormat'
@@ -11,6 +11,7 @@ import { hasAuthorization } from 'decentraland-dapps/dist/modules/authorization/
 import { isMobile } from 'decentraland-dapps/dist/lib/utils'
 import { formatWeiMANA } from '../../../lib/mana'
 import {
+  canBeClaimed,
   getMaxPriceOfPeriods,
   getRentalEndDate,
   hasRentalEnded,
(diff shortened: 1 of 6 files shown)


webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -49,6 +54,12 @@ const SaleRentActionBox = ({
   const isMobileView = isMobile()
   const isRentalOpen = isRentalListingOpen(rental)
   const isOwner = isOwnedBy(nft, wallet, rental ? rental : undefined)
+  const isTenant =
+    rental &&
+    wallet &&
+    addressEquals(rental.tenant ?? undefined, wallet.address)
+  console.log('Rental tenant', rental?.tenant)
+  console.log('Address', wallet?.address)
```
fzavalia, line 62:
```suggestion
```

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -247,13 +260,69 @@ const SaleRentActionBox = ({
                   </span>
                 </div>
               </div>
+            ) : isOwner && rental?.tenant && !rentalHasEnded ? (
```
fzavalia, line 263: https://github.com/decentraland/marketplace/pull/1139/files#diff-b1405057585dcb5e740824059979084501d068988037e64675615d8e4163d1c3R246 is it possible that the asset could have an order and the rest of the messages are ignored?

LautaroPetaccio: That's a great question but, if there's a rental with a tenant (meaning the rental was executed), the order wouldn't exist, as at the time it gets transferred, the order will be cancelled. This is also meant to be shown with executed rental listings, so if there's an open rental listing, this won't be shown.

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
LautaroPetaccio: Lol, missed them, great catch!

fzavalia (approved):

PR #1140, opened 2022-12-14
feat: Add rental chips

### LINKED ISSUE #1125: Add rental's chips to the lands in "my assets"
## 🎉 Description

Add the chips shown in the images below to the LANDs in the "My assets" menu.

* [ ] 🐛 This is a bug report.

LautaroPetaccio:
Screen Shot 2022-12-13 at 19 23 10

Closes #1125

### DIFF
--- a/webapp/src/components/AssetCard/AssetCard.container.ts
+++ b/webapp/src/components/AssetCard/AssetCard.container.ts
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux'
 import { getLocation } from 'connected-react-router'
-import { RentalStatus } from '@dcl/schemas'
+import { RentalListing } from '@dcl/schemas'
 import { RootState } from '../../modules/reducer'
 import { getData } from '../../modules/order/selectors'
 import { getActiveOrder } from '../../modules/order/utils'
@@ -9,16 +9,13 @@ import { getView } from '../../modules/ui/browse/selectors'
 import { getAssetPrice, isNFT } from '../../modules/asset/utils'
 import { locations } from '../../modules/routing/locations'
 import { View } from '../../modules/ui/types'
-import {
-  getMaxPriceOfPeriods,
-  getOpenRentalId
-} from '../../modules/rental/utils'
+import { getOpenRentalId } from '../../modules/rental/utils'
 import { getRentalById } from '../../modules/rental/selectors'
 import { MapStateProps, OwnProps, MapDispatchProps } from './AssetCard.types'
 import AssetCard from './AssetCard'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
-  const { order, asset, rental } = ownProps
+  const { order, asset } = ownProps
   const view = getView(state)
   let price: string | null = null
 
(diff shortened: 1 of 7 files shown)


webapp/src/components/AssetCard/AssetCard.container.ts
```diff
@@ -28,25 +25,20 @@ const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
...
     price,
     showListedTag:
       Boolean(view === View.CURRENT_ACCOUNT && price) &&
       getLocation(state).pathname !== locations.root(),
     isClaimingBackLandTransactionPending: isNFT(asset)
       ? isClaimingBackLandTransactionPending(state, asset)
-      : false
+      : false,
+    rental: rentalOfNFT,
+    showRentalBubble: view === View.CURRENT_ACCOUNT
```
juanmahidalgo, line 41:
```suggestion
    showRentalBubble: view === View.CURRENT_ACCOUNT &&
      getLocation(state).pathname !== locations.root(),
```
Should we add the `&&
      getLocation(state).pathname !== locations.root()` here as it does the `showListedTags`? I think it should have the same behavior. If I recall correctly, it was added to solve a bug where the tag was shown in the home page when coming back from the current account view.

webapp/src/components/AssetCard/AssetCard.tsx
```diff
@@ -58,16 +112,16 @@ const AssetCard = (props: Props) => {
...
-      {isClaimingBackLandTransactionPending ? (
-        <div className="LandBubble">
-          <Icon className="warning-icon" name="warning sign" />
-          {t('manage_asset_page.rent.claiming_back', {
-            asset: isParcel(asset) ? t('global.the_parcel') : t('global.the_estate')
-          })}
-        </div>
-      ) : showListedTag ? (
-        <ListedBadge className="listed-badge" />
+      {showRentalBubble ? (
```
juanmahidalgo, line 115: Should we add `!!rental && ` here? I'd rather not render the component at all if we don't have a rental object. Wdyt? Just avoiding the extra computation made by React by running the `RentalChip` side-effects

webapp/src/components/AssetCard/AssetCard.tsx
```diff
@@ -35,19 +43,65 @@ const RentalPrice = ({
...
-    isClaimingBackLandTransactionPending
+    isClaimingBackLandTransactionPending,
+    rental
   } = props
 
   const title = getAssetName(asset)
   const { parcel, estate, wearable, emote, ens } = asset.data
+  const rentalPricePerDay: string | null = useMemo(
+    () =>
+      rental && rental.status === RentalStatus.OPEN
```
juanmahidalgo, line 100:
```suggestion
      isRentalListingOpen(rental)
```
we could use the util here

juanmahidalgo (commented):
looks gorgeous, just a few comments/suggestions

webapp/src/components/AssetCard/AssetCard.tsx
LautaroPetaccio: Great catch, I've just added a check in the container to set the `showRentalBubble` as false value if there's no rental.

webapp/src/components/AssetCard/AssetCard.tsx
LautaroPetaccio: Changing it!

webapp/src/components/AssetCard/AssetCard.container.ts
LautaroPetaccio: You're right, great catch, if I didn't add that, the chips / bubbles will be shown in the home page, changing it!

juanmahidalgo (approved):
LGTM! 🚀

PR #1142, opened 2022-12-14
fix: Fix rentals signature with V different from 27 or 28

LautaroPetaccio:
This PR fixes the issue where rental listings couldn't be accepted by the contracts because of ending in 0 or 1, making the validation fail due to an invalid V value.


webapp/src/modules/rental/sagas.ts
```diff
@@ -314,6 +320,8 @@ function* handleAcceptRentalListingRequest(
       fingerprint
     ]
 
+    console.log('Transaction params', txParams)
```
cazala, line 323: should this be here or did it slip thru?

cazala (approved):

PR #1152, opened 2022-12-19
fix: Enforce signatures to have a V as 27 or 28

### LINKED ISSUE #1145: Signature support is incomplete
## 🎉 Description

In order to support signatures made with wallets that sign data leaving the last two bites as 0 or 1, we've deployed a fix to check the ending value of the signatures to add 27 to make it work. This has a couple of issues, there will be signatures that might need to be splitted and checked and adding 27 to the signature will eventually break the refresh procedure, as the signature is used to look for the rental in the DB.

To fix this:
- Implement in the signature server a mechanism to prohibit the creation of rentals that don't suffice the 27 or 28 requirement.
- Implement in the UI a mechanism to fix the signatures before sending them to the signature server.
- Implement in the UI the mechanism to deal with longer signatures when accepting a listing (although this shouldn't ever be reached).

* [x] 🐛 This is a bug report.

LautaroPetaccio:
Closes #1145

### DIFF
--- a/webapp/src/modules/rental/sagas.spec.ts
+++ b/webapp/src/modules/rental/sagas.spec.ts
@@ -163,7 +163,8 @@ describe('when handling the request action to upsert a rental listing', () => {
           }
         ]
         const expiration = 1234567
-        const signature = 'the-signature'
+        const signature =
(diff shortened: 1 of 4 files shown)


juanmahidalgo (approved):
great job 👏

PR #1155, opened 2022-12-20
fix: Don't make contracts request on startup

### LINKED ISSUE #1136: Evaluate how getContracts can be optimized so it is not called immediately when entering the app
Currently, when entering the app, a request to fetch contracts us being made to the nft-server. 
This request will fetch the subgraphs for absolutely all collections, including the thousands upon thousands of collections. 
This does not only cause lots of queries to the subgraph, but consumes a lot of cpu.
This has to be optimized in some way.

Ideas: 
- The request done in the marketplace is always the same, we can cache the contracts in memory and just fetch for newer contracts on each request.
- Don't fetch all contracts on start, just fetch all on the Settings page were authorizations are required. Ad hoc call for contracts that are required at the moment.

fzavalia:
Closes #1136

### DIFF
--- a/webapp/src/modules/contract/selectors.ts
+++ b/webapp/src/modules/contract/selectors.ts
@@ -10,6 +10,8 @@ export const getState = (state: RootState) => state.contract
 export const getContracts = (state: RootState) => getState(state).data
 export const getError = (state: RootState) => getState(state).error
 export const getLoading = (state: RootState) => getState(state).loading
+export const getHasIncludedMaticCollections = (state: RootState) =>
+  getState(state).hasIncludedMaticCollections
 
 export const getContract = createSelectorCreator(defaultMemoize, isEqual)<
   RootState,
--- a/webapp/src/modules/contract/selectors.ts
+++ b/webapp/src/modules/contract/selectors.ts
@@ -25,7 +25,9 @@ export const getContract = createSelectorCreator(defaultMemoize, isEqual)<
   getLoading,
   getContracts,
   (query: Partial<Contract>, loading: LoadingState, contracts: Contract[]) => {
-    if (isLoadingType(loading, FETCH_CONTRACTS_REQUEST)) return null
+    if (isLoadingType(loading, FETCH_CONTRACTS_REQUEST)) {
+      return null
+    }
 
     const found = contracts.find(contract =>
       Object.keys(query).every(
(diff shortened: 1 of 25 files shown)


webapp/src/modules/contract/selectors.ts
```diff
@@ -33,10 +37,6 @@ export const getContract = createSelectorCreator(defaultMemoize, isEqual)<
       )
     )
 
-    if (!found) {
-      throw new Error(`Contract not found, query=${JSON.stringify(query)}`)
```
fzavalia, line 37: Removed the error because it prevented the solution to be implemented (Would break when checking if the returned value was null)

webapp/src/modules/nft/sagas.ts
```diff
@@ -90,11 +93,24 @@ function* handleFetchNFTRequest(action: FetchNFTRequestAction) {
...
     yield call(getOrWaitForContracts)
 
-    const contract: ReturnType<typeof getContract> = yield select(getContract, {
+    let contract: ReturnType<typeof getContract> = yield select(getContract, {
       address: contractAddress
     })
 
-    if (!contract || !contract.vendor) {
+    if (!contract) {
+      contract = {
```
fzavalia, line 101: This is kind of a weird workaround that is needed when entering a detail page were a single nft is fetched. Originally, it would have failed because the contract didn't exist in the case of collection wearables/emotes. 

As only collection wearable contracts are not present due to the changes made in this pr, I'm assuming that if the contract does not exist, it is a wearable and add it as such to the stored contracts.

webapp/src/modules/vendor/decentraland/contracts.ts
```diff
@@ -0,0 +1,409 @@
+import { NFTCategory, ChainId, Network } from '@dcl/schemas'
+import { Network as AppNetwork } from '../../contract/types'
+
+export function getMarketplaceContracts(network: AppNetwork) {
```
fzavalia, line 4: This is a replica of the contracts found in https://github.com/decentraland/nft-server/blob/a7e32237066cfc407193174b132011864ab0a97a/src/logic/contracts.ts#L75

Replicated here to avoid requesting static contracts from the nft-server

fzavalia:
Closing in favor of https://github.com/decentraland/marketplace/pull/1162

PR #1157, opened 2022-12-20
fix: Remove lenght check from V check

LautaroPetaccio:
This PR removes the length check from the ECDSA V check.


juanmahidalgo (approved):
LGTM

PR #1163, opened 2022-12-23
feat: Add price filter

meelrossi:
Add new price filters to nft sidebar

image


webapp/src/components/FiltersSidebar/PriceFilter/PriceFilter.tsx
```diff
@@ -0,0 +1,31 @@
...
+  function handlePriceChange(newValue: [string, string]) {
+    setValue(newValue);
+    if (timeout) {
+      clearTimeout(timeout);
+    }
+    timeout = setTimeout(() => onChange(newValue), 500)
+  }
+
+  return (
+    <Box header="PRICE" collapsible>
```
juanmahidalgo, line 27: We should translate the header title, something like:
```suggestion
    <Box header={t('filters.price')} collapsible>
```

webapp/src/components/FiltersSidebar/PriceFilter/PriceFilter.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+let timeout: NodeJS.Timeout;
+
+export const PriceFilter = ({ onChange, minPrice, maxPrice }: PriceFilterProps) => {
+  const [value, setValue] = useState<[string, string]>([minPrice, maxPrice])
+
+  useEffect(() => setValue([minPrice, maxPrice]), [minPrice, maxPrice])
+
+  function handlePriceChange(newValue: [string, string]) {
```
juanmahidalgo, line 18: Can we wrap this function with a `useCallback` to avoid re-definitions of it on each render? What do you think?

webapp/src/components/FiltersSidebar/PriceFilter/PriceFilter.tsx
```diff
@@ -0,0 +1,31 @@
...
+    setValue(newValue);
+    if (timeout) {
+      clearTimeout(timeout);
+    }
+    timeout = setTimeout(() => onChange(newValue), 500)
```
juanmahidalgo, line 23: we should clear this timeout on unmounting the component, in order to avoid react errors when executing the timeout of the component gets unmount too quickly.
I think it would be something like the following:
```
useEffect(() => {
     return () => clearTimeout(timeout)
}, [timeout])
```

webapp/src/components/Vendor/decentraland/NFTSidebar/NFTSidebar.tsx
```diff
@@ -7,6 +7,7 @@ import { NFTSections } from '../NFTSections'
 import { NFTLandFilters } from '../NFTLandFilters'
 import { Props } from './NFTSidebar.types'
 import './NFTSidebar.css'
+import { FiltersSidebar } from '../../../FiltersSidebar'
```
juanmahidalgo, line 10: Nitpick: Can we move this import a few lines above please? 🙏

juanmahidalgo (commented):
looks good 👏 ! just a few comments/suggestions

webapp/src/components/FiltersSidebar/PriceFilter/PriceFilter.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import './PriceFilter.css'
+
+export type PriceFilterProps = {
+  minPrice: string,
+  maxPrice: string,
+  onChange: (value: [string, string]) => void
+}
+
+let timeout: NodeJS.Timeout;
```
fzavalia, line 12: What do you think about keeping the timeout inside the component? 
Yo can use:

```
const timeout = useRef<NodeJS.Timeout | null>(null)
...
if (timeout.current) {
  clearTimeout(timeout.current);
}
timeout.current = setTimeout(() => onChange(newValue), 500)
```

Using ref over state is good to prevent rerenders.

fzavalia (changes requested):

fzavalia (approved):

PR #1180, opened 2022-12-28
feat: Search Input for collections now looks for options on demand

fzavalia:
The cut in the middle of the video is because I refreshed the page to show that the selected collection is still displayed.

https://user-images.githubusercontent.com/24811313/209815002-b74a2aa1-f5b5-4a27-86ba-750de6099414.mov

### DIFF
--- a/webapp/src/components/AssetBrowse/AssetBrowse.tsx
+++ b/webapp/src/components/AssetBrowse/AssetBrowse.tsx
@@ -253,7 +253,7 @@ const AssetBrowse = (props: Props) => {
               </Container>
             </div>
           ) : (
-            <NFTFilters isMap={Boolean(isMap)} contracts={contracts} />
+            <NFTFilters isMap={Boolean(isMap)} />
           )}
           {isMap ? (
             <div className="Atlas">
--- a/webapp/src/components/Vendor/NFTFilters/FiltersMenu/FiltersMenu.tsx
+++ b/webapp/src/components/Vendor/NFTFilters/FiltersMenu/FiltersMenu.tsx
@@ -16,6 +16,7 @@ import { WearableGender } from '../../../../modules/nft/wearable/types'
 import { ArrayFilter } from '../ArrayFilter'
 import { SelectFilter } from '../SelectFilter'
 import { Props } from './FiltersMenu.types'
+import { collectionAPI } from '../../../../modules/vendor/decentraland'
 
 export const ALL_FILTER_OPTION = 'ALL'
 
@@ -35,7 +36,7 @@ const FiltersMenu = (props: Props) => {
     onOnlySmartChange
   } = props
 
-  const collectionOptions = useMemo(() => {
+  const defaultCollectionOptions = useMemo(() => {
     return [
       {
         value: ALL_FILTER_OPTION,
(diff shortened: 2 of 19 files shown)


webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -253,7 +253,7 @@ const AssetBrowse = (props: Props) => {
               </Container>
             </div>
           ) : (
-            <NFTFilters isMap={Boolean(isMap)} contracts={contracts} />
```
fzavalia, line 256: Removed the usage of contracts on the whole tree for this component as it is not required anymore

webapp/src/components/Vendor/NFTFilters/FiltersMenu/FiltersMenu.tsx
```diff
@@ -143,8 +113,48 @@ const FiltersMenu = (props: Props) => {
...
+                value: collection.contractAddress
+              }))
+            } catch (e) {
+              console.warn('Could not fetch options')
+              return []
+            }
+          }}
+          fetchOptionFromValue={async value => {
+            try {
+              const { data } = await collectionAPI.fetch({
```
fzavalia, line 139: Could have used fetchOne instead but that one receives a urn as parameter and the contract address is more direct.

webapp/src/components/Vendor/NFTFilters/SelectFilter/SelectFilter.tsx
```diff
@@ -1,27 +1,110 @@
...
+
+      if (!search) {
+        setProvidedOptions(options)
+        return
+      }
+
+      const result = await fetchOptions(search)
+
+      setProvidedOptions([...options, ...result])
+    }, 500)
```
fzavalia, line 74: This number could be set by env but half a second seems enough after the last key inputed to make the request.

webapp/src/modules/routing/search.ts
```diff
@@ -346,12 +346,13 @@ export function getURLParamArray<T extends string>(
   paramName: string,
   validValues: string[] = []
 ) {
-  const param = getURLParam<T>(search, paramName)
-  return param === null
-    ? []
-    : (param
-        .split(SEARCH_ARRAY_PARAM_SEPARATOR)
-        .filter(item => validValues.includes(item as T)) as T[])
+  let params = new URLSearchParams(search).getAll(paramName) as T[]
```
fzavalia, line 349: Refactored this because is wasn't parsing url params like the standard.

webapp/src/modules/routing/search.ts
```diff
@@ -345,6 +345,25 @@ export function getURLParamArray<T extends string>(
...
+
+  return params
+}
+
+// TODO: This is currently using a non standard way of parsing query params
+// This might be because of an old functionality but for example, rarities 
+// from the URL are parsed from rarities=common_uncommon instead of 
+// rarities=common&rarities=uncommon I'll leave it as it is for now to prevent
+// further refactoring but should be changed in the future.
+export function getURLParamArray_nonStandard<T extends string>(
```
fzavalia, line 363: This function has a non standard way to parse query params so I renamed it, I'll create an issue to refactor the way params are written for these cases.

LautaroPetaccio (approved):

PR #1186, opened 2022-12-28
feat: change menu style for categories

meelrossi:
Change style of categories section to match new designs
image


webapp/src/components/Vendor/decentraland/NFTSections/NFTSections.tsx
```diff
@@ -1,11 +1,14 @@
+import { Header } from 'decentraland-ui'
 import React from 'react'
```
fzavalia, line 2: Move react to the top

fzavalia (changes requested):

fzavalia (approved):

PR #1187, opened 2022-12-28
feat: Remove contracts param from nft filters and get or wait for con…

fzavalia:
…tracts function


webapp/src/modules/vendor/decentraland/nft/api.ts
```diff
@@ -121,15 +121,6 @@ class NFTAPI extends BaseAPI {
           queryParams.append('wearableGender', wearableGender)
         }
       }
-      if (filters.contracts) {
-        for (const address of filters.contracts) {
-          if (
-            params.contracts?.some(contract => contract.address === address)
```
fzavalia, line 127: The logic here was flawed and would never work, I removed it because it is not needed in the application.

LautaroPetaccio (approved):

PR #1197, opened 2022-12-30
feat: move play mode filter to pills

meelrossi:
- Change play mode filter to have chips instead of radio buttons as in the future there will be more play modes
- Add mana icon to price filter 
image


fzavalia (approved):
🚀

PR #1205, opened 2023-01-02
fix: Show all bids instead of just 20

### LINKED ISSUE #1199: Only 20 bids are being shown when the user has more than that
For example, this request https://nft-api.decentraland.org/v1/bids?bidder=0xdA59714bfD9019588f8782B66fDbcF68443D8CbA&status=open only responds with 20 results but the address has much more than that according to the https://thegraph.com/hosted-service/subgraph/decentraland/collections-matic-mainnet subgraph.

Update 1: User does not have any bids in ethereum, only on polygon

Update 2: The nft-server adds a expiredAt_gt to the current timestamp if the open status is provided. The user still has more than the 20 returned

Update 3: When first is not provided, the nft-server defaults to 20. https://github.com/decentraland/nft-server/blob/master/src/ports/merger/component.ts#L36

fzavalia:
Closes #1199

### DIFF
--- a/webapp/src/modules/vendor/decentraland/bid/api.ts
+++ b/webapp/src/modules/vendor/decentraland/bid/api.ts
@@ -3,6 +3,8 @@ import { BaseAPI } from 'decentraland-dapps/dist/lib/api'
 import { NFT_SERVER_URL } from '../nft'
 import { retryParams } from '../utils'
 
+const FIRST = '1000'
+
 class BidAPI extends BaseAPI {
   async fetch(options: Record<string, string>): Promise<Bid[]> {
     const queryParams = new URLSearchParams()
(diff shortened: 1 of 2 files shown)


webapp/src/modules/vendor/decentraland/bid/api.ts
```diff
@@ -3,6 +3,8 @@ import { BaseAPI } from 'decentraland-dapps/dist/lib/api'
 import { NFT_SERVER_URL } from '../nft'
 import { retryParams } from '../utils'
 
+const FIRST = '1000'
```
juanmahidalgo, line 6: silly question: Any particular reason to make it a string? could be a number as well

juanmahidalgo (approved):
great catch, LGTM!

webapp/src/modules/vendor/decentraland/bid/api.ts
fzavalia: The object were this value is provided is a Record<string,string> so I'm just avoiding an unnecessary cast

PR #1211, opened 2023-01-03
feat: add selected filters section

meelrossi:
Add new selected filters section

image


webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/Pill/Pill.tsx
```diff
@@ -0,0 +1,15 @@
+import Icon from 'semantic-ui-react/dist/commonjs/elements/Icon'
```
juanmahidalgo, line 1:
```suggestion
import { Icon } from 'decentraland-ui'
```
I believe we can use the one exported on our lib, it's just the semantic-ui one

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useCallback, useEffect } from 'react'
+import { Pill } from './Pill/Pill'
+import { Props } from './SelectedFilters.types'
+import styles from './SelectedFilters.module.css'
+import { getCollectionNameByAddress } from './utils'
+import { Mana } from 'decentraland-ui'
+import classNames from 'classnames'
+
+export const SelectedFilters = ({ browseOptions, isLandSection, onBrowse }: Props)  => {
+  const { rarities, network, onlySmart, contracts, wearableGenders, onlyOnSale, emotePlayMode, minPrice, maxPrice, onlyOnRent } = browseOptions;
```
juanmahidalgo, line 13: Can we run prettier on this file? I think this line should be splitted into multiple ones

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [collection, setCollection] = useState<string | undefined>(undefined)
```
juanmahidalgo, line 14:
```suggestion
  const [collection, setCollection] = useState<string | undefined>()
```
I believe the `undefined` is redundant here, it will be the default value if not specified

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  useEffect(() => {
+    const fetchData = async (contract: string, onlyOnSale: boolean | undefined) => {
+      const collectionName = await getCollectionNameByAddress(contract, onlyOnSale)
+      return collectionName;
+    }
+  
+    if (contracts?.length) {
```
juanmahidalgo, line 22:
```suggestion
    if (contracts?.length && !collectionName) {
```
Should we add the `&& !collectionName` check? Because if we toggle the `onlyOneSale` filter, the `useEffect` will be triggered again and the collection name will be fetched once again, even if we already have the name

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const handleDeleteNetwork = useCallback(() => {
+    onBrowse({ network: undefined })
+  }, [onBrowse])
+
+  const handleDeleteOnlySmart = useCallback(() => {
+    onBrowse({ onlySmart: undefined })
+  }, [onBrowse])
+  
+  const handleDeleteOnlySale = useCallback(() => {
+    onBrowse({ onlySmart: true })
```
juanmahidalgo, line 82:
```suggestion
    onBrowse({ onlyOnSale: false })
```
right?

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }, [onBrowse])
+
+  const handleDeleteGender = useCallback((gender: string) => {
+    onBrowse({ wearableGenders: wearableGenders?.filter((g) => g !== gender) })
+  }, [onBrowse, wearableGenders])
+  
+  const handleDeleteEmotePlayMode = useCallback(() => {
+    onBrowse({ emotePlayMode: undefined })
```
juanmahidalgo, line 90: Should we put some sort of logic with a `filter` here? Since the play mode will be chips as well, I believe we could end up having more than one applied, so removing one should leave the other one on, right?

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const { data } = await collectionAPI.fetch({
+    contractAddress: address,
+    isOnSale: onlyOnSale
+  })
+
+  if (data.length === 0) {
+    return undefined
+  }
+
+  return data[0].name
```
juanmahidalgo, line 13:
```suggestion
  return data[0]?.name
```
Just a suggestion, if you like this one, you can delete the `if (data.length === 0) {` from above since it will return `undefined` if the array has no length

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { collectionAPI } from "../../../../../modules/vendor/decentraland"
+
+export async function getCollectionNameByAddress(address: string, onlyOnSale: boolean | undefined) {
+  const { data } = await collectionAPI.fetch({
+    contractAddress: address,
+    isOnSale: onlyOnSale
```
juanmahidalgo, line 6: If we just want the name, do we need to send the `isOnSale` filter? It might not be necessary

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -0,0 +1,118 @@
+import { useState, useMemo } from 'react'
+import { Rarity } from '@dcl/schemas'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { useCallback, useEffect } from 'react'
+import { Pill } from './Pill/Pill'
+import { Props } from './SelectedFilters.types'
+import styles from './SelectedFilters.module.css'
+import { getCollectionNameByAddress } from './utils'
+import { Mana } from 'decentraland-ui'
+import classNames from 'classnames'
```
juanmahidalgo, line 10: Can we order these imports a little bit 🙏 ?

juanmahidalgo (commented):
this is awesome, looks great, just a few comments/suggestions left before approving

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
meelrossi: the default is true here, but yes i should change the property my bad

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
meelrossi: yesss i didn't have the code yet for play mode to be an array but i just merge it so i will update it

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
juanmahidalgo: true! good catch! yeah the alternative you mentioned is perfect 👍

juanmahidalgo (approved):
Looks great 👏 👏 👏

PR #1222, opened 2023-01-03
fix: Don't allow renting with address zero

### LINKED ISSUE #1177: Prevent adding the zero address as operator when renting
Prevent adding the zero address as operator when renting

LautaroPetaccio:
Disallow rents with the beneficiary as 0x0...0.
Closes #1177

### DIFF
--- a/webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
+++ b/webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
@@ -19,6 +19,8 @@ import { ManaField } from '../../ManaField'
 import { Props } from './ConfirmRentModal.types'
 import styles from './ConfirmRentModal.module.css'
 
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
+
 const ConfirmRentModal = ({
   wallet,
   metadata: { nft, rental, selectedPeriodIndex },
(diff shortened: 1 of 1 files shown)


webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
```diff
@@ -19,6 +19,8 @@ import { ManaField } from '../../ManaField'
 import { Props } from './ConfirmRentModal.types'
 import styles from './ConfirmRentModal.module.css'
 
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
```
fzavalia, line 22: If using ethers, there is a constant for the zero address that we can use instead of hardcoding it like this

fzavalia (changes requested):

webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
fzavalia: https://docs.ethers.org/v5/api/utils/constants/#constants-AddressZero

fzavalia: Might be better to use that instead

LautaroPetaccio: Totally forgot about the ethers constants. Great catch!

fzavalia (approved):

PR #1227, opened 2023-01-04
feat: change body shape filter

meelrossi:
Change body shape filter to include unisex value
image


webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/utils.ts
```diff
@@ -9,3 +10,19 @@ export async function getCollectionByAddress(
 
   return data[0]
 }
+
+export function getGenderFilterLabel(bodyShapes: (WearableGender | GenderFilterOption)[] | undefined): string {
```
juanmahidalgo, line 14: Can we re-use the logic from the other util so save some code here?
maybe something like this:
```
const bodyShapeValue = getBodyShapeValue(bodyShapes)
const labels = {
     [AVAILABLE_FOR_FEMALE]: 'nft_filters.body_shapes.available_for_female',
     [AVAILABLE_FOR_MALE]: 'available_for_female',
}
return labels[bodyShapeValue] || 'nft_filters.body_shapes.all_items'

```
what do you think?

webapp/src/components/Vendor/decentraland/NFTTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -154,13 +157,13 @@ export const SelectedFilters = ({
           onDelete={handleDeleteRarity}
         />
       ) : null}
-      {wearableGenders?.map(gender => (
+      {wearableGenders && wearableGenders.length ? (
```
juanmahidalgo, line 160: this could also be:
```suggestion
      {wearableGenders?.length ? (
```

webapp/src/components/FiltersSidebar/BodyShapeFilter/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { GenderFilterOption } from "@dcl/schemas"
+
+export const AVAILABLE_FOR_MALE = 'AVAILABLE_FOR_MALE'
+export const AVAILABLE_FOR_FEMALE = 'AVAILABLE_FOR_FEMALE'
+
+export function getBodyShapeValue(bodyShapes: GenderFilterOption[] | undefined): string | undefined {
+  const hasUnisex = bodyShapes?.includes(GenderFilterOption.UNISEX);
```
juanmahidalgo, line 7: we might want to include an early return for the `undefined` case so we avoid all the extra computations all the way to the `return undefined` from the bottom, what do you think? 
```
export function getBodyShapeValue(bodyShapes: GenderFilterOption[] | undefined): string | undefined {
     if (!bodyShapes) {
         return undefined
     }
     ....
```

juanmahidalgo (approved):
LGTM! 🙌 🙌🙌

PR #1263, opened 2023-01-07
feat: Add rentals to tile map

LautaroPetaccio:
This PR adds the capability to process rental listing that come from the atlas server and show them painted as light blue (the same color as the ones for sale) in the map. It also changes the popups to show the rental price.
Screen Shot 2023-01-07 at 20 22 54

### DIFF
--- a/webapp/src/components/Atlas/Popup/Popup.tsx
+++ b/webapp/src/components/Atlas/Popup/Popup.tsx
@@ -1,17 +1,30 @@
 import * as React from 'react'
 import { ethers } from 'ethers'
-import { Row, Section, Header } from 'decentraland-ui'
+import { Row, Section, Header, HeaderSubheader } from 'decentraland-ui'
 import { Profile } from 'decentraland-dapps/dist/containers'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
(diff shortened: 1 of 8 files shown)


webapp/src/components/Atlas/Popup/Popup.tsx
```diff
@@ -1,17 +1,30 @@
...
 
 export default class Popup extends React.PureComponent<Props> {
+  subPriceHeader() {
+    const { tile } = this.props
+    if (tile.price && tile.rentalListing) {
+      return t('atlas.for_sale_and_rent')
+    } else if (tile.price) {
+      return t('atlas.for_sale')
+    }
+    return t('atlas.for_rent')
```
fzavalia, line 21: This will always show subheader for_rent even if it doesn't have rentalListing. Is this correct?

fzavalia: Seems that this functions is only called when having price or a rentalListing. Seems correct 👍

fzavalia (approved):

webapp/src/components/Atlas/Popup/Popup.tsx
LautaroPetaccio: That's right!

PR #1269, opened 2023-01-09
Feat/filters mobile

meelrossi:
Add logic to permit filters to be shown correctly in mobile devices

Wearables
image

Emotes
image

Names
image
Land

image


webapp/src/components/AssetFilters/AssetFilters.tsx
```diff
@@ -122,15 +119,15 @@ export const AssetFilters = ({
     [section]
   )
 
-  if (isInLandSection && isNotMobile && isRentalsEnabled) {
-    return (
+  if (isInLandSection) {
```
juanmahidalgo, line 122: Could we add the `Price` filter for the land section as well in this PR? I noticed that is missing, I believe we can release this deliverable with at least that one :), wdyt?

webapp/src/components/AssetTopbar/SelectedFilters/utils.tsx
```diff
@@ -0,0 +1,10 @@
+import { collectionAPI } from '../../../modules/vendor/decentraland'
```
juanmahidalgo, line 1: We can rename this file to `.ts` right? seems like there is no JSX here

juanmahidalgo (approved):
looks awesome 🤩 nice job @meelrossi!

PR #1278, opened 2023-01-10
feat: Map new FF and add the new selector

### LINKED ISSUE #1276: Create feature flag for the Buy NFT with FIAT feature
Create the feature flag here https://features.decentraland.systems/#/features

Add the name and the selector in the files inside this directory so it can be consumed https://github.com/decentraland/marketplace/tree/master/webapp/src/modules/features

kevinszuchet:
Closes #1276


fzavalia:
Can you add a REACT_APP_FF_MARKETPLACE-BUY-NFTS-WITH-FIAT env var to the `.env.example`? The features module provides a way to override a value by reading it from the local environment

webapp/.env.example
```diff
@@ -1,3 +1,4 @@
 REACT_APP_DCL_DEFAULT_ENV=dev
 GEN_STATIC_LOCAL=true
 PUBLIC_URL=/
+REACT_APP_FF_MARKETPLACE_BUY_NFTS_WITH_FIAT=true
```
fzavalia, line 4: https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/features/selectors.ts#L76 enabled value is 1

fzavalia, line 4: Add an empty line at the end

webapp/src/modules/features/selectors.ts
```diff
@@ -85,3 +85,15 @@ export const getIsCampaignBrowserEnabled = (state: RootState) => {
     return false
   }
 }
+
+export const getIsBuyNftsWithFiatEnabled = (state: RootState) => {
+  try {
```
fzavalia, line 90: Instead of the try, you can use the selector here https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/features/selectors.ts#L64

If the flags were not loaded yet return false, if they were, return the value

fzavalia (changes requested):

fzavalia (approved):

PR #1279, opened 2023-01-10
fix: clear search on clear filters and style adjustments

juanmahidalgo:
## Removes the "Search the store/listing" text for names results empty state
image

### DIFF
--- a/webapp/src/components/AssetTopbar/AssetTopbar.module.css
+++ b/webapp/src/components/AssetTopbar/AssetTopbar.module.css
@@ -72,8 +72,8 @@
 }
 
 .clearFilters {
-  font-size: 12px;
-  line-height: 12px;
+  font-size: 13px;
+  line-height: 11px;
   text-transform: uppercase;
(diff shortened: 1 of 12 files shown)


webapp/src/components/AssetTopbar/AssetTopbar.module.css
```diff
@@ -72,8 +72,8 @@
 }
 
 .clearFilters {
-  font-size: 12px;
-  line-height: 12px;
+  font-size: 13px;
+  line-height: 11px;
```
meelrossi, line 76: usually line-height is bigger than  font-size. does this work correctly?

meelrossi (approved):

webapp/src/components/AssetTopbar/AssetTopbar.module.css
juanmahidalgo: mmm works fine, I'm not sure it's the correct thing tho... Updated so they have the same value :)

PR #1282, opened 2023-01-11
feat: show search result in the asset empty state

juanmahidalgo:
image

### DIFF
--- a/webapp/src/components/AssetList/AssetList.tsx
+++ b/webapp/src/components/AssetList/AssetList.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useMemo } from 'react'
 import { Card, Button, Loader } from 'decentraland-ui'
 import { Item, NFTCategory } from '@dcl/schemas'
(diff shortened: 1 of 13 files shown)


meelrossi (approved):

webapp/src/components/AssetList/AssetList.tsx
```diff
@@ -46,6 +47,20 @@ const AssetList = (props: Props) => {
 
   const isLoadingNewPage = isLoading && nfts.length >= PAGE_SIZE
 
+  const getEmptyStateTranslationString = useCallback(() => {
```
meelrossi, line 50: I think this could be done with useMemo and be a variable but its just a suggestion is ok for me :)

juanmahidalgo: Yeah, it works too! I used the `useCallback` so it only gets called when the list is empty (`assets.length` === 0) but the `useMemo` could have an early return to avoid the rest. Overkill thinking tho, all of these computations is super cheap for react. Changed to `useMemo`!

meelrossi (approved):

PR #1287, opened 2023-01-11
fix: do not show selected filters on map view

meelrossi:
- REmove selected filters if the user is in map view
- Fix filter tag appearing when all lands are selected

image


juanmahidalgo (approved):
nice!! 👏

PR #1290, opened 2023-01-11
feat: add categories filter

meelrossi:
Add categories filter to modal

image


webapp/src/components/AssetFilters/AssetFilters.types.ts
```diff
@@ -16,9 +18,11 @@ export type Props = {
   emotePlayMode?: EmotePlayMode[]
   assetType?: AssetType
   collection: string
-  section: Section
+  section?: Section
```
juanmahidalgo, line 21: Is this `?` needed? I think it will always have a value

juanmahidalgo (approved):
Looks really good 🙌 , a few comments/suggestions

webapp/src/components/AssetFilters/AssetFilters.types.ts
meelrossi: as it can come from a Partial<BrowseOptions> it can be undefined here

PR #1298, opened 2023-01-12
fix: Loading spinner position for AssetBrowse component

### LINKED ISSUE #1297: [BUG] The asset loader is not visible when there are too many cards
## 🎉 Description

When there are too many cards, the asset loader spinner is not visible since it's being pushed too far down like the following image shows:

the first page doesn't show the spinner, it needs to be scrolled to be seen

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #1297
## Issue (the spinner was not visible sometimes)
image

##Solution
https://user-images.githubusercontent.com/8763687/212130097-f2bbe020-8b08-42ad-af46-68ece5dc73f7.mov

### DIFF
--- a/webapp/src/components/AssetList/AssetList.tsx
+++ b/webapp/src/components/AssetList/AssetList.tsx
@@ -48,6 +48,12 @@ const AssetList = (props: Props) => {
 
   return (
     <>
+      {isLoading ? (
+        <>
+          <div className="overlay" />
+          <Loader size="massive" active className="asset-loader" />
+        </>
+      ) : null}
       <Card.Group>
         {assets.length > 0
           ? assets.map((assets, index) => (
@@ -58,13 +64,6 @@ const AssetList = (props: Props) => {
               />
             ))
           : null}
-
-        {isLoading ? (
-          <>
-            <div className="overlay" />
-            <Loader size="massive" active />
-          </>
-        ) : null}
       </Card.Group>
 
       {assets.length === 0 && !isLoading ? (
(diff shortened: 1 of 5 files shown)


webapp/src/components/AssetList/AssetList.tsx
```diff
@@ -48,6 +48,12 @@ const AssetList = (props: Props) => {
 
   return (
     <>
+      {isLoading ? (
```
juanmahidalgo, line 51: Moving the overlay + loader outside the `Card.Group` so it's relative to the main container and the `top` position is always the same

webapp/src/components/AssetTopbar/AssetTopbar.tsx
```diff
@@ -157,16 +157,21 @@ export const AssetTopbar = ({
...
         <div className={styles.infoRow}>
-          <div className={styles.countContainer}>
-            <p className={styles.countText}>{getCountText(count, search)}</p>
-            {hasFiltersEnabled && !isMobile && (
-              <button className={styles.clearFilters} onClick={onClearFilters}>
-                {t('filters.clear')}
-              </button>
-            )}
-          </div>
+          {!isLoading ? (
```
juanmahidalgo, line 162: Not related to the main issue of this PR but I added this as a drive-by.
I've moved the `isLoading` checked inside so the component is always render and in the loading state it doesn't move the cards a little bit above. I also added a `min-height` to the container so it's always using the same space

meelrossi (approved):

PR #1299, opened 2023-01-12
feat: Add detail box component in item detail

### LINKED ISSUE #1244: Create Detail Box for wearables/emote items
This will show under the Action Box for items

kevinszuchet:
Closes #1244 

image

image

### DIFF
--- a/webapp/src/modules/asset/utils.ts
+++ b/webapp/src/modules/asset/utils.ts
@@ -1,4 +1,4 @@
-import { NFTCategory, Order, RentalListing } from '@dcl/schemas'
+import { Item, NFTCategory, Order, RentalListing } from '@dcl/schemas'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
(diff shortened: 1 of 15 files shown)


webapp/src/modules/asset/utils.ts
```diff
@@ -101,3 +101,7 @@ export function isOwnedBy(
 export function isNFT(asset: Asset): asset is NFT {
```
fzavalia, line 101: Should't the return type be a boolean?

fzavalia (changes requested):

webapp/src/modules/asset/utils.ts
kevinszuchet: From what I understood, we are telling Typescript that the asset is an NFT, and we can treat it that way. If we return a boolean, we humans know that it is an NFT, but it still typed as an Asset

fzavalia: That's great, forgot that you could do that

fzavalia (dismissed):

fzavalia (approved):

PR #1301, opened 2023-01-12
feat: add asset type filter

meelrossi:
Add assetType filter to modal

image
image


webapp/src/components/Modals/AssetFiltersModal/AssetFiltersModal.tsx
```diff
@@ -35,6 +45,7 @@ const AssetFiltersModal = (props: Props) => {
         />
       </Modal.Header>
       <Modal.Content>
+        {view === View.ACCOUNT ? <AssetTypeFilter onChange={handleAssetTypeChange} assetType={filters.assetType || assetType} /> : null}
```
juanmahidalgo, line 48:
```suggestion
        {view === View.ACCOUNT ||  isWearableOrEmoteSection  ? <AssetTypeFilter onChange={handleAssetTypeChange} assetType={filters.assetType || assetType} /> : null}
```
I believe we need to show this also for the Collectible section (for wearables & emotes).
We could define a new variable like the one above:
```
const isWearableOrEmoteSection = useMemo(() => {
    return (
      view === View.MARKET &&
      (getCategoryFromSection(section) === NFTCategory.EMOTE ||
        getCategoryFromSection(section) === NFTCategory.WEARABLE)
    )
  }, [section, view])
```
For this, we would need to add `section` to the container.
Wdyt?

juanmahidalgo (commented):
looks perfect! Just a little comment of something that I believe is missing.

webapp/src/components/Modals/AssetFiltersModal/AssetFiltersModal.tsx
juanmahidalgo: Ignore my comment 😆 , we have the toggle outside the modal, so we're good :)  
image

juanmahidalgo (approved):
LGTM!! 👏

PR #1309, opened 2023-01-16
fix: filters ui issues

meelrossi:
Fix several ui issues in filter section

- Fix collection filter field overgrowing when filtering with a very large collection name
- Reset button not applied until user press save
- Add rounded corners to menu in mobile view


juanmahidalgo (approved):
LGTM! 🙌

PR #1310, opened 2023-01-16
feat: add category dropdown collapsible option

meelrossi:
- collapse dropdown menu item when clicked twice
- if the section doesn't change do not call onChange again
https://user-images.githubusercontent.com/11800206/212762453-985ba4b7-cad2-448e-b22f-411ab1f214df.mov


juanmahidalgo (approved):
LGTM! 👏

PR #1313, opened 2023-01-17
feat: Open explanation modal only the 1st time

### LINKED ISSUE #1260: When the user clicks on the buy with card button on the buy with card page, open the buy with card explanation modal
This should only happen once for the user.

UPDATE: Lets make it so the modal stops appearing the first time the user makes a transak purchase without waiting for it to be completed. Best to do it like this because the user might try to buy something else in the meantime and if the previous purchase is not finished, it will still see the modal.

kevinszuchet:
Closes #1260

### DIFF
--- a/webapp/src/components/Modals/BuyWithCardExplanationModal/BuyWithCardExplanationModal.tsx
+++ b/webapp/src/components/Modals/BuyWithCardExplanationModal/BuyWithCardExplanationModal.tsx
@@ -1,71 +1,52 @@
-import React, { useCallback, useEffect } from 'react'
+import React from 'react'
 import { Button, Modal, ModalNavigation } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Props } from './BuyWithCardExplanationModal.types'
 import styles from './BuyWithCardExplanationModal.module.css'
 
-const BUY_NFTS_WITH_CARD_EXPLANATION_POPUP_KEY =
-  'buy-nfts-with-card-explanation-popup-key'
-
-const BuyWithCardExplanationModal = ({ onContinue, onClose }: Props) => {
-  const handleOnClose = useCallback(() => {
-    localStorage.setItem(BUY_NFTS_WITH_CARD_EXPLANATION_POPUP_KEY, 'true')
-    onClose()
-  }, [onClose])
-
-  useEffect(() => {
-    if (
-      localStorage.getItem(BUY_NFTS_WITH_CARD_EXPLANATION_POPUP_KEY) === 'true'
-    ) {
-      // TODO(buy nfts with card): Dispatch the opening of Transak widget
-      onClose()
-    }
-  }, [onClose])
-
-  return (
-    <Modal open className={styles.buyWithCardExplanationModal}>
-      <ModalNavigation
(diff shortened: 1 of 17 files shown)


webapp/src/components/Modals/BuyWithCardExplanationModal/BuyWithCardExplanationModal.tsx
```diff
@@ -1,52 +1,71 @@
...
+  const handleOnClose = useCallback(() => {
+    localStorage.setItem(BUY_NFTS_WITH_CARD_EXPLANATION_POPUP_KEY, 'true')
+    onClose()
+  }, [onClose])
+
+  useEffect(() => {
+    if (
+      localStorage.getItem(BUY_NFTS_WITH_CARD_EXPLANATION_POPUP_KEY) === 'true'
+    ) {
+      // TODO(buy nfts with card): Dispatch the opening of Transak widget
```
fzavalia, line 20: I don't think the component should be responsible of handling this. The one opening the modal should be responsible of doing one thing or the other.

fzavalia (changes requested):

webapp/src/modules/item/actions.ts
```diff
@@ -54,9 +54,9 @@ export type FetchTrendingItemsFailureAction = ReturnType<
 >
 
 // Buy Item
-export const BUY_ITEM_REQUEST = '[Request] Buy item'
-export const BUY_ITEM_SUCCESS = '[Success] Buy item'
-export const BUY_ITEM_FAILURE = '[Failure] Buy item'
+export const BUY_ITEM_REQUEST = '[Request] Buy Item'
```
fzavalia, line 57: Is this action being tracked by segment? I think changing the name can mess things up.

webapp/src/modules/item/actions.ts
```diff
@@ -79,6 +79,13 @@ export type BuyItemRequestAction = ReturnType<typeof buyItemRequest>
 export type BuyItemSuccessAction = ReturnType<typeof buyItemSuccess>
 export type BuyItemFailureAction = ReturnType<typeof buyItemFailure>
 
+// Buy Item With Card
+export const BUY_ITEM_WITH_CARD = 'Buy Item with Card'
+
+export const buyItemWithCard = () => action(BUY_ITEM_WITH_CARD)
```
fzavalia, line 85: This will probably end up being a request, failure, success action. Remember to change it when it is required.

webapp/src/modules/item/sagas.spec.ts
```diff
@@ -82,6 +84,30 @@ describe('when handling the buy items request action', () => {
...
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  describe('when the explanation modal is shown and the user closes it', () => {
+    it('should not set the item in the local storage to show the modal again later', () => {
+      return expectSaga(itemSaga)
+        .provide([[matchers.call.fn(localStorage.getItem), null]])
```
fzavalia, line 99: You should use call directly and provide the parameter that is expecting instead of using matchers.

fzavalia (changes requested):

fzavalia (changes requested):

fzavalia (approved):

PR #1322, opened 2023-01-18
feat: Use Transak to buy nfts with card

### LINKED ISSUE #1261: Integrate Transak buy with FIAT into the dapp
When clicking on continue on the buy with card explanation modal, trigger the purchase with transak and handle the flow.

If the buy with card explanation modal is not needed because it is not the first time purchasing an item with card for the user, trigger transak when clicking on buy with card button on the buy with card page.

UPDATE:

In order to open transak in the buy nft flow, you have to open this URL: 

https://global.transak.com/?apiKey={apiKey}&network=polygon&contractAddress={collectionAddress}&tokenId={itemId|nftId}&tradeType={primary|secondary}&productsAvailed=BUY&isNFT=true

kevinszuchet:
Closes #1261


fzavalia (approved):

webapp/package.json
```diff
@@ -6,6 +6,7 @@
     "@dcl/schemas": "^5.28.0",
     "@dcl/ui-env": "^1.2.0",
     "@ethersproject/providers": "^5.6.2",
+    "@transak/transak-sdk": "^1.2.3",
```
fzavalia, line 9: This can be removed as it is installed in dapps. Remember to also revert the package-lock file.

fzavalia (changes requested):

webapp/src/modules/transak/actions.ts
```diff
@@ -0,0 +1,7 @@
+import { action } from 'typesafe-actions'
+import { Asset } from '../asset/types'
+
+// Open Transak
+export const OPEN_TRANSAK = 'Open Transak'
+export const openTransak = (asset: Asset) => action(OPEN_TRANSAK, { asset })
```
fzavalia, line 6: Why is this action here and not on dapps?

kevinszuchet: Until now, this is an action dispatched only by the marketplace. Maybe, in the following tasks, we should extend even more the Transak module so this action, and others such as going to a specific status page after purchasing an NFT, can be generic for other projects.

fzavalia (approved):

PR #1331, opened 2023-01-20
fix: ui issues in full width mode

meelrossi:
This pr fixes several issues in full width mode:
- The View All should be alligned to the right as it was before (now with the new width, but same proportion). In the Parcels and Estates module and Names it's well alligned, it needs fixing in the latest listings and newest items
-  I'm seeing the Clear filters CTA in this page but there are no filters applied. And if I click on clear filters, it loads again but nothiing changes (it keeps showing me the Clear filters CTA)
- The items in this module should be centered
- Activity Screen: Column width in Activity is at 1064px should 1320px
- Footer still in the old format, it's not full width and it should
- Change the wording of the place holder for "Search Collections"
- Change the length of the search bar in the LAND section (grid view) so that the toggle to change between grid/map is outside the search box, here is an example of how it should look like


juanmahidalgo (approved):
LGTM!

PR #1332, opened 2023-01-23
feat: mobile links to builder not available

### LINKED ISSUE #1316: RentConfirmedModal should not allow mobile links to the builder
The first and third option should not be available on mobile as the builder is not available. It should be marked with a gray shadow and a disclaimer below that button saying "Only available in Desktop"

flobarreto:
closes #1316 

image


fzavalia:
image

The icon </> looks bigger than the rest on mobile, and the text is not aligned, maybe we should do something to align everything

flobarreto:
image

what about now?

webapp/src/components/Modals/RentConfirmedModal/RentConfirmedModal.tsx
```diff
@@ -50,30 +53,55 @@ const RentConfirmedModal = ({
...
-                      `rental_modal.rent_confirmed_step.action_${i + 1}.title`
-                    )}
-                  </span>
-                  <span className={styles.ctaSubtitle}>
-                    {t(
-                      `rental_modal.rent_confirmed_step.action_${i +
-                        1}.subtitle`
-                    )}
-                  </span>
+            {CTAs.map((cta, i) =>
```
fzavalia, line 56: I'm not a fan of doing a map of the ctas when they are not dynamic. Can we write them directly into the jsx and disable the corresponding one if the viewport is mobile?

fzavalia (changes requested):

webapp/src/components/Modals/RentConfirmedModal/RentConfirmedModal.tsx
LautaroPetaccio: Be aware that there are only a few things that change in the code and they can be highly abstracted. We can create sub-components to keep the code abstraction though.

webapp/src/components/Modals/RentConfirmedModal/CTA.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { CTAProps } from './RentConfirmedModal.types'
+import styles from './RentConfirmedModal.module.css'
+import classNames from 'classnames'
+
+type Props = {
+  cta: CTAProps
+}
+
+const CTA = ({ cta }: Props) => {
```
fzavalia, line 12: I think it would be cleaner that this component received as props: to, icon and disabled. Receiving an object with its props seems unnecessary and the index is kind of abstract for this component.

if disabled is true, do the same as you are doing here as if it were  isMobileView.

webapp/src/components/Modals/RentConfirmedModal/CTA.tsx
```diff
@@ -0,0 +1,52 @@
+import { useMobileMediaQuery } from 'decentraland-ui'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+
+import { CTAProps } from './RentConfirmedModal.types'
+import styles from './RentConfirmedModal.module.css'
```
fzavalia, line 5: Styles used here should belong to a CTA.module.css file

webapp/src/components/Modals/RentConfirmedModal/CTA.tsx
```diff
@@ -0,0 +1,52 @@
+import { useMobileMediaQuery } from 'decentraland-ui'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+
+import { CTAProps } from './RentConfirmedModal.types'
```
fzavalia, line 4: Types used here should belong to a CTA.types.ts file

webapp/src/components/Modals/RentConfirmedModal/RentConfirmedModal.tsx
```diff
@@ -1,27 +1,30 @@
...
 import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Modal } from 'decentraland-dapps/dist/containers'
 import { builderUrl } from '../../../lib/environment'
-import { Props } from './RentConfirmedModal.types'
+import { CTAProps, Props } from './RentConfirmedModal.types'
 import styles from './RentConfirmedModal.module.css'
+import CTA from './CTA'
 
-const CTAs = [
+const CTAs: CTAProps[] = [
```
fzavalia, line 12: This can be remove and placed inside the render like:

```
<CTA to={builderUrl} icon="build-more" disabled={isMobileView} />
<CTA to="https://docs.decentraland.org/creator/development-guide/coding-scenes/" icon="get-creative" />
<CTA to={`${builderUrl}/land`} icon="manage-land" disabled={isMobileView} />
```

fzavalia (changes requested):

webapp/src/components/Modals/RentConfirmedModal/CTA/CTA.tsx
```diff
@@ -0,0 +1,41 @@
+import classNames from 'classnames'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+
+import styles from './CTA.module.css'
+import { Props } from './CTA.types'
+
+const CTA = ({ to, name, isDisabledOnMobile }: Props) => {
```
fzavalia, line 7: ❤️

webapp/src/components/Modals/RentConfirmedModal/CTA/CTA.tsx
```diff
@@ -0,0 +1,41 @@
...
+        )}
+      </div>
+    </>
+  )
+
+  return isDisabledOnMobile ? (
+    <div className={styles.mobileAvailabilityContainer}>{renderContent()}</div>
+  ) : (
+    <a
+      key={name}
```
fzavalia, line 30: This is not handled in a map so the key is unnecessary

webapp/src/components/Modals/RentConfirmedModal/RentConfirmedModal.types.ts
```diff
@@ -12,3 +12,9 @@ export type Props = Omit<ModalProps, 'metadata'> & {
 
 export type OwnProps = Pick<Props, 'metadata'>
 export type MapStateProps = Props
+
+export type CTAProps = {
```
fzavalia, line 16: This can be deleted

fzavalia (changes requested):

fzavalia (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #1364, opened 2023-01-30
feat: start commit to set the ESLint correctly

juanmahidalgo:


webapp/tsconfig.json
```diff
@@ -14,8 +14,8 @@
     "forceConsistentCasingInFileNames": true,
     "module": "esnext",
     "moduleResolution": "node",
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
kevinszuchet, line 18: Both `noUnusedLocals` and `noUnusedParameters` should be true, right?

PR #1389, opened 2023-02-07
feat: Add the Estate Size Filter

### LINKED ISSUE #1353: [5.6] Add estate size filter
## 🎉 Description

Add new state size filter
- Add state min size and max size query params to nft endpoint in nft server (there is already a property searchEstateSize that can be use to filter)
- Add new component in UI to filter sizes using this new query params
- This should only be shown when the user is in estate section
- Add new endpoint that brings all the different states sizes to show in the inventory chart

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #1353

## Available for rent

image

## All Lands

image


webapp/src/components/AssetFilters/AssetFilters.tsx
```diff
@@ -12,6 +12,8 @@ import { isLandSection } from '../../modules/ui/utils'
 import { LANDFilters } from '../Vendor/decentraland/types'
 import { Menu } from '../Menu'
 import PriceFilter from './PriceFilter'
+import EstateSizeFilter from './EstateSizeFilter'
+import { Sections } from '../../modules/routing/types'
```
kevinszuchet, line 16: Would you move this import a few places up? 🙏

webapp/src/components/AssetFilters/EstateSizeFilter/EstateSizeFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+      ) : (
+        t('filters.estate_size.label')
+      ),
+    [minPrice, maxPrice, network, isMobileOrTablet]
+  )
+
+  const fetcher = useCallback(async () => {
+    if (landStatus === LANDFilters.ONLY_FOR_RENT) {
+      // for rents, we don't have the data yet, so let's just resolve the promise with an empty object so the chart is not rendered
+      return Promise.resolve({})
```
kevinszuchet, line 41: Is the `Promise.resolve` mandatory? Why not return the object itself?

webapp/src/components/AssetFilters/Inventory/Inventory.tsx
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useState } from 'react'
+import { BarChart } from 'decentraland-ui'
+import { Props } from './Inventory.types'
+import { Network } from '@dcl/schemas/dist/dapps/network'
```
kevinszuchet, line 4: Would you move this one place up? 🙏

webapp/src/components/AssetFilters/Inventory/index.ts
```diff
@@ -0,0 +1,3 @@
+import PriceFilter from './Inventory.container'
+
+export default PriceFilter
```
kevinszuchet, line 3: Would you mind replacing `PriceFilter` with `Inventory`?

webapp/src/modules/routing/selectors.ts
```diff
@@ -258,6 +258,16 @@ export const getMaxPrice = createSelector<RootState, string, string>(
...
+export const getMinEstateSize = createSelector<RootState, string, string>(
+  getRouterSearch,
+  search => (getURLParam(search, 'minEstateSize') as string) || ''
+)
+
+export const getMaxEstateSize = createSelector<RootState, string, string>(
+  getRouterSearch,
+  search => (getURLParam(search, 'maxEstateSize') as string) || ''
+)
+
```
kevinszuchet, line 270: WDYT about writing something like:

```js
const getStringParamFromURL = (param: string) =>
  createSelector<RootState, string, string>(
    getRouterSearch,
    search => (getURLParam(search, param) as string) || ''
  )

export const getMinPrice = getStringParamFromURL('minPrice')
export const getMaxPrice = getStringParamFromURL('maxPrice')
export const getMinEstateSize = getStringParamFromURL('minEstateSize')
export const getMaxEstateSize = getStringParamFromURL('maxEstateSize')
```

webapp/src/modules/translation/locales/zh.json
```diff
@@ -158,6 +158,12 @@
     "size": "大小",
     "price": "价格",
     "price_min_greater_max": "最高价格应高于最低价格",
+    "estate_size": {
+      "label": "庄园规模",
+      "min_label": "最小（土地）",
+      "max_label": "马克斯（陆地）",
+      "error": "最大地产规模应高于最小规模"
+    },
```
kevinszuchet, line 166: I think the translations for the `nft_filters.estate_size` property are missing. In the `en.json` and `es.json` we also have the following props:

```json
   "estate_size": {
      "title": "Price",
      "all_sizes": "All sizes",
      "more_than_size": "More than {size} parcels",
      "less_than_size": "Less than {size} parcels",
      "size_between": "{min} to {max} parcels"
    },
```

webapp/src/utils/filters.tsx
```diff
@@ -52,6 +52,31 @@ export function getPriceLabel(
...
+
+  if (min && !max) {
+    return t('nft_filters.estate_size.more_than_size', {
+      size: priceFormatter.format(Number(min))
+    })
+  }
+
+  if (max && !min) {
+    return t('nft_filters.estate_size.less_than_size', {
+      price: priceFormatter.format(Number(max))
```
kevinszuchet, line 70: I think the expected property in the translation is called `size` instead of `price`.

webapp/src/components/AssetFilters/AssetFilters.tsx
juanmahidalgo: done!

webapp/src/components/AssetFilters/EstateSizeFilter/EstateSizeFilter.tsx
juanmahidalgo: I thought typescript was going to complain since the function is `async`, but returning `{}` works just fine 👍 updated

webapp/src/components/AssetFilters/Inventory/Inventory.tsx
juanmahidalgo: done!

webapp/src/components/AssetFilters/Inventory/index.ts
juanmahidalgo: copy pasta 😂 done!

webapp/src/modules/routing/selectors.ts
juanmahidalgo: I think we would loose the memoization that comes out of the box using the `createSelector` but I'd have to test it out to be sure

webapp/src/utils/filters.tsx
juanmahidalgo: fixed! good catch :)

webapp/src/modules/routing/selectors.ts
kevinszuchet: You're right. Maybe it's not worth the refactor.

kevinszuchet (approved):

PR #1397, opened 2023-02-07
feat: add distance search properties

meelrossi:
- Add two new properties to be able to search land by its distance to an important point. 
-- `searchDistanceToPlaza`: will have the distance the parcel/estate to the nearest plaza. If theere are no plaza in a distance less than 9 tiles away then this values will be null
-- `searchAdjacentToRoad`: will be true if the parcel is next to a road and false if it isn't. In case of estates it will be true if any parcel that belongs to the state is next to a road.

- Add `proximity.ts` file that includes all parcel coords with its distances to plaza and road. This is to save execution time when doing the indexing of the graph. We are only saving the parcels which are adjacent to a road or have a min distance to a plaza that is smaller than 9 tiles. This is to reduce the size of the proximity file as well as to not exceed the max size of the map that was causing the subgraph to not be able ot compile correctly


indexer/src/handlers/estate.ts
```diff
@@ -1,152 +1,202 @@
...
+  getParcelDistances,
+  shouldRecalculateMinDistance,
+} from "../modules/estate";
+import * as categories from "../modules/category/categories";
+import * as addresses from "../data/addresses";
 
 export function handleCreateEstate(event: CreateEstate): void {
-  let estateId = event.params._estateId.toString()
-  let data = event.params._data.toString()
+  let estateId = event.params._estateId.toString();
```
juanmahidalgo, line 26: can we revert the prettier changes in these files? it added a lof of `;` 😅

juanmahidalgo (dismissed):
looks great @meelrossi! great job 👏

juanmahidalgo (approved):

PR #1406, opened 2023-02-09
fix: Remove estateSize filters on the Clear All Filters action

### LINKED ISSUE #1405: Remove decimals for BarChart at small ranges & fix clear all filters for estate sizes
## 🎉 Description

As the title stands, let's add a way to fix the decimals for small ranges, since for estate sizes we don't want any decimals.
Also, fix the clear all filters that is currently not cleaning the estate sizes filter.

* [X] 🐛 This is a bug report.

juanmahidalgo:
and also remove the decimals for the small ranges (<10)

Closes #1405

### DIFF
--- a/webapp/src/components/AssetFilters/Inventory/Inventory.tsx
+++ b/webapp/src/components/AssetFilters/Inventory/Inventory.tsx
@@ -12,7 +12,8 @@ export const Inventory = ({
   network,
   upperBound,
   errorMessage,
-  onChange
+  onChange,
+  ...rest
 }: Props) => {
   const [isLoading, setIsLoading] = useState(false)
   const [data, setData] = useState<Record<string, number>>()
@@ -50,6 +51,7 @@ export const Inventory = ({
       network={network}
       onChange={onChange}
       errorMessage={errorMessage}
+      {...rest}
     />
   )
 }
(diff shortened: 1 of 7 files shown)


webapp/src/components/AssetFilters/Inventory/Inventory.tsx
```diff
@@ -50,6 +51,7 @@ export const Inventory = ({
       network={network}
       onChange={onChange}
       errorMessage={errorMessage}
+      {...rest}
```
kevinszuchet, line 54: Is rest an object with the new prop `rangeDecimals`? Why not sending directly the new prop to keep track of what we are passing to the `BarChart` component?

juanmahidalgo: Since I added `& BarChartProps` to the `Inventory` props, I prepared it so we can send any of the BarChart props through it and they can get to the `BarChart` as props. In this case, it's just the `rangeDecimals`, but I wanted to leave it prepared so if we have to send more props, we don't have to add them manually. Does it make any sense?

kevinszuchet: Yes, it SGTM.

kevinszuchet (approved):

PR #1423, opened 2023-02-14
feat: add distance to plaza and adjacent to road filter

meelrossi:
Add new location filter section
Add testing for location filter component and selected filters component
image


webapp/src/components/AssetFilters/LocationFilter/LocationFilter.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+      getByRole('checkbox', { name: t('nft_filters.distance_to_plaza.title') })
+    ).toBeInTheDocument()
+  })
+
+  test('should call onAdjacentToRoadChange callback when toggle changes', async () => {
+    const onAdjacentToRoadChangeMock = jest.fn()
+    const { getByTestId } = renderLocationFilter({
+      onAdjacentToRoadChange: onAdjacentToRoadChangeMock
+    })
+    await userEvent.click(getByTestId('adjacent-to-road-toggle'))
```
juanmahidalgo, line 36: Can we get this element by `role`? I _believe_ the `getByTestId` should be the one if there's no alternative

webapp/src/components/AssetFilters/LocationFilter/LocationFilter.tsx
```diff
@@ -0,0 +1,140 @@
...
+    (_evt, data) => {
+      onDistanceToPlazaChange(data)
+    },
+    [onDistanceToPlazaChange]
+  )
+
+  const handleToggleDistanceFilter = useCallback(
+    (_evt, props: CheckboxProps) => {
+      if (!!props.checked) {
+        onDistanceToPlazaChange(['2', '10'])
```
juanmahidalgo, line 49: nit-pick: can we define this 2 values in two diferente constants? so it's clear that are the default values :)

webapp/src/components/AssetFilters/LocationFilter/LocationFilter.tsx
```diff
@@ -0,0 +1,140 @@
...
+      } else {
+        onDistanceToPlazaChange(['', ''])
+      }
+    },
+    [onDistanceToPlazaChange]
+  )
+
+  const locationSelectionText = useMemo(() => {
+    if (!adjacentToRoad && !minDistanceToPlaza && !maxDistanceToPlaza) {
+      return 'All locations'
```
juanmahidalgo, line 59: We need to translate it, right? wrapping with `t`

webapp/src/components/AssetFilters/LocationFilter/LocationFilter.tsx
```diff
@@ -0,0 +1,140 @@
...
+              : undefined
+          }
+          valueTo={
+            maxDistanceToPlaza
+              ? Number.parseFloat(maxDistanceToPlaza)
+              : undefined
+          }
+          range
+          header=""
+          label="Distance from a plaza (in parcels)"
```
juanmahidalgo, line 134: same, we need to translate this one :)

webapp/src/components/AssetFilters/LocationFilter/LocationFilter.spec.tsx
meelrossi: In this case if we "click" on the element with role checkbox nothing will happen because it is "hidden". React testing library simulates user behavior, as that element doesn't "appear" on the screen then clicking it won't do anything. In this case we need to click on the wrapper that is a div, so i think by role will be hard

juanmahidalgo (approved):
looks great :) thanks for adding the UI tests 👏

juanmahidalgo (approved):
looks great 💯

PR #1426, opened 2023-02-15
feat: Introduce the Creators filter for wearables on Polygon

### LINKED ISSUE #1414: [5.9] Creators filter
## 🎉 Description

Add a new filter in the new left panel section for creators. It should be get stacked when selecting multiple creators. It should work as the collection one does, only that this one can have multiple values at the same time.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
## Loading state

image

## Top creators

image

## With a search

image

Closes #1414


webapp/src/modules/account/sagas.spec.ts
```diff
@@ -129,3 +146,176 @@ describe('when handling the request to fetch account metrics', () => {
     })
   })
 })
+
+describe('when handling the request to fetch creators accounts', () => {
+  let search: string
+  let accounts: Account[]
+  let addresses: string[]
+
+  describe('when a call to the nftAPI fails', () => {
```
LautaroPetaccio, line 155: What if we fuse the two describes into one?
```suggestion
  describe('when the request with a search term fails while fetching the NFT API', () => {
```

webapp/src/modules/account/sagas.spec.ts
```diff
@@ -129,3 +146,176 @@ describe('when handling the request to fetch account metrics', () => {
...
+            ]
+          ])
+          .put(fetchCreatorsAccountFailure(search, error))
+          .dispatch(fetchCreatorsAccountRequest(search))
+          .silentRun()
+      })
+    })
+  })
+
+  describe('when a call to the accountAPI fails', () => {
```
LautaroPetaccio, line 184: What if we fuse the two describes into one?
```suggestion
  describe('when the request without a search term fails while fetching the Accounts API', () => {
```

webapp/src/modules/account/sagas.spec.ts
```diff
@@ -129,3 +146,176 @@ describe('when handling the request to fetch account metrics', () => {
...
+    const error = 'request failed with error'
+    let filters: AccountFilters
+
+    describe('when the call with search with a term', () => {
```
LautaroPetaccio, line 188: There's no search term here, right?

webapp/src/modules/account/sagas.spec.ts
```diff
@@ -129,3 +146,176 @@ describe('when handling the request to fetch account metrics', () => {
...
+        nftResults = [
+          { nft: { owner: addresses[0] } } as NFTResult,
+          { nft: { owner: addresses[1] } } as NFTResult
+        ]
+        profiles = [
+          { avatars: [{ ethAddress: addresses[0] }] } as Profile,
+          { avatars: [{ ethAddress: addresses[1] }] } as Profile
+        ]
+      })
+      it('should fetch the ens that match the search term using the nftAPI and then and their profiles using the catalyst lambdas and accounts using the nftAPI', () => {
```
LautaroPetaccio, line 303: Although we might have written the `its` wrongly, the expectation from the sagas is to eventually put an action, what do you think about adding this little peace of text?
```suggestion
      it('should fetch the ens that match the search term using the nftAPI and then and their profiles using the catalyst lambdas and accounts using the nftAPI and put the success action with the creators\' profiles', () => {
```

webapp/src/components/AssetFilters/CreatorsFilter/CreatorsFilter.tsx
```diff
@@ -0,0 +1,187 @@
...
+  )
+
+  useEffect(() => {
+    const fetchData = async (addresses: string[]) => {
+      return await getCreatorsByAddress(addresses)
+    }
+
+    if (creators?.length) {
+      setIsFetchingNames(true)
+      fetchData(creators).then(creators => {
```
LautaroPetaccio, line 45: I might be missing something, but isn't it possible to do:
```suggestion
      getCreatorsByAddress(creators).then(creators => {
```

webapp/src/components/AssetTopbar/SelectedFilters/SelectedFilters.tsx
```diff
@@ -55,6 +62,18 @@ export const SelectedFilters = ({
     }
   }, [contracts, onlyOnSale, collection?.address])
 
+  useEffect(() => {
+    const fetchData = async (addresses: string[]) => {
+      return await getCreatorsByAddress(addresses)
+    }
+
+    if (creators?.length) {
+      fetchData(creators).then(creators => setSelectedCreators(creators))
```
LautaroPetaccio, line 71: Would this be fetched twice if some creators get loaded by the URL? I mean, we would be loading them in the `CreatorsFilters.tsx` and in this component at the same time, right?

LautaroPetaccio (changes requested):
The feature looks amazing! 🚀  I left some comments about some tests, but it looks great!

LautaroPetaccio:
There's a slight misalignment with the `x` in the filter:
Screenshot 2023-02-17 at 17 33 56
And, if you write something and then click outside of the field, the placeholder changes color to the primary color instead of the grayed out one.

webapp/src/modules/account/sagas.ts
```diff
@@ -1,21 +1,97 @@
...
+    try {
+      let addresses: Set<string> = new Set()
+      let accounts: Account[] | undefined = undefined
+      if (search) {
+        const { data }: { data: NFTResult[] } = yield call([nftAPI, 'fetch'], {
+          category: NFTCategory.ENS,
+          search,
+          first: DEFAULT_FIRST_VALUE,
+          skip: DEFAULT_SKIP_VALUE
+        })
```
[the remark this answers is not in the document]
juanmahidalgo: The user will only see the firsts `DEFAULT_FIRST_VALUE` matches. If what they want is not there, they can continue typing to narrow it down. It's set to `20`, I could increase it a little bit more but I think it should be enough. We might want to paginate it in the future if we want it to scale but for the time being I think this is good enough

webapp/src/modules/account/actions.ts
```diff
@@ -29,3 +29,32 @@ export type FetchAccountMetricsSuccessAction = ReturnType<
...
+  '[Success] Fetch creators accounts'
+export const FETCH_CREATORS_ACCOUNT_FAILURE =
+  '[Failure] Fetch creators accounts'
+
+export const fetchCreatorsAccountRequest = (search: string) =>
+  action(FETCH_CREATORS_ACCOUNT_REQUEST, { search })
+export const fetchCreatorsAccountSuccess = (
+  search: string,
+  creatorAccounts: CreatorAccount[]
+  // accountMetrics: Record<Network, AccountMetrics[]>
```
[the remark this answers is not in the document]
juanmahidalgo: uppps, yeah, good catch :)

webapp/src/modules/routing/selectors.spec.ts
```diff
@@ -340,6 +349,20 @@ describe('when there is a maxEstateSize defined', () => {
...
+  beforeEach(() => {
+    anAddress = '0xAddress'
+    url = `creators=${anAddress}`
+  })
+
+  it('should return the value', () => {
+    expect(getCreators.resultFunc(url)).toEqual([anAddress])
+  })
+})
+
```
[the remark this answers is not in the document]
juanmahidalgo: good idea! added

webapp/src/modules/account/utils.ts
```diff
@@ -17,3 +18,19 @@ function addStrings(a: string, b: string) {
     .add(b)
     .toString()
 }
+
+export function fromProfilesToCreators(
+  profiles: Profile[],
+  accounts: Account[]
+): CreatorAccount[] {
+  return profiles
```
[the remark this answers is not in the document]
juanmahidalgo: absolutely, added!

webapp/src/components/AssetFilters/CreatorsFilter/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { Profile } from '@dcl/schemas'
+import { CatalystClient } from 'dcl-catalyst-client'
+import { peerUrl } from '../../../lib/environment'
+
+export async function getCreatorsByAddress(address: string[]) {
```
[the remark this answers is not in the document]
juanmahidalgo: absolutely

webapp/src/modules/account/sagas.spec.ts
juanmahidalgo: right, text updated

webapp/src/components/AssetFilters/CreatorsFilter/CreatorsFilter.tsx
juanmahidalgo: true 🤦 good call

LautaroPetaccio (dismissed):

PR #1428, opened 2023-02-16
feat: Poll pending purchases from Transak

kevinszuchet:
Related to https://github.com/decentraland/decentraland-dapps/pull/370


webapp/src/modules/order/sagas.ts
```diff
@@ -149,26 +156,60 @@ function* handleExecuteOrderWithCardRequest(
...
+}
+
 function* handleSetNftPurchaseWithCard(action: SetPurchaseAction) {
   try {
     const { purchase } = action.payload
+    const { status, txHash } = purchase
 
-    if (!isManaPurchase(purchase) && purchase.nft.tokenId) {
+    if (
+      !isManaPurchase(purchase) &&
```
fzavalia, line 176: Isn't there a isNFTPurchase?

webapp/src/modules/order/sagas.ts
```diff
@@ -149,26 +156,60 @@ function* handleExecuteOrderWithCardRequest(
...
+        }: {
+          success: FetchNFTSuccessAction
+          failure: FetchNFTFailureAction
+        } = yield race({
+          success: take(FETCH_NFT_SUCCESS),
+          failure: take(FETCH_NFT_FAILURE)
+        })
+
+        if (success) {
+          nft = yield call(getNFT, contractAddress, tokenId)
```
fzavalia, line 206: I think the FetchNFTSuccessAction has the nft in the action payload, no need to get it to the store

kevinszuchet: Nice catch, thanks!

fzavalia (dismissed):

fzavalia (approved):

fzavalia (approved):

PR #1429, opened 2023-02-17
fix: Remove scientific notation in small numbers

kevinszuchet:
It works both for small numbers and for the max amount of MANA that a user could have.


LautaroPetaccio (approved):
Looks great 👍

PR #1446, opened 2023-02-28
fix: period dropdown placeholder

### LINKED ISSUE #1148: Add placeholder for the rental period dropdown instead of having a default value
The current implementation always shows a default value for the periods dropdown:

But it should show only a placeholder with the "Select period" text.

flobarreto:
closes #1148


webapp/src/components/AssetPage/SaleRentActionBox/PeriodsDropdown/PeriodsDropdown.tsx
```diff
@@ -29,6 +39,8 @@ const Trigger = ({ period }: { period: RentalListingPeriod }) => {
 }
 
 const PeriodsDropdown = ({ value, periods, className, onChange }: Props) => {
+  const [isOpenDropdown, setIsOpenDropdwon] = useState(false);
```
juanmahidalgo, line 42:
```suggestion
  const [isOpenDropdown, setIsOpenDropdown] = useState(false);
```
small typo here

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -104,7 +104,7 @@ const SaleRentActionBox = ({
   const handleOnRent = useCallback(() => {
     if (!!authorization && hasAuthorization(authorizations, authorization)) {
```
juanmahidalgo, line 105:
```suggestion
    if (!!authorization && hasAuthorization(authorizations, authorization) && selectedRentalPeriodIndex !== undefined) {
```
what if we add this check here so we avoid the ternaries from below? 
like: `onRent(selectedRentalPeriodIndex ? selectedRentalPeriodIndex : 0)`, it would turn back to `onRent(selectedRentalPeriodIndex)`

PR #1456, opened 2023-03-07
feat: Enhancements for the creators filter

### LINKED ISSUE #1455: Enhancements for the creators filter
## 🎉 Description

Right now, the filter is live but it has a few things to improve:
- Add a tooltip explaning what it does and what the suggestions are
- Remove the search term on click
- Reset the search after clicking on one suggestion
- Do not show the menu on the loading state
- Performance: re-use the last result if the search term is the same (to avoid extra requests to the nft-server and catalysts)
- Add a 'Suggestions' menu header so it's clear what the default list is

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #1455

image

### DIFF
--- a/webapp/src/components/AssetFilters/CreatorsFilter/CreatorsFilter.tsx
+++ b/webapp/src/components/AssetFilters/CreatorsFilter/CreatorsFilter.tsx
@@ -8,7 +8,7 @@ import { Props } from './CreatorsFilter.types'
 import { profileToCreatorAccount } from './utils'
 import './CreatorsFilter.css'
 
-type Creator = {
+export type Creator = {
   name: string
   address: string
 }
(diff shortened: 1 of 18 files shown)


webapp/src/components/AssetFilters/CreatorsFilter/CreatorsFilter.tsx
```diff
@@ -109,67 +115,97 @@ export const CreatorsFilter = ({
...
-      handleCreatorsChange(data.value as string)
-      if (searchTerm && !data.value) {
-        setSearchTerm('')
-        if (dropdownRef.current) {
-          // typing as any since the type of the Legacy ref component is not exposed by the lib
-          ;(dropdownRef.current as any).clearSearchQuery?.()
-        }
+  // tracks the click outside the main div and close suggestions if needed
+  useEffect(() => {
+    const handleClickOutside = (event: any) => {
```
cyaiox, line 122: could the type `React.MouseEvent<HTMLButtonElement>` work here instead of any?

juanmahidalgo: Oh, I forgot about that `any`. That one that you mentioned doesn't work, typescripts complains it doesn't fit the `addEventListener` expected type. I used `MouseEvent` instead of `any`

cyaiox (dismissed):
🚀

cyaiox (approved):

PR #1465, opened 2023-03-16
fix: First rental period was not buyable because it always showed "not enough mana"

LautaroPetaccio:
This PR fixes an issue that was induced by the change of having the periods dropdown without a default value. The issue was caused because of a condition that was used to check if there was enough MANA didn't take into consideration that the index of the selected period could be 0, a false value, resulting in the first period being always disabled because of "not having enough mana".


fzavalia (approved):
Looks good 🚀

PR #1472, opened 2023-03-21
Feat: mint or buy component

### LINKED ISSUE #1436: [3.2] Create the new module to mint/buy cheapest listing
This module shows the best buying option: 

1. When the item is available for minting, it shows the buttons for minting. (data already available from the `/items` request)
2. When the item is not available for minting, it shows the cheapest listing. (this data will be available once the listings table finishes its request, the cheapest price will be the first one of the table) 
3. If the item is not for minting and there are no available listings, it encourages users to make offers to available if owners.

This component should be created taking in mind we need a similar one on nft page

flobarreto:
closes #1436 

**available for listing** (not available for minting)
image

**available for listing wit no bids** (not available for minting)
image

**available for minting**
image

**not available**
image

### DIFF
--- a/webapp/src/components/AssetPage/ListingsTable/ListingsTable.module.css
+++ b/webapp/src/components/AssetPage/ListingsTable/ListingsTable.module.css
@@ -30,11 +30,6 @@
   align-items: center;
 }
 
-.ListingsTable .badge {
-  color: white;
-  background-color: var(--background) !important;
-}
-
 .ListingsTable .goToNFT {
   color: white;
   padding-right: 25px;
@@ -45,6 +40,12 @@
   display: flex;
   align-items: center;
   justify-content: center;
+}
+
+.ListingsTable .manaField {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
 }
 
 .ListingsTable :global(.ui.header.dcl.mana) {
--- a/webapp/src/components/AssetPage/ListingsTable/ListingsTable.tsx
+++ b/webapp/src/components/AssetPage/ListingsTable/ListingsTable.tsx
@@ -1,9 +1,9 @@
 import React, { useState, useEffect } from 'react'
+import { Link } from 'react-router-dom'
 import { ListingStatus, Network } from '@dcl/schemas'
 import { Table, Loader, Row, Pagination, Icon, Mana } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Order, OrderFilters, OrderSortBy } from '@dcl/schemas/dist/dapps/order'
-import { Link } from 'react-router-dom'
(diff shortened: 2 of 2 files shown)


webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
```diff
@@ -0,0 +1,252 @@
...
+import styles from './BestBuyingOption.module.css'
+
+const BestBuyingOption = ({ asset }: Props) => {
+  const [buyOption, setBuyOption] = useState<BuyOptions | null>(null)
+  const [isLoading, setIsLoading] = useState(false)
+  const [listing, setListing] = useState<{
+    order: Order
+    total: number
+  } | null>(null)
+  const [bid, setBid] = useState<Bid | null>(null)
```
juanmahidalgo, line 34:
```suggestion
  const [bid, setBid] = useState<Bid>()
```

webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
```diff
@@ -0,0 +1,252 @@
...
+          .then(response => {
+            if (response.data.length > 0) {
+              setBuyOption(BuyOptions.BUY_LISTING)
+              setListing({ order: response.data[0], total: response.total })
+              bidAPI
+                .fetchByNFT(
+                  asset.contractAddress,
+                  response.data[0].tokenId,
+                  ListingStatus.OPEN,
+                  BidSortBy.MOST_EXPENSIVE,
```
juanmahidalgo, line 68:
```suggestion
                  BidSortBy.MOST_EXPENSIVE,
```
I believe this is going to return us the most expensive one, right? We want the cheapest. Can we add a `orderDirection`?

webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
juanmahidalgo: I'd also call this with a more explicit name. Maybe something like `cheapestBid` ?

webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
```diff
@@ -0,0 +1,252 @@
...
+          <div className={styles.containerRow}>
+            <BuyNFTButtons asset={asset} />
+          </div>
+          <Button
+            href={locations.nft(asset.contractAddress, listing.order.tokenId)}
+          >
+            {t('best_buying_option.buy_listing.view_listing')}
+          </Button>
+          <span>
+            {t('best_buying_option.buy_listing.expires')}
```
juanmahidalgo, line 221:
```suggestion
            {t('best_buying_option.buy_listing.expires')}{' '}
```
to fix this:
image

PR #1486, opened 2023-03-22
feat: add tracking for chart filters

meelrossi:
Add track event for all range filters. This will allow us to track how the user interacts with the component:
- If it uses the min/max inputs
- If it uses the slider
- If it uses the chart and select the ranges


webapp/src/components/AssetFilters/AssetFilters.tsx
```diff
@@ -62,22 +63,18 @@ export const AssetFilters = ({
...
+  const analytics = getAnalytics()
 
   const handleRangeFilterChange = useCallback(
-    (filterNames: [string, string], value: [string, string]) => {
+    (filterNames: [string, string], value: [string, string], source) => {
       const [filterMinName, filterMaxName] = filterNames
       const [minPrice, maxPrice] = value
+      analytics.track(
+        `${filterMinName} and ${filterMaxName} changed using ${source}`
+      )
```
LautaroPetaccio, line 74: Just out of curiosity, following the segment API I see the following:
```
analytics.track("User Registered", {
  plan: "Pro Annual",
  accountType: "Facebook"
});
```
Which means that the first parameter is the event name and the second one is the properties. Don't we need the name to be fixed and the properties variable?
```
analytics.track('Slider change', { filterMinName, filterMaxName, source })
```

LautaroPetaccio (approved):

PR #1498, opened 2023-03-27
feat: save scroll when going back

meelrossi:
Scroll to selected item when going from detail page to collectibles

https://user-images.githubusercontent.com/11800206/228032880-769f34b9-4896-418e-ad24-38fcd8ca9332.mov

### DIFF
--- a/webapp/src/components/AssetBrowse/AssetBrowse.tsx
+++ b/webapp/src/components/AssetBrowse/AssetBrowse.tsx
@@ -1,12 +1,7 @@
 import React, { ReactNode, useEffect, useState } from 'react'
+import { matchPath } from 'react-router-dom'
 import classNames from 'classnames'
-import {
-  Container,
-  Mobile,
-  NotMobile,
-  Page,
-  Tabs,
-} from 'decentraland-ui'
+import { Container, Mobile, NotMobile, Page, Tabs } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { View } from '../../modules/ui/types'
 import { Section as DecentralandSection } from '../../modules/vendor/decentraland'
@@ -33,6 +28,7 @@ import { Props } from './AssetBrowse.types'
 import MapTopbar from './MapTopbar'
 import MapBrowse from './MapBrowse'
 import './AssetBrowse.css'
+import { locations } from '../../modules/routing/locations'
 
 const AssetBrowse = (props: Props) => {
   const {
@@ -51,6 +47,7 @@ const AssetBrowse = (props: Props) => {
     onlySmart,
     viewInState,
     onlyOnRent,
+    visitedLocations,
     isMapViewFiltersEnabled
   } = props
 
@@ -107,12 +104,18 @@ const AssetBrowse = (props: Props) => {
         !isAccountView(view) &&
         isMapPropertyPersisted === false
       ) {
+        const previousPageIsLandDetail = !!matchPath(
+          visitedLocations[1]?.pathname,
+          { path: locations.nft(), strict: true, exact: true }
+        )
         // Update the browser options to match the ones persisted.
         browseOpts.isMap = isMap
         browseOpts.isFullscreen = isFullscreen
         browseOpts.onlyOnSale =
-          (!onlyOnSale && onlyOnRent === false) ||
-          (onlyOnSale === undefined && onlyOnRent === undefined) ||
+          (!onlyOnSale && onlyOnRent === false && !previousPageIsLandDetail) ||
+          (onlyOnSale === undefined &&
+            onlyOnRent === undefined &&
+            !previousPageIsLandDetail) ||
           onlyOnSale
 
         // We also set the fetch function as onBrowse because we need the url to be updated.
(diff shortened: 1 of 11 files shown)


webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -33,6 +28,7 @@ import { Props } from './AssetBrowse.types'
 import MapTopbar from './MapTopbar'
 import MapBrowse from './MapBrowse'
 import './AssetBrowse.css'
+import { locations } from '../../modules/routing/locations'
```
juanmahidalgo, line 31: can we move this import a few lines up please? 🙏

webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -107,12 +104,18 @@ const AssetBrowse = (props: Props) => {
...
+          visitedLocations[1]?.pathname,
+          { path: locations.nft(), strict: true, exact: true }
+        )
         // Update the browser options to match the ones persisted.
         browseOpts.isMap = isMap
         browseOpts.isFullscreen = isFullscreen
         browseOpts.onlyOnSale =
-          (!onlyOnSale && onlyOnRent === false) ||
-          (onlyOnSale === undefined && onlyOnRent === undefined) ||
+          (!onlyOnSale && onlyOnRent === false && !previousPageIsLandDetail) ||
```
juanmahidalgo, line 115: I'm missing some context here, what is this trying to solve? also, should we be more specific in the `previousPageIsLandDetail` definition? seems it's only matching the `nft` location, so it could be wearable/emote/name as well, right? (not just `Land`)

webapp/src/components/AssetBrowse/AssetBrowse.types.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from '../../modules/routing/actions'
 import { Section } from '../../modules/vendor/routing/types'
 import { AssetType } from '../../modules/asset/types'
+import { RouterLocation } from 'connected-react-router'
```
juanmahidalgo, line 14: can we move this import a few lines above please? 😅

webapp/src/components/AssetCard/AssetCard.tsx
```diff
@@ -106,6 +106,7 @@ const AssetCard = (props: Props) => {
       as={Link}
       to={getAssetUrl(asset, isManager && isLand(asset))}
       onClick={onClick}
+      id={`${asset.contractAddress}-${'tokenId' in asset ? asset.tokenId : asset.itemId}`}
```
juanmahidalgo, line 109:
```suggestion
      id={`${asset.contractAddress}-${asset.tokenId ?? asset.itemId}`}
```
would this work? looks a little bit simplier

webapp/src/components/AssetList/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+  return matchPath<T>(path, { path: route, strict: true, exact: true })
+}
+
+export function getLastVisitedElementId(
+  currentLocation: string,
+  lastVisitedLocation: string
+) {
+  const matchLands = matchAppRoute(currentLocation, locations.lands())
+  const matchCollectibles = matchAppRoute(currentLocation, locations.browse())
+  const previousMatchItems =  matchAppRoute<{ contractAddress: string; itemId: string }>(lastVisitedLocation, locations.item())
```
juanmahidalgo, line 14: can we move this line underneath the `if ((matchLands && previousMatchNfts) || (matchCollectibles && previousMatchNfts))` ? so we compute the value just if it's not coming from lands or collectibles nft view. WDYT?

webapp/src/modules/routing/reducer.ts
```diff
@@ -21,6 +28,15 @@ export function routingReducer(
         isLoadMore: action.payload.isLoadMore
       }
     }
+    case LOCATION_CHANGE: {
+      return {
+        ...state,
+        visitedLocations: [
+          action.payload.location,
+          ...state.visitedLocations,
+        ].slice(0, 5) // only save last 5 locations
```
juanmahidalgo, line 37: Is it worth it to store five latest locations? we always check against the previous one, maybe storing 2 is enough for the issue we're trying to solve. WDYT?

juanmahidalgo (approved):
Looks great, I really like the solution proposed. I left a couple of small suggestions and comments!

webapp/src/components/AssetBrowse/AssetBrowse.tsx
meelrossi: The problem im trying to solve is that if you choose `All land` it will set onlyOnSale and onlyOnRent to undefined (not appearing in the url). If later you go to the detail page of some land (that is not on sale) and come back, this condition here is setting `onlyOnSale` to be true automatically causing all lands that are not on sale not to be visible (so there is no possibility to "scrollToView")

The problem is that im not sure how to detect that is a land from the location. as the url doesn't have anything (not even a query param) that is different from a wearable/emote/etc nft.

juanmahidalgo (approved):
LGTM!

PR #1503, opened 2023-03-28
feat: Allowance not enough

fzavalia:


webapp/src/modules/contract/sagas.ts
```diff
@@ -270,3 +279,12 @@ function* handleChangeAccount() {
 function* handleConnectWalletSuccess() {
   yield put(resetHasFetched())
 }
+
+function* handleFetchTransactionSuccess(action: FetchTransactionSuccessAction) {
+  const { transaction } = action.payload
+
+  if (transaction.actionType === GRANT_TOKEN_SUCCESS) {
+    const authorization: Authorization = transaction.payload.authorization
+    yield put(fetchAuthorizationRequest(authorization))
```
juanmahidalgo, line 288: Everything looks great, the only question that I have is why this saga was needed. Couldn't we use the already existing one `fetchAuthorizationsRequest` and use an array of 1 element? 
So this would be `yield put(fetchAuthorizationsRequest([authorization]))`. I've been checking the dcl-dapps code for this saga and couldn't find the difference with the other method

juanmahidalgo (approved):
LGTM!! I left a question that could be an improvement in the future, but for now works great 👍

webapp/src/modules/contract/sagas.ts
cyaiox: Updated logic in `decentraland-dapps` to properly use `fetchAuthorizationsRequest` without duplicating authorizations in the state.

cyaiox (approved):

PR #1506, opened 2023-03-29
removed the react-copy-to-clipboard lib

Chakravarthy7102:
Completely removed the dependency @react-copy-to-clipboard in the project and replaced it with a custom hook to copy the text to clipboard.


webapp/src/hooks/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,9 @@
+const useCopyToClipboard = (setHasCopied: () => void) => {
+  const copy = (text: string) => {
+    navigator.clipboard.writeText(text)
+    setHasCopied()
+  }
+  return copy
+}
```
LautaroPetaccio, line 7: What if we build this as a un utility function instead of a hook? The main reason is that it doesn't need to store anything or use any other hook to operate.
```suggestion
const copyToClipboard = (onCopy: () => void) => {
    navigator.clipboard.writeText(text)
    onCopy()
}
```

LautaroPetaccio (changes requested):
Hi! Thanks for your contribution to the project. The PR looks great. I've left some comments to address before merging it.

webapp/src/hooks/useCopyToClipboard.tsx
Chakravarthy7102: Yup!, your argument make sense a lot, there is no need for a hook for this. I'll change it and also do the other changes you suggested.

Chakravarthy7102: But here in this copyToClipboard function, we also need to add the text as parameter to access the text inside the func
`const copyToClipboard = (text:string, onCopy:()=>void)`, if there is possibility please edit your suggestion or else I'll make a new pull request!

webapp/src/lib/copyText.ts
```diff
@@ -0,0 +1,6 @@
+const copyText = (text: string, onCopy: () => void) => {
```
LautaroPetaccio, line 1: Would you mind adding a test for this function?
I've [here](https://www.notion.so/decentraland/Testing-standards-46797744fccf4f3eba52335f9866d0eb) the testing standards we're using for you to take a look.
It would be nice to have a test that assert that when calling this function, the clipboard gets written and that `onCopy` gets called.
It should look something like:
```
describe('when calling the copy text function', () => {
  let writeTextSpy: jest.SpyInstance
  let onCopyMock: jest.Mock
  beforeEach(() => {
    onCopyMock = jest.fn()
    writeTextSpy = jest.spyOn(navigator.clipboard, 'writeText')
  })

  it('should write the text to the clipboard and execute the onCopy callback', () => {
    copyText('someText', onCopyMock)
    expect(onCopyMock).toHaveBeenCalled()
    expect(writeTextSpy).toHaveBeenCalledWith('someText)
  })
})
```

LautaroPetaccio (changes requested):
Just one more little thing and we're ready to go.

webapp/src/lib/copyText.ts
Chakravarthy7102: Yup sure will do that!

Chakravarthy7102:
tests added! @LautaroPetaccio

LautaroPetaccio (approved):
Looks great! If you want, you can also replicate this change in the [Builder](https://github.com/decentraland/builder), we'll be glad to have this change there as well.

LautaroPetaccio:
I'm unable to merge this PR because we require all contributors to sign all commits with their own GPG key as a security measure.

Would you mind:
1. [Creating and adding a GPG key to your Github account](https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key)
2. [Signing all of your commits in this branch](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits)

For the commits to work they must say "Verified" next to them in the Github interface.

Chakravarthy7102:
Signed  all the commits, Thanks for linking relevant resources!, I think now we are good to goww!

LautaroPetaccio (approved):

LautaroPetaccio:
I'm sorry @Chakravarthy7102, but there are still a few commits that are not signed, all of them must be signed for Github to allow me to merge the PR.
These are the commits in this PR that are not signed:
- 46f81d15904fca99e85bacfc64820a1b669d5040
- c468d34a950c5a05c2cde377da17d8be3c9c03d7
- 451390681ad7ade9e8a845c765b218aa7eecc936
- b9e2569662558ee752247a4dd09fd77f06e2445c
You might need to do a force push after signing them becase the history will change.

Chakravarthy7102:
@LautaroPetaccio Hi, I'm not tried different things but couldn't able to verify my all the commits. Instead i guess made lots of bad things into commit history!

One thing that comes into my mind right now to fix this situation is to delete my fork and create a new pull request for this issue.

or any other way that you can suggest would also help!

LautaroPetaccio:
Hi @Chakravarthy7102, it seems that the history broke. You can delete your branch, close this PR and create a new one with all the commits signed.

PR #1513, opened 2023-03-31
feat: Add use partially supported network alert

LautaroPetaccio:
This PR adds the partially supported network alert alongside a custom message for the BuyPage.

Screenshot 2023-03-31 at 16 53 54
Screenshot 2023-03-31 at 16 53 33
Screenshot 2023-03-31 at 16 53 07


webapp/src/components/BuyPage/BuyPage.tsx
```diff
@@ -27,7 +27,7 @@ const BuyPage = (props: Props) => {
 
   return (
     <>
-      <Navbar isFullscreen />
+      <Navbar isFullscreen enablePartialSupportAlert={false} />
```
cyaiox, line 30: should we use a FF here?

cyaiox (approved):
LGTM

webapp/src/components/BuyPage/BuyPage.tsx
LautaroPetaccio: This flag is to prevent the alert to being shown. It needs to be prevent in this specific page only.

cyaiox (approved):

PR #1521, opened 2023-04-04
feat: redirect after log in

meelrossi:
Add `redirectTo` query param in signin page so when sign in finishes it redirects to the desired page instead of always redirecting to account view

### DIFF
--- a/webapp/src/components/Bids/Bids.tsx
+++ b/webapp/src/components/Bids/Bids.tsx
@@ -6,6 +6,7 @@ import { locations } from '../../modules/routing/locations'
 import { Bid } from '../Bid'
 import { Props } from './Bids.types'
 import './Bids.css'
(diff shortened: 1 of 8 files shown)


webapp/src/components/Bids/Bids.tsx
```diff
@@ -6,6 +6,7 @@ import { locations } from '../../modules/routing/locations'
 import { Bid } from '../Bid'
 import { Props } from './Bids.types'
 import './Bids.css'
+import { useLocation } from 'react-router-dom'
```
LautaroPetaccio, line 9: Would you mind moving this import up?

LautaroPetaccio (approved):
Looks great Meli! Love this change 🚀

LautaroPetaccio (approved):

PR #1538, opened 2023-04-11
fix: Fix styles on loading an empty states

kevinszuchet:
Fixed the loader vertically, and centered it horizontally:
image


juanmahidalgo (approved):
LGTM!

PR #1539, opened 2023-04-11
feat: Add authentication

LautaroPetaccio:
This PR does the following:
- Adds authentication to the Marketplace upon logging in.
- Modifies the item's `api.ts` to inherit from the new `BaseClient` class which uses `signedFetch` to send signed requests when the identity is available.
- Fixes some tests that were making a real request to the nft-server.

### DIFF
--- a/webapp/src/modules/identity/sagas.ts
+++ b/webapp/src/modules/identity/sagas.ts
@@ -1,7 +1,8 @@
-import { takeLatest, put, call } from 'redux-saga/effects'
+import { takeLatest, put, call, select } from 'redux-saga/effects'
 import { ethers } from 'ethers'
 import { Authenticator, AuthIdentity } from '@dcl/crypto'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { CONNECT_WALLET_SUCCESS } from 'decentraland-dapps/dist/modules/wallet/actions'
 import { isErrorWithMessage } from '../../lib/error'
 import { getEth } from '../wallet/utils'
 
@@ -9,12 +10,16 @@ import {
   GENERATE_IDENTITY_REQUEST,
   GenerateIdentityRequestAction,
   generateIdentityFailure,
+  generateIdentityRequest,
   generateIdentitySuccess
 } from './actions'
 import { IDENTITY_EXPIRATION_IN_MINUTES } from './utils'
+import { ConnectWalletSuccessAction } from 'decentraland-dapps/dist/modules/wallet/actions'
+import { getCurrentIdentity } from './selectors'
 
 export function* identitySaga() {
   yield takeLatest(GENERATE_IDENTITY_REQUEST, handleGenerateIdentityRequest)
+  yield takeLatest(CONNECT_WALLET_SUCCESS, handleConnectWalletSuccess)
 }
 
 function* handleGenerateIdentityRequest(action: GenerateIdentityRequestAction) {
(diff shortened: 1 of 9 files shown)


webapp/src/modules/identity/sagas.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        .dispatch(connectWalletSuccess(wallet))
+        .run({ silenceTimeout: true })
+    })
+  })
+
+  describe("and there's an identity", () => {
+    it('should not put an action to generate the identity', () => {
+      return expectSaga(identitySaga)
+        .provide([[select(getCurrentIdentity), {}]])
+        .not.put(generateIdentityRequest(wallet.address))
```
kevinszuchet, line 35: I didn't know this one. Nice!

webapp/src/modules/store.ts
```diff
@@ -71,9 +73,17 @@ export function initStore() {
...
   const enhancer = composeEnhancers(middleware)
-  const store = createStore(rootReducer, enhancer)
-
-  sagasMiddleware.run(rootSaga)
+  const store = createStore(
+    (rootReducer as unknown) as ReturnType<typeof createRootReducer>,
+    enhancer
+  )
+  const getIdentity = () => {
+    store.getState()
```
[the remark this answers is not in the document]
LautaroPetaccio: It wasn't :( removed it!

webapp/src/modules/identity/sagas.ts
```diff
@@ -56,7 +56,9 @@ function* handleGenerateIdentityRequest(action: GenerateIdentityRequestAction) {
 }
 
 function* handleConnectWalletSuccess(action: ConnectWalletSuccessAction) {
+  console.log('Handling wallet connect success')
   const identity: AuthIdentity = yield select(getCurrentIdentity)
+  console.log('Identity exists', identity)
   if (!identity) {
```
kevinszuchet, line 62: Do we need this two `console.log` statements?

webapp/src/modules/vendor/decentraland/item/api.ts
```diff
@@ -15,13 +15,13 @@ export class ItemAPI extends BaseClient {
...
-      body: JSON.stringify({
-        contractAddress,
-        itemId
-      })
+    const queryParams = this.buildItemsQueryString({
+      contracts: [contractAddress],
+      itemId
     })
+    const response: ItemResponse = await this.fetch(
+      `/v1/items?${queryParams.toString()}`
```
kevinszuchet, line 23: This was a POST with the query parameters in the request body, and now it is a GET that uses those query parameters in the URL. Is this going to work?

webapp/src/modules/wallet/utils.ts
```diff
@@ -40,3 +46,13 @@ export function formatBalance(balance: number) {
     ? removeScientificNotationForSmallNumbers(balance)
     : balance.toString()
 }
+
+export function* waitForWalletConnectionIfConnecting() {
+  const isConnectingToWallet: boolean = yield select(isConnecting)
+  if (isConnectingToWallet) {
+    yield race({
+      success: take(CONNECT_WALLET_SUCCESS),
+      failure: take(CONNECT_WALLET_FAILURE)
```
kevinszuchet, line 55: Aren't we interested in the race winner?

LautaroPetaccio: We aren't we just want the wallet connection process to end as, in this case, items can be requested by being logged in or not. We must work with those two cases because the Marketplace does not requires you to log in.

webapp/src/modules/identity/sagas.ts
LautaroPetaccio: Did you make this comment over a commit? I'm asking that because [this commit](https://github.com/decentraland/marketplace/pull/1539/commits/ecbadee0bd4bf7cd2fcb9b8d4568c407b7eeb00b) has the comments removed.

webapp/src/modules/vendor/decentraland/item/api.ts
LautaroPetaccio: I think that you made this comment out of an old commit which had a `POST` request that was mistakenly added. The original request is a GET request with query parameters.

webapp/src/modules/identity/sagas.ts
kevinszuchet: Weird, my bad!

PR #1549, opened 2023-04-12
fix: Modal styles and data loading

LautaroPetaccio:
This PR does the following:
- Adds a bar to split the modal text from the list.
- Changes the styles of the Profile component so the user's address / text is bigger and is centered.
- Fixes how picks were fetched.

Screenshot 2023-04-12 at 16 50 37
Screenshot 2023-04-12 at 16 50 24

### DIFF
--- a/webapp/src/components/Modals/FavoritesModal/FavoritesModal.tsx
+++ b/webapp/src/components/Modals/FavoritesModal/FavoritesModal.tsx
@@ -68,6 +68,7 @@ const FavoritesModal = ({ metadata: { itemId }, onClose }: Props) => {
       <div style={style} tabIndex={0}>
         {isItemLoaded(index) ? (
           <LinkedProfile
+            className={styles.user}
             size="huge"
             key={favorites.addresses[index]}
             sliceAddressBy={isMobile ? 18 : 40}
@@ -83,6 +84,7 @@ const FavoritesModal = ({ metadata: { itemId }, onClose }: Props) => {
 
   useEffect(() => {
     fetchNextPage(0, 100)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
   // Makes the modal dynamic in size.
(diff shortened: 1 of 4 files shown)


webapp/src/modules/vendor/decentraland/favorites/api.ts
```diff
@@ -12,6 +12,25 @@ export const MARKETPLACE_FAVORITES_SERVER_URL = config.get(
   'MARKETPLACE_FAVORITES_SERVER_URL'
 )!
 
+type PaginatedResponse<T> = {
+  results: T[]
+  total: number
+  page: number
+  pages: number
+  limit: number
+}
```
kevinszuchet, line 21: Definitely, we need this type in a common place (@dcl/schemas).

webapp/src/components/Modals/FavoritesModal/FavoritesModal.tsx
```diff
@@ -68,6 +68,7 @@ const FavoritesModal = ({ metadata: { itemId }, onClose }: Props) => {
       <div style={style} tabIndex={0}>
         {isItemLoaded(index) ? (
           <LinkedProfile
+            className={styles.user}
```
kevinszuchet, line 71: Where does this style come from? I can't see it in the `FavoritesModal.module.css`.

webapp/src/modules/vendor/decentraland/favorites/api.ts
LautaroPetaccio: We definitely do

webapp/src/components/Modals/FavoritesModal/FavoritesModal.tsx
LautaroPetaccio: I completely forgot to remove it, great catch!

kevinszuchet (approved):

PR #1568, opened 2023-04-14
fix: Fix Lists related toast styles

### LINKED ISSUE #1562: Review the separation between words in the toasts showing pick/unpick success and failure

### LINKED ISSUE #1563: When many toasts related to pick/unpick appear on top of each other, some have different sizes
## 🎉 Description

* [x] 🐛 This is a bug report.

The sizes of the toasts should be the same.

### LINKED ISSUE #1565: When saving more than 5 items it starts to load slow and gets stuck

### LINKED ISSUE #1574: Picking / unpicking an item should show only one toast
When picking / unpicking an item, only one toast should be shown, overwriting any other toasts that appear in the screen.

kevinszuchet:
Closes #1562 
Closes #1563 
Closes #1565
Closes #1574

### DIFF
--- a/webapp/src/index.css
+++ b/webapp/src/index.css
@@ -57,8 +57,10 @@ code {
   padding: 0;
 }
 
-.dcl.toast .toast-info .body .list-flow-toast {
-  width: 350px;
+@media only screen and (min-width: 1200px) {
+  .dcl.toast .toast-info .body .list-flow-toast {
+    width: 350px;
+  }
 }
 
 @media (max-width: 767px) {
--- a/webapp/src/index.css
+++ b/webapp/src/index.css
@@ -57,6 +57,10 @@ code {
   padding: 0;
 }
 
+.dcl.toast .toast-info .body .list-flow-toast {
+  width: 350px;
+}
+
 @media (max-width: 767px) {
   .Navigation,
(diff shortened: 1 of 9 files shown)


webapp/src/modules/translation/locales/en.json
```diff
@@ -897,23 +897,23 @@
       "refresh": "Refresh"
     },
     "pick_item_as_favorite_success": {
-      "title": "Item added to Favorites",
-      "body": "{name} was successfully added to Favorites",
+      "title": "Item added to my list",
```
LautaroPetaccio, line 900: The phrase that was suggested by the UX team was `Wishlist` which is the name of the list, would you mind changing it?
```suggestion
      "title": "Item added to Wishlist",
```

webapp/src/index.css
```diff
@@ -57,6 +57,10 @@ code {
   padding: 0;
 }
 
+.dcl.toast .toast-info .body .list-flow-toast {
+  width: 350px;
+}
```
LautaroPetaccio, line 62: Can we use this width only for the desktop version of the toast? The mobile one should stretch through all the width of the screen. If that's not set at the moment, we can set it in the UI when we modify the toast.

LautaroPetaccio (approved):

PR #1581, opened 2023-04-14
fix: Loading more than 20 items in my lists broke pagination

### LINKED ISSUE #1580: Loading more than 20 items in My Lists breaks pagination
There's an issue where if the user loads more than 20 items at the time, it breaks pagination because the nft-server is still waiting for the `first` filter to be applied when sending the items requests.

LautaroPetaccio:
This PR fixes an issue where loading more than 20 items broke the My List pagination due to the nft-server waiting for the `first` parameter to come along with the queried ids.

Closes #1580

### DIFF
--- a/webapp/src/modules/favorites/sagas.ts
+++ b/webapp/src/modules/favorites/sagas.ts
@@ -147,6 +147,7 @@ function* handleFetchFavoritedItemsRequest(
     const options: ItemBrowseOptions = {
       ...action.payload,
       filters: {
+        first: results.length,
         ids: results.map(({ itemId }) => itemId)
       }
     }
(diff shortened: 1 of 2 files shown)


webapp/src/modules/favorites/sagas.ts
```diff
@@ -147,6 +147,7 @@ function* handleFetchFavoritedItemsRequest(
     const options: ItemBrowseOptions = {
       ...action.payload,
       filters: {
+        first: results.length,
```
kevinszuchet, line 150: Why the length of the results instead of using `...action.payload.filters`. I think I forgot to use it there.

kevinszuchet (approved):

PR #1591, opened 2023-04-18
fix: Loading empty list of favorited items

### LINKED ISSUE #1590: Can't load the My List tab if the user doesn't have any favorites
If the user doesn't have any favorites, the tab breaks due to it trying to perform a items fetch without item ids.
The URL requested ends up being: `https://nft-api.decentraland.zone/v1/items?` which fails with a 500.

LautaroPetaccio:
This PR fixes an error that prevented users with no favorited items from loading the tab correctly.
The fix puts either the `fetchItemsRequest` or the `fetchItemsSuccess` depending on if the favorites API response resulted in more than one id or not.

Closes #1590


kevinszuchet (approved):
Nice catch!

meelrossi (approved):

PR #1600, opened 2023-04-20
fix: Only the avatar (or blockie) and address are clickable

### LINKED ISSUE #1561: The clickable area of the NAMEs in the modal that shows the users that saved an item is too big
@kevinszuchet commented on [Fri Apr 14 2023](https://github.com/decentraland/account/issues/265)

## 🎉 Description

* [x] 🐛 This is a bug report.

It is not so clear where one name link ends and the following starts, just the name should be clickable, not the space between names

kevinszuchet:
Closes #1561


LautaroPetaccio (approved):
Send it 🚀

PR #1605, opened 2023-04-20
fix: Wait until finish loading to render assets

### LINKED ISSUE #1592: When switching from My Assets to My List the items can display the on sell icon
## 🎉 Description

When switching a bit fast from My Assets to My List, the view gets stuck at the one that is used in My Assets. This might be due to a race condition on the `setView` action that ends up overwriting the views when switching between one another.

Steps to reproduce:
1. Have several items favorited and some owned ones in the wallet.
2. Go to My Assets and kind of fast, click on the My List tab
3. The items in My List tab will appear with the on sale icon.

kevinszuchet:
Closes #1592

### DIFF
--- a/webapp/src/components/AssetList/AssetList.tsx
+++ b/webapp/src/components/AssetList/AssetList.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useEffect } from 'react'
+import React, { useCallback, useState, useMemo, useEffect } from 'react'
 import { Card, Loader } from 'decentraland-ui'
 import { Item, NFTCategory } from '@dcl/schemas'
 import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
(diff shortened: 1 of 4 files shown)


meelrossi (approved):

webapp/src/components/AssetList/AssetList.tsx
```diff
@@ -91,7 +91,7 @@ const AssetList = (props: Props) => {
         </>
       ) : null}
       <Card.Group>
-        {assets.length > 0
+        {assets.length > 0 && (!isLoading || (isLoading && isLoadingMore))
```
kevinszuchet, line 94: I think in the second statement: `isLoading && isLoadingMore` the `isLoading` is redundant.

`(!isLoading && isLoading) || (!isLoading && isLoadingMore)` could be written as: `!isLoading || isLoadingMore`

kevinszuchet (commented):
Nice approach!

I left a minor comment to refactor a comparison statement.

webapp/src/components/AssetList/AssetList.tsx
LautaroPetaccio: You're completely right! Removed it!

meelrossi (approved):

meelrossi (approved):

PR #1606, opened 2023-04-20
feat: Remove favorited items & start using skip instead of pages

LautaroPetaccio:
This PR introduces the following:
- The capability to remove favorited items and make them disappear from the My Lists UI.
- The usage of skip instead of pages, which allows the removal of elements to not break the pagination.
- Simplifies the `AssetList` component so it receives assets instead of nfts or items.

### DIFF
--- a/webapp/src/components/InfiniteScroll/InfiniteScroll.tsx
+++ b/webapp/src/components/InfiniteScroll/InfiniteScroll.tsx
@@ -2,18 +2,19 @@ import { useCallback, useEffect, useState } from 'react'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Button } from 'decentraland-ui'
 import { Props } from './InfiniteScroll.types'
+import { PAGE_SIZE } from '../../modules/vendor/api'
 
 export function InfiniteScroll({
-  page,
+  skip,
   hasMorePages,
   isLoading,
   children,
-  maxScrollPages,
+  maxScrollAssets,
   onLoadMore
 }: Props) {
   const [scrollPage, setScrollPage] = useState(0)
   const [showLoadMoreButton, setShowLoadMoreButton] = useState(
-    maxScrollPages === 0
+    maxScrollAssets === 0
   )
 
   const onScroll = useCallback(() => {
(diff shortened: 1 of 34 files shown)


webapp/src/components/InfiniteScroll/InfiniteScroll.tsx
```diff
@@ -24,30 +20,27 @@ export function InfiniteScroll({
...
-    setScrollPage(0)
-  }, [onLoadMore, page])
+    onLoadMore()
+    setScrolledTimes(0)
+  }, [onLoadMore])
+
+  const showLoadMoreButton =
+    !isLoading &&
+    maxInfiniteScrolls !== undefined &&
+    scrolledTimes === maxInfiniteScrolls
```
kevinszuchet, line 43: `scrolledTimes` cannot be greater than `maxInfiniteScrolls` right?

webapp/src/modules/favorites/sagas.spec.ts
```diff
@@ -29,11 +26,7 @@ import {
 } from './actions'
 import { favoritesSaga } from './sagas'
 import { getListId } from './selectors'
-import {
-  FETCH_ITEMS_SUCCESS,
-  fetchItemsRequest,
-  fetchItemsSuccess
-} from '../item/actions'
+import { fetchItemsRequest, fetchItemsSuccess } from '../item/actions'
```
kevinszuchet, line 29: Would you mind moving this up 🙏 ?

kevinszuchet (commented):
Amazing job 🚀 

I left 2 small comments

webapp/src/components/InfiniteScroll/InfiniteScroll.tsx
LautaroPetaccio: That's right! Whenever it reaches scrolled times, the `load more button` appears and upon clicking it, it resets the counter.

webapp/src/modules/routing/sagas.spec.ts
```diff
@@ -245,7 +243,7 @@ describe('when handling the browse action', () => {
     browseOptions = {
       onlyOnSale: undefined,
       sortBy: undefined,
-      page: 1,
+      skip: 1,
```
meelrossi, line 246: shouldn't this be 0? or 1 is ok?

meelrossi (dismissed):

cyaiox (approved):
🚀

webapp/src/modules/ui/browse/selectors.ts
```diff
@@ -23,8 +23,10 @@ import { ItemState } from '../../item/reducer'
 import { VendorName } from '../../vendor'
 import { getAddress, getWallet } from '../../wallet/selectors'
 import { getTransactionsByType } from '../../transaction/selectors'
+import { getFavoritedItems as getFavoritedItemsFromState } from '../../favorites/selectors'
 import { View } from '../types'
 import { OnRentNFT, OnSaleElement, OnSaleNFT } from './types'
+import { FavoritesData } from '../../favorites/types'
```
juanmahidalgo, line 29: can we move this import a few lines above? 🙏

juanmahidalgo (commented):
More of a general question, if we're using `skip` now, should we also use `first`? 
In the past, `page` 2 (having a page size of 10 items) would be the combination of `first: 10`, `skip: 10`. Am I missing something or the `skip` is enough to replicate the old behavior?

LautaroPetaccio:
Closed due to #1614

PR #1609, opened 2023-04-21
feat: authorization modal tests

meelrossi:
Add tests for authorization modal component


webapp/src/components/HOC/withAuthorizedAction/withAuthorizedAction.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+    it('should show authorization modal', async () => {
+      (hasAuthorization as jest.Mock).mockReturnValue(false);
+      (hasAuthorizationAndEnoughAllowance as jest.Mock).mockReturnValue(false);
+      const screen = renderComponentWithAuthorizedAction()
+      await userEvent.click(screen.getByRole('button', { name: 'Action' }))
+      expect(screen.getByTestId('authorization-modal')).toBeInTheDocument()
+    })
+  })
+
+  describe('and user has enought allowance', () => {
```
LautaroPetaccio, line 62:
```suggestion
  describe('and user has enough allowance', () => {
```

webapp/src/components/HOC/withAuthorizedAction/withAuthorizedAction.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+    it('should call action callback', async () => {
+      (hasAuthorization as jest.Mock).mockReturnValue(true);
+      (hasAuthorizationAndEnoughAllowance as jest.Mock).mockReturnValue(true);
+      const actionCallbackMock = jest.fn()
```
LautaroPetaccio, line 66: Would you mind moving the set up of the tests context into a `beforeEach`?

webapp/src/components/HOC/withAuthorizedAction/withAuthorizedAction.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+it('should render wrapped component', () => {
+  const screen = renderComponentWithAuthorizedAction()
+  expect(screen.getByRole('button', { name: 'Action' })).toBeInTheDocument()
+})
+
+describe('when onAuthorizedAction is called', () => {
+  describe("and user doesn't have enough allowance", () => {
+    it('should show authorization modal', async () => {
+      (hasAuthorization as jest.Mock).mockReturnValue(false);
+      (hasAuthorizationAndEnoughAllowance as jest.Mock).mockReturnValue(false);
```
LautaroPetaccio, line 55: Do both functions return the same when the user doesn't have enough allowance or should we test them separately?

webapp/src/components/HOC/withAuthorizedAction/withAuthorizedAction.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+  )
+  const WithAuthorizedActionComponent = withAuthorizedAction(
+    Component,
+    AuthorizationAction.BID
+  )
+  return renderWithProviders(<WithAuthorizedActionComponent />)
+}
+
+it('should render wrapped component', () => {
```
LautaroPetaccio, line 46: Should we add a test to exercise that closing the modal closes it?

webapp/src/components/HOC/withAuthorizedAction/AuthorizationModal/AuthorizationModal.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+    await userEvent.click(
+      screen.getByRole('button', { name: t('global.close') })
+    )
+    expect(onCloseMock).toHaveBeenCalled()
+  })
+})
+
+describe('when shouldAuthorized is true', () => {
+  it('should render two steps', () => {
+    const screen = renderAuthorizationModal({ shouldAuthorize: true })
```
LautaroPetaccio, line 39: What if we move the instantiation of the `screen` variable to a `beforeEach` to avoid repeating it?

webapp/src/components/HOC/withAuthorizedAction/AuthorizationModal/AuthorizationModal.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+  it('should call onClose action', async () => {
+    const onCloseMock = jest.fn()
+    const screen = renderAuthorizationModal({ onClose: onCloseMock })
+    await userEvent.click(
+      screen.getByRole('button', { name: t('global.close') })
+    )
+    expect(onCloseMock).toHaveBeenCalled()
+  })
+})
+
```
LautaroPetaccio, line 36: Should we add a test to be sure that the steps get changed when clicking the correct button?

LautaroPetaccio (commented):
Looks great! I've left some comments to review.

webapp/src/components/HOC/withAuthorizedAction/AuthorizationModal/AuthorizationModal.spec.tsx
meelrossi: The logic to change the steps will be different so i would have eto eerase that test in the next pr. I will add a test to make sure that if the currentStep is 1 then steo 0 shouldn't show the action

webapp/src/components/HOC/withAuthorizedAction/withAuthorizedAction.spec.tsx
meelrossi: if `hasAuthorizationAndEnoughAllowance` returns true then `hasAuthorization` should be true but not the other way around. I added an additional case when `hasAuthorization` is true and `hasAuthorizationAndEnoughAllowance` is false

LautaroPetaccio (approved):
Looks great ✅

PR #1610, opened 2023-04-21
feat: Show loader when picking & unpicking

### LINKED ISSUE #1588: Add a loading state to the bookmark/pick icon
In order to show the user that the picking of an item is in progress, change the bookmark icon for a loading icon.

kevinszuchet:
Closes #1588 

[Preview](https://www.loom.com/share/960abbfdfe23439e86796d88bf58e5ff)

Mobile:
image
image

Desktop:
image
image

### DIFF
--- a/webapp/src/components/FavoritesCounter/FavoritesCounter.container.ts
+++ b/webapp/src/components/FavoritesCounter/FavoritesCounter.container.ts
@@ -2,16 +2,13 @@ import { connect } from 'react-redux'
 import { Item } from '@dcl/schemas'
 import { openModal } from 'decentraland-dapps/dist/modules/modal/actions'
 import {
-  PICK_ITEM_AS_FAVORITE_REQUEST,
-  UNDO_UNPICKING_ITEM_AS_FAVORITE_REQUEST,
-  UNPICK_ITEM_AS_FAVORITE_REQUEST,
   pickItemAsFavoriteRequest,
   unpickItemAsFavoriteRequest
 } from '../../modules/favorites/actions'
 import {
   getIsPickedByUser,
   getCount,
-  getLoading
+  isPickingOrUnpicking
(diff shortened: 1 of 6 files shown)


webapp/src/components/FavoritesCounter/FavoritesCounter.container.ts
```diff
@@ -14,14 +21,19 @@ import {
...
     item: { id: itemId }
   } = ownProps
   return {
     isPickedByUser: getIsPickedByUser(state, itemId),
-    count: getCount(state, itemId)
+    count: getCount(state, itemId),
+    isLoading:
+      isLoadingType(getLoading(state), PICK_ITEM_AS_FAVORITE_REQUEST) ||
+      isLoadingType(getLoading(state), UNPICK_ITEM_AS_FAVORITE_REQUEST) ||
+      isLoadingType(getLoading(state), UNDO_UNPICKING_ITEM_AS_FAVORITE_REQUEST)
```
LautaroPetaccio: If we leave this loading state as is, it will set all favorites counters as being loaded when one of them is being loaded. What if we create a selector to get from the waited action which item id is being loaded?

webapp/src/components/FavoritesCounter/FavoritesCounter.container.ts
```diff
@@ -14,14 +21,19 @@ import {
   MapStateProps,
   OwnProps
 } from './FavoritesCounter.types'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
```
LautaroPetaccio: Would you mind re-ordering the import?

LautaroPetaccio (commented):
Looks great, just a comment on what we could do to show only one loader.

LautaroPetaccio (approved):

PR #1616, opened 2023-04-26
fix: Use page name instead of view

### LINKED ISSUE #1592: When switching from My Assets to My List the items can display the on sell icon

LautaroPetaccio:
This PR introduces the `getPageName` selector which replaces the `getView` selector in cases where we want to know what the page the user is in.

Closes #1592

### DIFF
--- a/webapp/src/modules/routing/selectors.ts
+++ b/webapp/src/modules/routing/selectors.ts
@@ -1,4 +1,5 @@
 import { createSelector } from 'reselect'
+import { matchPath } from 'react-router'
 import {
   getSearch as getRouterSearch,
   getLocation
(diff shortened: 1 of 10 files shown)


webapp/src/modules/routing/selectors.ts
```diff
@@ -523,3 +524,77 @@ export const getIsBuyWithCardPage = createSelector<RootState, string, boolean>(
     return withCard !== null && withCard === 'true'
   }
 )
+
+const buildExactMatchProps = (path: string) => {
+  return { path, exact: true }
+}
+
+export const getPageName = createSelector<RootState, string, PageName>(
```
kevinszuchet, line 532: WDTY about having a map with the locations and the page name, maybe in another file? I think the selector will be clearer.

kevinszuchet (commented):
Great solution!

I left a comment to discuss if we can write the new selector clearer.

kevinszuchet (approved):

PR #1633, opened 2023-04-28
fix: When switching tabs counter show different numbers

### LINKED ISSUE #1624: When switching from a tab to My Lists, the counter shows different numbers in the displayed number of items

kevinszuchet:
Closes #1624

### DIFF
--- a/webapp/src/components/AssetTopbar/AssetTopbar.tsx
+++ b/webapp/src/components/AssetTopbar/AssetTopbar.tsx
@@ -168,6 +168,7 @@ export const AssetTopbar = ({
       {!isMap && (
         <div className={styles.infoRow}>
           <div className={styles.countContainer}>
+            {/* TODO: it should wait for the 2 requests to finish (wishlist items & items) */}
             <p className={styles.countText}>{getCountText(count, search)}</p>
             {hasFiltersEnabled && !isMobile && (
               <button className={styles.clearFilters} onClick={onClearFilters}>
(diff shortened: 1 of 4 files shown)


webapp/src/modules/ui/browse/reducer.ts
```diff
@@ -185,7 +194,8 @@ export function browseReducer(
             ...state,
             view,
             itemIds: action.payload.items.map(item => item.id),
-            count: action.payload.total,
+            count:
+              section === Section.LISTS ? state.count : action.payload.total,
```
LautaroPetaccio, line 198: Would you mind explaining this change to me? If we're seeing the lists, we shouldn't be in the market, current account or account views. Is it because of a race condition?

webapp/src/modules/ui/browse/reducer.ts
```diff
@@ -158,6 +161,12 @@ export function browseReducer(
       }
     }
 
+    case FETCH_FAVORITED_ITEMS_REQUEST:
+      return {
+        ...state,
+        count: undefined
+      }
```
LautaroPetaccio, line 168: We'll be fetching favorited items each time we ask for more of them, should we clear the number?

webapp/src/components/AssetTopbar/AssetTopbar.tsx
```diff
@@ -168,6 +168,7 @@ export const AssetTopbar = ({
       {!isMap && (
         <div className={styles.infoRow}>
           <div className={styles.countContainer}>
+            {/* TODO: it should wait for the 2 requests to finish (wishlist items & items) */}
```
LautaroPetaccio, line 171: I've realised that this component doesn't use the `isLoading` prop at all. What if we use it alongside the loading of the favorites loading request to show a `Loading` placeholder or to not show anything until the loading phase finishes?

LautaroPetaccio (commented):
I've just left some comments so we can review them

PR #1645, opened 2023-05-01
refactor: Refactor fetch favorited items

### LINKED ISSUE #1624: When switching from a tab to My Lists, the counter shows different numbers in the displayed number of items

LautaroPetaccio:
This PR refactors how we retrieve and process favorited items to fix #1624.

### DIFF
--- a/webapp/src/modules/item/reducer.spec.ts
+++ b/webapp/src/modules/item/reducer.spec.ts
@@ -8,12 +8,19 @@ import { loadingReducer } from 'decentraland-dapps/dist/modules/loading/reducer'
 import { NetworkGatewayType } from 'decentraland-ui'
 import { View } from '../ui/types'
 import {
+  FETCH_FAVORITED_ITEMS_REQUEST,
+  fetchFavoritedItemsRequest,
+  fetchFavoritedItemsSuccess
+} from '../favorites/actions'
+import {
   buyItemFailure,
   buyItemRequest,
   buyItemSuccess,
   buyItemWithCardFailure,
   buyItemWithCardRequest,
   buyItemWithCardSuccess,
+  FETCH_ITEM_SUCCESS,
+  FETCH_TRENDING_ITEMS_REQUEST,
   fetchItemFailure,
   fetchItemRequest,
   fetchItemsFailure,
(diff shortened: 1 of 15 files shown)


webapp/src/modules/item/reducer.spec.ts
```diff
@@ -160,6 +167,44 @@ describe('when reducing the successful action of fetching items', () => {
...
 })
 
+describe.each([
+  [
+    FETCH_ITEM_SUCCESS,
+    fetchItemRequest(item.contractAddress, item.itemId),
+    fetchItemSuccess(item)
+  ],
+  [
+    FETCH_TRENDING_ITEMS_REQUEST,
```
kevinszuchet, line 177:
```suggestion
    FETCH_TRENDING_ITEMS_SUCCESS,
```

webapp/src/modules/item/reducer.spec.ts
```diff
@@ -160,6 +167,44 @@ describe('when reducing the successful action of fetching items', () => {
...
+    fetchTrendingItemsRequest(trendingItemsBatchSize),
+    fetchTrendingItemsSuccess([item])
+  ],
+  [
+    FETCH_FAVORITED_ITEMS_REQUEST,
```
kevinszuchet, line 182:
```suggestion
    FETCH_FAVORITED_ITEMS_SUCCESS,
```

webapp/src/modules/ui/browse/reducer.ts
```diff
@@ -165,9 +165,28 @@ export function browseReducer(
...
+            count: action.payload.total
+          }
+        }
+        default:
+          return {
+            ...state,
+            itemIds: action.payload.items.map(item => item.id),
+            count: action.payload.total
+          }
       }
```
kevinszuchet, line 190: WDYT about writing:
```ts
return {
  ...state,
  itemIds:
    view === View.LOAD_MORE
      ? [...state.itemIds, ...action.payload.items.map((item) => item.id)]
      : action.payload.items.map((item) => item.id),
  count: action.payload.total,
}
```?

File: webapp/src/modules/item/reducer.spec.ts (hunk 1 above)
Comment (LautaroPetaccio) [reply webapp/src/modules/item/reducer.spec.ts:177]: Fixed it!

File: webapp/src/modules/item/reducer.spec.ts (hunk 2 above)
Comment (LautaroPetaccio) [reply webapp/src/modules/item/reducer.spec.ts:182]: Fixed it!

File: webapp/src/modules/ui/browse/reducer.ts (hunk 3 above)
Comment (LautaroPetaccio) [reply webapp/src/modules/ui/browse/reducer.ts:190]: I've just re-written the reducer to work with the load more view, would you mind checking it again?

File: webapp/src/modules/ui/browse/sagas.ts [hunk 4]
```diff
@@ -18,19 +18,22 @@ export function* browseSaga() {
...
-      fetchFavoritedItemsRequest({
-        filters: { first: 1, skip: page * PAGE_SIZE - 1 },
-        page,
-        section: Section.LISTS
-      })
+      fetchFavoritedItemsRequest(
+        {
+          filters: { first: 1, skip: currentPage * PAGE_SIZE - 1 },
+          // Sets the page taking into consideration the 1 item page size
+          page: currentPage * PAGE_SIZE - 1
```
kevinszuchet, line 33: NSIU why the `page` needs to be equal to the `skip` property. Would you mind helping me understand this?

LautaroPetaccio: The page property here can be deleted, I'll remove it so it doesn't cause confusion, but the page is the same as the skip property because the pages are of size 1.

juanmahidalgo (approved):
the flag approach seems the most straightfoward at the moment, so LGTM!

PR #1647, opened 2023-05-01
fix: authorization after rent confirmation

meelrossi:
Move authorization check after rent confirmation instead of before

https://user-images.githubusercontent.com/11800206/235533101-d807b89b-a7de-43a8-a728-9389e453442f.mov

### DIFF
--- a/webapp/src/modules/rental/selector.spec.ts
+++ b/webapp/src/modules/rental/selector.spec.ts
@@ -2,15 +2,17 @@ import { NFTCategory, RentalListing } from '@dcl/schemas'
 import { LoadingState } from 'decentraland-dapps/dist/modules/loading/reducer'
 import { RootState } from '../reducer'
 import { NFT } from '../nft/types'
-import { claimAssetRequest, removeRentalRequest } from './actions'
+import { acceptRentalListingRequest, claimAssetRequest, removeRentalRequest } from './actions'
 import {
   getState,
   getData,
   getRentalById,
   isSubmittingTransaction,
   isClaimingAsset,
-  isRemovingRental
+  isRemovingRental,
+  getRentConfirmationStatus
 } from './selectors'
+import { AuthorizationStepStatus } from '../../components/HOC/withAuthorizedAction/AuthorizationModal'
 
 let rootState: RootState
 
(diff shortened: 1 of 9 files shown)


webapp/src/modules/rental/selectors.ts
```diff
@@ -23,3 +24,19 @@ export const isAcceptingRental = (state: RootState) =>
   isLoadingType(getLoading(state), ACCEPT_RENTAL_LISTING_REQUEST)
 export const isSubmittingTransaction = (state: RootState) =>
   getState(state).isSubmittingTransaction
+
+export const getRentStatus = (state: RootState) => {
```
LautaroPetaccio, line 28: Would you mind adding tests for this selector?

LautaroPetaccio, line 28: What do you think about renaming the function to be in accordance to the name of the operation? The idea is to avoid confusion with the rental status per se (the one that's on the server).
```suggestion
export const getRentConfirmationStatus = (state: RootState) => {
```

LautaroPetaccio (changes requested):
Looks good! ✅  I've just left two comments to be reviewed.

webapp/src/modules/rental/selector.spec.ts
```diff
@@ -134,3 +136,41 @@ describe('when getting if the transaction is being submitted', () => {
     })
   })
 })
+
+describe("when getting rental status", () => {
```
LautaroPetaccio, line 140: The word `a` is missing here, as we need to use an article:
```suggestion
describe("when getting a rental confirmation status", () => {
```
Also, I've added the `confirmation` word so it doesn't get confused with the rental status.

webapp/src/modules/rental/selector.spec.ts
```diff
@@ -134,3 +136,41 @@ describe('when getting if the transaction is being submitted', () => {
...
+  describe("and there is a transaction submitting", () => {
+    beforeEach(() => {
+      rootState.rental.isSubmittingTransaction = true
+    })
+    
+    it('should return authorization status WAITING', () => {
```
LautaroPetaccio, line 146: The word `an` is missing here, as we need to use an article:
```suggestion
    it('should return an authorization status WAITING', () => {
```

LautaroPetaccio (approved):
Looks great, I've left to minor typos

LautaroPetaccio (approved):
Ship it 🛥️

PR #1650, opened 2023-05-02
feat: deprecate the LOAD_MORE View

### LINKED ISSUE #1594: Collectibles tab showing Trending items while loading new ones.
## 🎉 Description

As it can be seen in the video, the collectibles tab shows the items from the trending section while loading new ones.

https://images.zenhubusercontent.com/226880692/ef5f8b1a-7e9a-446b-bead-ba085a28bf3c/listing_bug.mov

* [X] 🐛 This is a bug report.

### LINKED ISSUE #1649: [Refactor] Deprecate View.LOAD_MORE
## 🎉 Description

The `View` enum right now has an special entry for the `LOAD_MORE` interaction. When we're loading more assets (either nfts or items), the current UI view is overriden by the LOAD_MORE so we don't know where the user is navigating (we could read the URL, but we want to keep using the `view` field). Let's deprecate the LOAD_MORE in favor of storing the `page` in the UI state and with it, computing if we need to concatenate results or not.  

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #1649
Closes #1594

This PR deprecates the `View.LOAD_MORE` in favor of storing the `page` in the reducer and add logic to know if we're loading more elements or not.

### DIFF
--- a/webapp/src/modules/ui/browse/reducer.ts
+++ b/webapp/src/modules/ui/browse/reducer.ts
@@ -24,6 +24,7 @@ import { BrowseAction, BROWSE } from '../../routing/actions'
 import { Section } from '../../vendor/decentraland'
 import { SetViewAction, SET_VIEW } from '../actions'
 import { View } from '../types'
+import { isLoadingMoreResults } from './utils'
 
 export type BrowseUIState = {
   view?: View
@@ -102,8 +103,9 @@ export function browseReducer(
         return state
       }
       const newNftIds = nfts.map(nft => nft.id)
-      const isLoadingMore = !!state.page && !!page && page > state.page
-      const nftIds = isLoadingMore ? [...state.nftIds, ...newNftIds] : newNftIds
+      const nftIds = isLoadingMoreResults(state, page)
+        ? [...state.nftIds, ...newNftIds]
+        : newNftIds
       switch (view) {
         case View.MARKET:
         case View.CURRENT_ACCOUNT:
@@ -139,8 +141,7 @@ export function browseReducer(
         }
       }
 
-      const isLoadingMore = !!page && page > (state.page || 0)
-      const elements = isLoadingMore ? [...state[key]] : []
+      const elements = isLoadingMoreResults(state, page) ? [...state[key]] : []
       switch (view) {
         case View.ATLAS:
           return state
@@ -182,8 +183,7 @@ export function browseReducer(
       }
 
       const newItemIds = items.map(item => item.id)
-      const isLoadingMoreResults = !!state.page && !!page && page > state.page
-      const itemIds = isLoadingMoreResults
+      const itemIds = isLoadingMoreResults(state, page)
         ? [...state.itemIds, ...newItemIds]
         : newItemIds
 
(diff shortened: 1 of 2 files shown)


webapp/src/components/AssetList/AssetList.tsx
```diff
@@ -79,6 +78,74 @@ const AssetList = (props: Props) => {
...
+      <div className="empty">
+        <div className="watermelon" />
+        <T
+          id={emptyStateTranslationString}
+          values={{
+            search,
+            currentSection:
+              assetType === AssetType.ITEM
+                ? t('browse_page.primary_market_title').toLocaleLowerCase()
+                : t('browse_page.secondary_market_title').toLocaleLowerCase(),
```
kevinszuchet, line 92: Would you mind storing these values in two variables called `primaryMarketTitle` and `secondaryMarketTitle` and re-using them in the section property when necessary?

webapp/src/components/AssetList/AssetList.tsx
```diff
@@ -79,6 +78,74 @@ const AssetList = (props: Props) => {
...
+  ])
+
+  const shouldRenderEmptyState = useMemo(
+    () => assets.length === 0 && !isLoading,
+    [assets.length, isLoading]
+  )
+
+  const renderAssetCards = useCallback(
+    () =>
+      assets.map((assets, index) => (
```
kevinszuchet, line 139: Is this a typo? Would you mind renaming it to `asset`?

kevinszuchet (approved):

flobarreto (approved):

webapp/src/components/AssetList/AssetList.tsx
LautaroPetaccio: We could also rename it to section as store it as:
```
const section = assetType === AssetType.ITEM
                ? t('browse_page.primary_market_title').toLocaleLowerCase()
                : t('browse_page.secondary_market_title').toLocaleLowerCase()
```
taking into consideration that they're being used in the same way in both section and currentSection, right?
Btw, this is an out of scope fix, as this is the old code.

webapp/src/modules/ui/browse/reducer.ts
```diff
@@ -85,74 +91,68 @@ export function browseReducer(
...
+      const isDifferentView = view !== state.view
+      if (isDifferentView) {
+        return {
+          ...state,
+          [key]: [],
+          count: undefined
+        }
+      }
+
+      const isLoadingMore = !!page && page > (state.page || 0)
```
LautaroPetaccio, line 142: What if we extract the `isLoadingMore` procedure into a small function so we can use it in all cases? I'm asking this because it's a repeated and important procedure.

webapp/src/modules/ui/browse/sagas.ts
```diff
@@ -29,7 +28,7 @@ function* handleUnpickItemAsFavoriteSuccess() {
     yield put(
       fetchFavoritedItemsRequest({
         filters: { first: 1, skip: page * PAGE_SIZE - 1 },
-        view: View.LOAD_MORE,
+        page,
```
LautaroPetaccio, line 31: This seems to break the my lists view. The page here is the same page that's being viewed at the moment, making the reducer clear the item ids. In this PR it might be a bit difficult to fix, as you'll need a flag to know when this happens. We might be able to merge my PR into this one and fix it.

LautaroPetaccio (commented):
Great job! I've left some comments that we'll need to review.

LautaroPetaccio (approved):
🚀

PR #1654, opened 2023-05-03
feat: Force getting the identity for the favorites feature

### LINKED ISSUE #1638: Error when trying to pick / unpick an item or going into My List if there's no identity
## 🎉 Description

As the log in process in the marketplace does not prevent the user from cancelling the creation of the identity in the Marketplace, the pick / unpick and the My List section could not work if the user decided not to sign the ephemeral key.

LautaroPetaccio:
This PR mitigates the case where there's no identity loaded and the user wants to perform a favorites action.
Closes #1638

### DIFF
--- a/webapp/src/modules/favorites/sagas.spec.ts
+++ b/webapp/src/modules/favorites/sagas.spec.ts
@@ -165,7 +165,7 @@ describe('when handling the request for unpicking a favorite item', () => {
     })
   })
 
-  describe('and the call to the favorites api fauks', () => {
+  describe('and the call to the favorites api fails', () => {
     it('should dispatch an action signaling the failure of the handled action', () => {
       return expectSaga(favoritesSaga, getIdentity)
         .provide([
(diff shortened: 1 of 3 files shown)


webapp/src/modules/favorites/sagas.spec.ts
```diff
@@ -138,10 +155,21 @@ describe('when handling the request for picking an item as favorite', () => {
...
+    it('should dispatch an action signaling the failure of the handled action', () => {
+      return expectSaga(favoritesSaga, getIdentity)
+        .provide([[call(getAccountIdentity), Promise.reject(error)]])
+        .put(unpickItemAsFavoriteFailure(item, error.message))
+        .dispatch(unpickItemAsFavoriteRequest(item))
+        .run({ silenceTimeout: true })
+    })
+  })
+
+  describe('and the call to the favorites api fauks', () => {
```
kevinszuchet, line 168:
```suggestion
  describe('and the call to the favorites api fails', () => {
```

PR #1661, opened 2023-05-04
fix: Asset list changing positions when loading

### LINKED ISSUE #1659: The content of My Lists moves a little bit to to top after loading completely

LautaroPetaccio:
This PR fixes an issue where when loading a set of assets, the Assets would shift a bit down the page until. This was caused because the sticky position of the loader made the site take into consideration the height of it, making it grow although it was in other position.
As the loading is always paired with an overlay, this PR creates a new overlay that is transparent with absolute position where the loader can sit without increasing the size while being in the center.

Closes #1659

### DIFF
--- a/webapp/src/components/AssetList/AssetList.css
+++ b/webapp/src/components/AssetList/AssetList.css
@@ -25,6 +25,12 @@
     left: 0;
     margin: auto;
   }
+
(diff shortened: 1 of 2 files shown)


kevinszuchet (approved):
LGTM 👍

PR #1664, opened 2023-05-04
feat: added empty state

flobarreto:


webapp/src/modules/routing/utils.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { omit, reset } from '../../lib/utils'
 import { View } from '../ui/types'
 import { getSearchParams } from './search'
+import { AssetStatusFilter } from '../../utils/filters'
```
juanmahidalgo, line 11: please move this import a few lines above 🙏

webapp/src/modules/routing/sagas.spec.ts
```diff
@@ -115,7 +116,7 @@ describe('when handling the clear filters request action', () => {
     })
 
     describe('and it is not the LAND section', () => {
-      it("should fetch assets and change the URL by clearing the filter's browse options and restarting the page counter and delete the onlyOnSale filter", () => {
+      fit("should fetch assets and change the URL by clearing the filter's browse options and restarting the page counter and delete the onlyOnSale filter", () => {
```
juanmahidalgo, line 119:
```suggestion
      it("should fetch assets and change the URL by clearing the filter's browse options and restarting the page counter and delete the onlyOnSale filter", () => {
```

juanmahidalgo (approved):
LGTM!

PR #1666, opened 2023-05-05
fix: authorization flow states

meelrossi:
- Add logic to call new authorization flow saga that will handle all steps of authorization.
- Add not enough allowance error


cyaiox (approved):
amazing job!

PR #1675, opened 2023-05-10
fix: Add fetch assets from route action to Going back router action

### LINKED ISSUE #1674: Going back should fetch results with the previous params.
## 🎉 Description

With the current UX, when a user goes back, the cards are not reloaded but the filters are removed. The expected behavior should be to go back to the previous param configuration and re-fetch the results.

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #1674

## The bug

https://github.com/decentraland/marketplace/assets/8763687/56df1b9e-604c-4480-94c0-726a635777c4

## The fix

https://github.com/decentraland/marketplace/assets/8763687/89538cac-4df1-419e-8532-61a963f2b666

### DIFF
--- a/webapp/src/modules/routing/sagas.spec.ts
+++ b/webapp/src/modules/routing/sagas.spec.ts
@@ -6,7 +6,7 @@ import {
   NFTCategory,
   Rarity
 } from '@dcl/schemas'
-import { getLocation, push } from 'connected-react-router'
+import { getLocation, LOCATION_CHANGE, LocationChangeAction, push, RouterLocation } from 'connected-react-router'
 import { expectSaga } from 'redux-saga-test-plan'
 import { call, select } from 'redux-saga/effects'
 import { AssetType } from '../asset/types'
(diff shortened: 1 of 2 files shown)


webapp/src/modules/routing/sagas.ts
```diff
@@ -70,6 +71,7 @@ import {
   BROWSE,
   BrowseAction,
   FETCH_ASSETS_FROM_ROUTE,
+  fetchAssetsFromRoute as FetchAssetsFromRoute,
```
kevinszuchet, line 74: I'd rename this to `fetchAssetsFromRouteAction` if possible for standards purposes. WDYT?

kevinszuchet (dismissed):

webapp/src/modules/routing/sagas.ts
juanmahidalgo: I was going to use that name but I saw the line from below that the same name. But I guess wthat if if doesn't start with capital letter it's just fine. Updated!

kevinszuchet: Could you rename it also in the tests file? We still have an import with the following alias: `fetchAssetsFromRoute as FetchAssetsFromRouteAction`.

webapp/src/modules/routing/sagas.spec.ts
```diff
@@ -25,7 +31,7 @@ import {
   fetchAssetsFromRoute as FetchAssetsFromRouteAction
```
kevinszuchet, line 31: I'd also rename this. WDYT?

juanmahidalgo: sure, done 👍

kevinszuchet (approved):

PR #1685, opened 2023-05-12
feat: Fetch lists

### LINKED ISSUE #1680: Create the fetch lists feature's sagas
## 🎉 Description

Create the actions, reducer and sagas to fetch the user's lists taking the following into consideration:
- The favorites reducer will now store lists. 
- The new sagas will use the `GET /v1/lists` endpoint to retrieve the user's list.
- The call to the `GET /v1/lists` endpoint must be authenticated.

LautaroPetaccio:
This PR adds the fetch lists sagas, reducers and selectors.
Closes #1680

### DIFF
--- a/webapp/src/modules/ui/browse/reducer.spec.ts
+++ b/webapp/src/modules/ui/browse/reducer.spec.ts
@@ -17,10 +17,10 @@ import { NFT, NFTsFetchOptions } from '../../nft/types'
 import { browse } from '../../routing/actions'
 import { VendorName } from '../../vendor'
 import { Section } from '../../vendor/decentraland'
+import { List } from '../../favorites/types'
 import { setView } from '../actions'
 import { View } from '../types'
 import { BrowseUIState, INITIAL_STATE, browseReducer } from './reducer'
-import { List } from '../../favorites/types'
 
 const assetIds = ['0x0-assetId1', '0x0-assetId2', '0x0-assetId3']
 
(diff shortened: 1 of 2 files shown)


webapp/src/modules/ui/browse/reducer.spec.ts
```diff
@@ -19,6 +20,7 @@ import { Section } from '../../vendor/decentraland'
 import { setView } from '../actions'
 import { View } from '../types'
 import { BrowseUIState, INITIAL_STATE, browseReducer } from './reducer'
+import { List } from '../../favorites/types'
```
kevinszuchet, line 23: Would you mind moving this up?

webapp/src/modules/vendor/decentraland/favorites/api.ts
```diff
@@ -79,4 +79,22 @@ export class FavoritesAPI extends BaseClient {
...
+    }
+
+    if (filters.skip) {
+      queryParams.append('offset', filters.skip.toString())
+    }
+
+    return this.fetch(
+      `/v1/lists` + (queryParams.toString() && `?${queryParams.toString()}`)
+    )
+  }
```
kevinszuchet, line 99: Would you mind abstracting the logic of appending the query parameters and concatenating them to the endpoint in a new function? We may call it `getPaginatedResources(endpoint: string, filters: ItemFilters, queryParams?: URLSearchParams)`. Both `getLists` and `getPicksByList` could use it. 

WDYT?

LautaroPetaccio: I think it's a great idea, we were doing it manually in several endpoints. I've build a new function, `buildURLWithParameters` that gets the ednpoint and the item filters. I've not included the queryParams parameter because they should all come from the `filters` parameter.

webapp/src/modules/ui/browse/reducer.spec.ts
LautaroPetaccio: Done!

kevinszuchet (approved):

PR #1688, opened 2023-05-12
feat: Delete list sagas

### LINKED ISSUE #1682: Create the delete list feature's sagas
## 🎉 Description

Create the delete lists feature's actions, sagas and reducers.
These sagas will use the `DELETE /v1/lists/:id` endpoint. Take into consideration that the call to the endpoint must be authenticated.

LautaroPetaccio:
This PR adds the delete list sagas and reducers.

Closes #1682

### DIFF
--- a/webapp/src/modules/vendor/decentraland/favorites/api.ts
+++ b/webapp/src/modules/vendor/decentraland/favorites/api.ts
@@ -1,7 +1,7 @@
 import { BaseClient } from 'decentraland-dapps/dist/lib/BaseClient'
 import { config } from '../../../../config'
-import { FavoritedItems } from '../../../favorites/types'
 import { isAPIError } from '../../../../lib/error'
+import { FavoritedItems, List } from '../../../favorites/types'
 import { ItemFilters } from '../item/types'
 
 export const DEFAULT_FAVORITES_LIST_ID = config.get(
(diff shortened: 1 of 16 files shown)


webapp/src/modules/vendor/decentraland/favorites/api.ts
```diff
@@ -97,4 +97,8 @@ export class FavoritesAPI extends BaseClient {
       `/v1/lists` + (queryParams.toString() && `?${queryParams.toString()}`)
     )
   }
+
+  async deleteList(listId: string): Promise<void> {
+    return this.fetch(`/v1/list/${listId}`)
```
kevinszuchet, line 102: I think the `DELETE` method is missing here, right?

webapp/src/modules/vendor/decentraland/favorites/api.ts
```diff
@@ -97,4 +97,8 @@ export class FavoritesAPI extends BaseClient {
       `/v1/lists` + (queryParams.toString() && `?${queryParams.toString()}`)
     )
   }
+
+  async deleteList(listId: string): Promise<void> {
+    return this.fetch(`/v1/list/${listId}`, { method: 'DELETE' })
```
kevinszuchet, line 102:
```suggestion
    return this.fetch(`/v1/lists/${listId}`, { method: 'DELETE' })
```

kevinszuchet (approved):

PR #1690, opened 2023-05-15
feat: Track Refunded purchases

kevinszuchet:
Related to https://github.com/decentraland/decentraland-dapps/pull/434


webapp/src/modules/asset/sagas.ts
```diff
@@ -51,7 +51,13 @@ function* handleSetAssetPurchaseWithCard(action: SetPurchaseAction) {
       yield put(push(statusPagePathname))
     }
 
-    if (status === PurchaseStatus.FAILED) {
+    const failStatuses = [
```
juanmahidalgo, line 54:
```suggestion
    export const failStatuses = [
```
so we can then import this in the test? so if we have to modify this in the future, it's only in one place, wdyt?

juanmahidalgo (approved):
LGTM!

PR #1695, opened 2023-05-15
Feat/my bid

### LINKED ISSUE #1442: [4.4] Add the “Other offers” section
This section will list other bids. Use the /bids endpoints that we currently have, it should return all the data needed for the table. It should be paginated and be sortable.

flobarreto:
closes #1442 

image
image


juanmahidalgo (approved):
looks great 🚀

PR #1702, opened 2023-05-17
feat: order nft page

### LINKED ISSUE #1439: [4.1] Re-order the nft basic info into the new component.
It should re-use the componentes used for the item. Let’s create those components in a way it renders the fields coming from both item/nft.

flobarreto:
closes #1439 

image


juanmahidalgo (approved):
Looks great! 👏

PR #1709, opened 2023-05-17
feat: Add the create list sagas

### LINKED ISSUE #1683: Create the create lists feature's sagas
## 🎉 Description

Create the create lists feature's actions, sagas and reducers.
These sagas will use the `POST /v1/lists` endpoint. Take into consideration that the call to the endpoint must be authenticated.
The endpoint will receive the list name in the content's body.

LautaroPetaccio:
This PR adds the create list sagas which will allow us, through the marketplace-favorites-server to create new lists.
Closes #1683


webapp/src/modules/favorites/actions.ts
```diff
@@ -222,3 +222,21 @@ export const updateListFailure = (id: string, error: string) =>
 export type UpdateListRequestAction = ReturnType<typeof updateListRequest>
 export type UpdateListSuccessAction = ReturnType<typeof updateListSuccess>
 export type UpdateListFailureAction = ReturnType<typeof updateListFailure>
+
+// Create List
+export const CREATE_LIST_REQUEST = '[Request] Create List'
+export const CREATE_LIST_SUCCESS = '[Success] Create List'
+export const CREATE_LIST_FAILURE = '[Failure] Create List'
+
+export const createListRequest = (name: string, isPrivate: boolean) =>
```
kevinszuchet, line 231: WDYT about receiving a partial list in the parameters? It's to avoid refactoring later if we decide to allow users to fill more information about the lists (i.e. `description`).

LautaroPetaccio: Added the description!

webapp/src/modules/vendor/decentraland/favorites/api.ts
```diff
@@ -113,4 +113,18 @@ export class FavoritesAPI extends BaseClient {
       }
     })
   }
+
+  async createList(
+    name: string,
+    isPrivate: boolean,
+    description?: string
+  ): Promise<List> {
```
kevinszuchet, line 121: WDYT about receiving the new list in the parameters instead of different arguments?
If we need to add a new property to the list, we'll need to change this signature and all its implementations.

LautaroPetaccio: Done!

kevinszuchet (approved):

PR #1710, opened 2023-05-17
feat: remove error toast on authorization modal

meelrossi:
Add new param `silent` in `executeOrderRequest` saga so that it doesn't show a toast if transaction fails. We are using this as true only when the action was triggered by the authorization modal, as we are already showing the error in the modal so there is no need to show also the toast


cyaiox (approved):
🚀

PR #1731, opened 2023-05-24
feat: Remove favorites FF

LautaroPetaccio:
This PR removes the favorites FF that is not longer being needed.


kevinszuchet (approved):
Thanks for this! 🚀

PR #1734, opened 2023-05-25
fix: do not fetch assets if they are in the store

juanmahidalgo:
Since the data is already in the store going back from the detail to the /browse again, we don't need to fetch it again. This PR avoids the re-fetch of the items.
The `fetchItemsRequest` is triggered in the detail page to get the collection items and it resets the `ui` state so I had to either:
- pass a flag to avoid reseting the `ui` state (so it doesn't fetch again the items)
- create a new action that doesn't override the `ui`. (fetching collection items for the detail shouldn't update the `ui`)

### DIFF
--- a/webapp/src/components/AssetBrowse/AssetBrowse.tsx
+++ b/webapp/src/components/AssetBrowse/AssetBrowse.tsx
@@ -56,11 +56,11 @@ const AssetBrowse = (props: Props) => {
   const location = useLocation()
   const history = useHistory()
   // Prevent fetching more than once while browsing
-  const latestLocation = visitedLocations[visitedLocations.length - 2]
+  const lastLocation = visitedLocations[visitedLocations.length - 2]
   const [hasFetched, setHasFetched] = useState(
     history.action === 'POP' &&
-      latestLocation?.pathname === location.pathname &&
-      latestLocation?.search === location.search
+      lastLocation?.pathname === location.pathname &&
+      lastLocation?.search === location.search
   )
   const isCurrentAccount = view === View.CURRENT_ACCOUNT
   const isAccountOrCurrentAccount = view === View.ACCOUNT || isCurrentAccount
--- a/webapp/src/components/AssetBrowse/AssetBrowse.tsx
+++ b/webapp/src/components/AssetBrowse/AssetBrowse.tsx
@@ -1,5 +1,5 @@
 import React, { ReactNode, useEffect, useState } from 'react'
-import { matchPath } from 'react-router-dom'
+import { matchPath, useHistory, useLocation } from 'react-router-dom'
 import classNames from 'classnames'
 import { Container, Mobile, NotMobile, Page, Tabs } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
@@ -53,8 +53,15 @@ const AssetBrowse = (props: Props) => {
     isMapViewFiltersEnabled
   } = props
 
+  const location = useLocation()
+  const history = useHistory()
   // Prevent fetching more than once while browsing
-  const [hasFetched, setHasFetched] = useState(false)
+  const latestLocation = visitedLocations[visitedLocations.length - 2]
+  const [hasFetched, setHasFetched] = useState(
+    history.action === 'POP' &&
+      latestLocation?.pathname === location.pathname &&
+      latestLocation?.search === location.search
+  )
   const isCurrentAccount = view === View.CURRENT_ACCOUNT
   const isAccountOrCurrentAccount = view === View.ACCOUNT || isCurrentAccount
   const [showOwnedLandOnMap, setShowOwnedLandOnMap] = useState(true)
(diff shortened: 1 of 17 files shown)


webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -53,8 +53,15 @@ const AssetBrowse = (props: Props) => {
     isMapViewFiltersEnabled
   } = props
 
+  const location = useLocation()
+  const history = useHistory()
   // Prevent fetching more than once while browsing
-  const [hasFetched, setHasFetched] = useState(false)
+  const latestLocation = visitedLocations[visitedLocations.length - 2]
+  const [hasFetched, setHasFetched] = useState(
```
juanmahidalgo, line 60: `hasFetched` is a mechanism that we have nowadays to fetch the data just once. If we're going back to the /browse and the data is in the store, we don't need to re-fetch (it should be `true`)

flobarreto (approved):

webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -53,8 +53,15 @@ const AssetBrowse = (props: Props) => {
     isMapViewFiltersEnabled
   } = props
 
+  const location = useLocation()
+  const history = useHistory()
   // Prevent fetching more than once while browsing
-  const [hasFetched, setHasFetched] = useState(false)
+  const latestLocation = visitedLocations[visitedLocations.length - 2]
```
LautaroPetaccio, line 59: What do you think about using `last` here instead of `latest`? Latest means "the most recent one".
```suggestion
  const lastLocation = visitedLocations[visitedLocations.length - 2]
```

webapp/src/modules/routing/sagas.ts
```diff
@@ -143,8 +145,16 @@ export function* routingSaga() {
...
-    yield put(fetchAssetsFromRouteAction(options))
+    const latestVisitedLocation: ReturnType<typeof getLocation> = yield select(
+      getLatestVisitedLocation
+    )
+    const isComingFromBrowse = !!matchPath(latestVisitedLocation?.pathname, {
+      path: locations.browse()
+    })
+    if (isComingFromBrowse) {
+      const options: BrowseOptions = yield select(getCurrentBrowseOptions)
+      yield put(fetchAssetsFromRouteAction(options))
```
LautaroPetaccio, line 156: I'm curious, the `fetchAssetsFromRouteAction` is used in the `HomePage` and in the `AssetBrowse` component, would this change affect the `HomePage` as well?

webapp/src/modules/item/sagas.spec.ts
```diff
@@ -352,6 +355,36 @@ describe('when handling the set purchase action', () => {
...
+        .put(fetchCollectionItemsSuccess(fetchResult.data))
+        .dispatch(
+          fetchCollectionItemsRequest({ contractAddresses: [], first: 10 })
+        )
+        .run({ silenceTimeout: true })
+    })
+  })
+
+  describe('when the request fails', () => {
+    it('should dispatching a failing action with the error and the options', () => {
```
kevinszuchet, line 374:
```suggestion
    it('should dispatch a failing action with the error and the options', () => {
```

webapp/src/modules/routing/sagas.spec.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from 'connected-react-router'
 import { expectSaga } from 'redux-saga-test-plan'
 import { call, select } from 'redux-saga/effects'
+import { locations } from './locations'
```
kevinszuchet, line 18: Would you mind moving this below?

webapp/src/modules/ui/browse/reducer.ts
```diff
@@ -71,6 +71,9 @@ export function browseReducer(
 ): BrowseUIState {
   switch (action.type) {
     case SET_VIEW: {
+      if (action.payload.view === state.view) {
```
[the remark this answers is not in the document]
juanmahidalgo: added!

webapp/src/components/AssetBrowse/AssetBrowse.tsx
juanmahidalgo: done!

webapp/src/modules/routing/sagas.ts
juanmahidalgo: good call! added some logic so in the home it doesn't trigger the re-fetch that is not needed

webapp/src/modules/item/sagas.spec.ts
juanmahidalgo: done!

webapp/src/modules/routing/sagas.spec.ts
juanmahidalgo: done!

LautaroPetaccio (approved):
Ship it 🚀

PR #1739, opened 2023-05-30
feat: refactor the BuyNFTButtons to accept asset fields instead of the entire object

juanmahidalgo:
Fixes a bug where the NFT can't be bought from the Item detail page because the `asset` passed was the item and not the NFT
Like the following case:
image

The item can't be mint, so the buy with mana button should send the user to the buy NFT flow.


webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
```diff
@@ -24,6 +24,7 @@ import noListings from '../../../images/noListings.png'
 import Mana from '../../Mana/Mana'
 import { ManaToFiat } from '../../ManaToFiat'
 import { formatWeiToAssetCard } from '../../AssetCard/utils'
+import { AssetType } from '../../../modules/asset/types'
```
kevinszuchet, line 27: Would you mind moving this import up?

PR #1747, opened 2023-05-31
feat: Add lists page

### LINKED ISSUE #1718: Create the Lists Page
In the Lists Page, the user should see:

1. A sorting dropdown with the following options:
- Name (asc & desc)
- Newest
- Oldest
- Recently Updated

2. A create new list button that should open the Create List modal
3. The number of displayed lists
4. A paginated list of the different Lists created by the user and the default list called Wishlist using the new List Card component

LautaroPetaccio:
This PR does the following:
- Renames the ListsPage component into ListPage
- Implements the ListsPage component, using the pagination hook.
- Adds the lists route to the `Routes` component.
- Sets the bases of a ListsPage that will contain the ListCard component and operate as a paginated component.

Closes #1718


webapp/src/modules/vendor/decentraland/favorites/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type ListsOptions = PaginationParameters & {
+  sortBy?: ListsSortBy
+  sortDirection?: SortDirection
+}
+
+export enum ListsSortBy {
+  NAME = 'name',
+  CREATED_AT = 'createdAt',
+  UPDATED_AT = 'updatedAt'
+}
```
kevinszuchet, line 19: We should [move this types to the common-schemas](https://github.com/decentraland/schemas/issues/184) lib to stop having them in two different places.

webapp/src/lib/pagination.ts
```diff
@@ -0,0 +1,87 @@
...
+    const page = parseInt(params.get('page') ?? '1')
+    const first = parseInt(params.get('first') ?? pageSize)
+    const offset = (page - 1) * first
+    return [page, first, offset]
+  }, [pageSize, search])
+
+  const goToNextPage = useCallback(() => {
+    const params = new URLSearchParams(search)
+    params.set('page', (page + 1).toString())
+    push(`${pathname}${params ? `?${params.toString()}` : ''}`)
```
kevinszuchet, line 52: Why is needed to check if the params exist? One line above we are adding a parameter, so at least it will have one, right?

If the validation is still necessary, WDYT about moving the call to `push` with `pathname/params` to a new function, we we can use it later in the `changeSorting` and `changeFilter` functions?

webapp/src/lib/pagination.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    })
+
+    describe('and there\'s a "first" parameter', () => {
+      beforeEach(() => {
+        useLocationMock.search = 'first=10'
+        renderedHook = renderHook(() => usePagination())
+        currentResult = renderedHook.result.current as UsePaginationResult
+      })
+
+      it('should return the value of the "first" parameter in the "first property"', () => {
```
kevinszuchet, line 47:
```suggestion
      it('should return the value of the "first" parameter in the "first" property', () => {
```

webapp/src/lib/pagination.spec.ts
```diff
@@ -0,0 +1,183 @@
...
```
kevinszuchet, line 77:
```suggestion
      it('should return the value of the "first" parameter in the "first" property', () => {
```

webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+const ListPage = ({ wallet, isConnecting, onRedirect }: Props) => {
+  // Redirect to signIn if trying to access current account without a wallet
+  const { pathname, search } = useLocation()
+
+  useEffect(() => {
+    if (!isConnecting && !wallet) {
+      onRedirect(locations.signIn(`${pathname}${search}`))
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isConnecting, wallet, onRedirect])
```
kevinszuchet, line 27: Why not using the `ProtectedRoute` component here?

webapp/src/components/ListsPage/ListsPage.container.ts
```diff
@@ -1,27 +1,28 @@
...
+    lists: getBrowseLists(state),
+    count: getCount(state)
   }
 }
 
-const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onRedirect: path => dispatch(replace(path))
-})
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps =>
+  bindActionCreators(
```
kevinszuchet, line 20: Are we using this instead of writing:
```ts
{
  onFetchLists: (...args) => dispatch(fetchListsRequest(...args)),
  onCreateList: () => dispatch(openModal('CreateListModal'))
}
```

webapp/src/components/ListsPage/ListsPage.tsx
```diff
@@ -1,52 +1,110 @@
...
+        onLoadMore={goToNextPage}
+        isLoading={isLoading}
+        maxScrollPages={3}
+      >
+        <Card.Group className={styles.cards}>
+          {lists.map(list => (
+            <Card key={list.id}>
+              <Card.Content>
+                <Card.Header>{list.name}</Card.Header>
+                <Card.Meta>x Items</Card.Meta>
```
kevinszuchet, line 101: The backend is already returning the number of items in the list. The property is called `itemsCount`. Could you use it here or is there something I'm missing?

webapp/src/components/PageLayout/PageLayout.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <div className={styles.page}>
+      <Navbar isFullscreen />
+      <Navigation activeTab={activeTab} />
+      <div className={styles.content}>{children}</div>
+      <Footer className={styles.footer} />
+    </div>
+  )
+}
+
+export default React.memo(PageLayout)
```
kevinszuchet, line 19: ❤️

webapp/src/components/ProtectedRoute/ProtectedRoute.container.ts
```diff
@@ -0,0 +1,14 @@
+import { connect } from 'react-redux'
+import ProtectedRoute from './ProtectedRoute'
```
kevinszuchet, line 2: Would you mind moving this import below?

webapp/src/components/ProtectedRoute/ProtectedRoute.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { Redirect, Route, useLocation } from 'react-router-dom'
+import { locations } from '../../modules/routing/locations'
+import { Props } from './ProtectedRoute.types'
+
+const ProtectedRoute = (props: Props) => {
+  const { isConnecting, wallet, ...rest } = props
```
kevinszuchet, line 7: WDYT about taking these props (`isConnecting` & `wallet`) from a custom container called `ProtectedRoute.container.ts`?

webapp/src/modules/favorites/sagas.spec.ts
```diff
@@ -43,7 +43,16 @@ import {
...
 import { favoritesSaga } from './sagas'
 import { getListId } from './selectors'
-import { FavoritedItems, List, Permission } from './types'
+import {
+  FavoritedItems,
+  List,
+  ListsBrowseOptions,
+  ListsBrowseSortBy,
+  Permission
+} from './types'
```
kevinszuchet, line 52: Would you mind moving this import below?

webapp/src/components/ListsPage/ListsPage.container.ts
LautaroPetaccio: `bindActionCreators` does exactly the same thing as the example and comes out of the box from redux. I can revert it, but I find it helpful as a helper, what do you think about it?

webapp/src/components/ListsPage/ListsPage.tsx
LautaroPetaccio: It is coming in the cards PR! I just wanted to show something in this PR to check that it worked.

webapp/src/components/ListPage/ListPage.tsx
kevinszuchet: Totally agree 👍

webapp/src/lib/pagination.ts
LautaroPetaccio: You're totally right, the validation is not required at all, removed it!

webapp/src/lib/pagination.spec.ts
LautaroPetaccio: Done!

webapp/src/components/ProtectedRoute/ProtectedRoute.container.ts
LautaroPetaccio: Done!

webapp/src/components/ProtectedRoute/ProtectedRoute.tsx
LautaroPetaccio: I'm not sure I understand the question because the container is above this file. Can you show me an example?

kevinszuchet: Sorry, I meant having a `ProtectedRoute.container.ts` like this:

```ts
const mapState = (state: RootState): MapStateProps =>
  ({
    wallet: getWallet(state),
    isConnecting: isConnecting(state),
  });
```

PR #1750, opened 2023-06-01
feat: Protect routes

LautaroPetaccio:
This PR uses the `ProtectedRoute` in the routes that require the page to be loaded with a wallet.
This protected route not only prevents the user from loading a page when there's no connected wallet but also automatically sets the `redirect` parameter so the user gets redirected automatically where it was when logging in.

### DIFF
--- a/webapp/src/components/Routes/Routes.tsx
+++ b/webapp/src/components/Routes/Routes.tsx
@@ -59,20 +59,32 @@ const Routes = ({ inMaintenance }: Props) => {
         />
         <Route exact path={locations.account()} component={AccountPage} />
         <ProtectedRoute exact path={locations.lists()} component={ListsPage} />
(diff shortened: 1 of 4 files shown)


webapp/src/components/Routes/Routes.tsx
```diff
@@ -69,20 +69,32 @@ const Routes = ({ inMaintenance }: Props) => {
         />
         <Route exact path={locations.account()} component={AccountPage} />
```
fzavalia, line 70: Shouldn't the account page be protected too?

fzavalia: I don't remember if it is also the page for other accounts which might not be necessary. But I think it is used for your own account as well.

fzavalia (approved):
biutiful

fzavalia (approved):
biutiful

PR #1757, opened 2023-06-02
feat: Create list card

### LINKED ISSUE #1700: Create the list card
Create the ListCard component that will be used to show cards in the Marketplace taking into consideration the following requirements:
- The card should load the thumbnail of the items that is has (the first 4 but can be less according to the amount of items the list has).
- It should show a tag if it’s private.
- It should have a thumbnail with an empty logo if there are no items in the list.
- It should have a kebab menu with a dropdown that has two options, Edit list and Delete list.
 - If the list is empty it should let the user to remove the list without confirmation, if not, it should open the Confirmation modal.
 - The edit button should open the Edit List modal. 
- It should display the list name and the amount of items.

LautaroPetaccio:
This PR adds the ListCard component that will be in charge of showing the user a preview of the list.
The code also modifies the types and the reducers to make a clear difference between what's being returned in the API and what's being stored in the reducers.

Closes #1700


webapp/src/modules/favorites/selectors.ts
```diff
@@ -62,3 +64,7 @@ export const isPickingOrUnpicking = (state: RootState, itemId: string) =>
   )
 
 export const getList = (state: RootState, id: string) => getLists(state)[id]
+export const getPreviewListItems = (state: RootState, id: string): Item[] =>
```
kevinszuchet, line 67: WDYT about adding some test cases for this selector?

webapp/src/components/ListsPage/ListCard/ListCard.spec.tsx
```diff
@@ -0,0 +1,193 @@
...
+  EDIT_LIST_DATA_TEST_ID,
+  EMPTY_PREVIEW_DATA_TEST_ID,
+  GRID_PREVIEW_DATA_TEST_ID,
+  ITEM_COUNT_DATA_TEST_ID,
+  LIST_NAME_DATA_TEST_ID,
+  default as ListCard,
+  PRIVATE_DATA_TEST_ID
+} from './ListCard'
+import { Props } from './ListCard.types'
+import { List } from '../../../modules/favorites/types'
```
kevinszuchet, line 15: Would you mind moving this up?

kevinszuchet (commented):
LGTM! I left a comment about adding a few tests and one for sorting up some imports.

kevinszuchet (approved):

kevinszuchet:
Print it 🖨️

PR #1760, opened 2023-06-02
feat: Add success page

meelrossi:
Add new success page when buying  an item or an nft


webapp/src/components/SuccessPage/SuccessPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        secondary
+                        to={locations.nft(contractAddress, tokenId)}
+                      >
+                        {t('success_page.success_state.view_item')}
+                      </Button>
+                    )}
+
+                    {(asset.category === NFTCategory.WEARABLE ||
+                      asset.category === NFTCategory.EMOTE) && (
+                      <Button
```
fzavalia, line 130: Can't the button be surrunded by an <a> or be made <Button as={"a"} /> So it can be interpreted as a link?

webapp/src/components/SuccessPage/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  const parsedLogs: LogDescription[] = logs
+    .map(log => {
+      try {
+        return collectionInterface.parseLog(log)
+      } catch (e) {
+        // Ignore log
+        return null
+      }
+    })
+    .filter(Boolean) as LogDescription[]
```
fzavalia, line 21: You can also:

```
for (const log of logs) {
  try {
    const parsedLog = collectionInterface.parseLog(log)

    if (log.name === 'Issue') {
      return log.args._issuedId
    }
  } catch (e) { }
}

return null
```

This prevents any extra iterations once found and the filter cast (`.filter(Boolean) as LogDescription[]`)

fzavalia (commented):
Good work!

Just added some observations :D

fzavalia:
The success animation is awesome!!!

webapp/src/components/AssetProvider/AssetProvider.tsx
```diff
@@ -12,16 +12,27 @@ const AssetProvider = (props: Props) => {
...
     rentalStatus,
     isLoadingFeatureFlags,
-    isLandOrEstate
+    isLandOrEstate,
+    retry,
+    error
   } = props
 
   const [hasLoadedInitialFlags, setHasLoadedInitialFlags] = useState(false)
+  const [hasFetchedOnce, setHasFetchedOnce] = useState(false)
```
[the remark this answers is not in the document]
meelrossi: i need to re render when this value changes as im using it to call the fetch again. The ref doesn't cause a rerender so doing fetchedOnce.current = false woudn't trigger the fetch again by itself

webapp/src/components/SuccessPage/SuccessPage.tsx
meelrossi: yess totally i missed this change

webapp/src/components/SuccessPage/SuccessPage.container.ts
```diff
@@ -0,0 +1,24 @@
...
+import { RootState } from '../../modules/reducer'
+import { getIssuedIdFromLogs } from './utils'
+import {
+  MapStateProps,
+} from './SuccessPage.types'
+import { SuccessPage } from './SuccessPage'
+
+const mapState = (state: RootState): MapStateProps => {
+  const search = new URLSearchParams(getSearch(state))
+  const transaction = getTransaction(state, search.get('txHash') || '')
```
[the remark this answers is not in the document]
meelrossi: It will not find any logs. I added so that it takes you to the item detail page in that case

fzavalia (approved):

PR #1761, opened 2023-06-05
fix: add sagas logic to cancel ongoing fetch items request effects

### LINKED ISSUE #1758: Race condition when applying filters
## 🎉 Description

If a filter is applied while a request in on the flight, it creats a race condition between the previous and the new request and might cause an invalid set of results (if the 2nd request takes longer than the first one).

* [X] 🐛 This is a bug report.

juanmahidalgo:
PR description
The `/browse` page can have multiple race-conditions when applying filters when there are ongoing requests that haven't been fullfiled yet. In order to avoid them, we need to cancel the ongoing fetch item requests effects.

Closes #1758

copilot:summary

### DIFF
--- a/webapp/src/modules/item/sagas.ts
+++ b/webapp/src/modules/item/sagas.ts
@@ -1,6 +1,17 @@
+import { matchPath } from 'react-router-dom'
+import { getLocation } from 'connected-react-router'
+import { SagaIterator } from 'redux-saga'
 import { Item } from '@dcl/schemas'
 import { put, takeEvery } from '@redux-saga/core/effects'
-import { call, race, select, take } from 'redux-saga/effects'
+import {
+  call,
+  cancel,
+  cancelled,
+  fork,
+  race,
+  select,
+  take
+} from 'redux-saga/effects'
 import { ContractName, getContract } from 'decentraland-transactions'
 import { AuthIdentity } from 'decentraland-crypto-fetch'
 import { sendTransaction } from 'decentraland-dapps/dist/modules/wallet/utils'
@@ -20,6 +31,7 @@ import { isCatalogView } from '../routing/utils'
 import { waitForWalletConnectionIfConnecting } from '../wallet/utils'
 import { retryParams } from '../vendor/decentraland/utils'
 import { CatalogAPI } from '../vendor/decentraland/catalog/api'
+import { locations } from '../routing/locations'
 import {
   buyItemFailure,
   BuyItemRequestAction,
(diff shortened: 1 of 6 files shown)


flobarreto (approved):

webapp/src/modules/item/sagas.ts
```diff
@@ -76,6 +89,26 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
...
+      const {
+        pathname: currentPathname
+      }: ReturnType<typeof getLocation> = yield select(getLocation)
+
+      // if we have a task running in the browse path, we cancel the previous one
+      if (matchPath(currentPathname, { path: locations.browse() })) {
+        if (task && task.isRunning()) {
+          yield cancel(task)
+        }
+      }
```
LautaroPetaccio, line 107: What do you think about merging the ifs here?
```suggestion
      if (matchPath(currentPathname, { path: locations.browse() }) && task && task.isRunning()) {
        if () {
          yield cancel(task)
        }
      }
```

LautaroPetaccio: I have two questions about this: 
1. We do know what the current path is, but, how do we know if the last browse was performed from the `browse` path? I'm a bit in doubt if it matters if it was launched from another path.
2. The path when switching filters in the browse or collectibles page does not change (the query parameters do). Do we want to prevent race conditions when switching filters as well?

webapp/src/modules/item/sagas.ts
```diff
@@ -76,6 +89,26 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
   yield takeEvery(SET_PURCHASE, handleSetItemPurchaseWithCard)
   yield takeEvery(FETCH_ITEM_REQUEST, handleFetchItemRequest)
 
+  // to avoid race conditions, just one fetch items request is handled at once in the browse page
+  function* handleOneFetchItemsRequestAtOnce(): SagaIterator {
```
LautaroPetaccio, line 93: What if we parametrize this function? We could test it independently from the items actions as well.
```suggestion
  function* takeLatestByPath(actionType: string, path: string): SagaIterator {
```

LautaroPetaccio (commented):
It looks great! I've left some questions to understand a bit how it works.

webapp/src/modules/item/sagas.ts
juanmahidalgo: excellent idea! updated :)

LautaroPetaccio (approved):
✅

PR #1763, opened 2023-06-06
feat: Dynamic list detail page

### LINKED ISSUE #1719: Update the List Detail Page to use the id of the URL to fetch the items that were saved on it
The new detail page should add to the previous one:

1. A link to go back to the previous section.
2. A button with a kebab icon allowing users to edit and delete the list: this should open the edit / delete list modal.
 a. The default list cannot be deleted and should disable the delete option. 
3. A button to share the list (enabled only if the list is public): this should open the share list modal.
4. An empty list with the new styles.

Empty List:

kevinszuchet:
Closes #1719


webapp/src/components/ListsPage/ListCard/ListCard.tsx
```diff
@@ -7,6 +7,9 @@ import { DEFAULT_FAVORITES_LIST_ID } from '../../../modules/vendor/decentraland/
 import { AssetImage } from '../../AssetImage'
 import { Props } from './ListCard.types'
 import styles from './ListCard.module.css'
+import { AssetType } from '../../../modules/asset/types'
+import { Section } from '../../../modules/vendor/decentraland'
+import { View } from '../../../modules/ui/types'
```
LautaroPetaccio, line 12: Would you mind re-ordering these inputs?

webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -1,52 +1,146 @@
...
-            vendor={VendorName.DECENTRALAND}
-          />
-        ) : (
-          <div className={styles.emptyState}></div>
-        )}
-        <Footer className={styles.footer} />
-      </div>
-    </div>
+    <PageLayout activeTab={NavigationTab.MY_LISTS}>
+      {list ? (
```
LautaroPetaccio, line 62: I'm not sure if this is covered, but, what would happen if there's an error that's not a 404 loading the list?

webapp/src/modules/favorites/sagas.ts
```diff
@@ -352,6 +354,9 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
           isErrorWithMessage(error) ? error.message : 'Unknown error'
         )
       )
+
+      if (isAPIError(error) && error.status === 404)
+        yield put(push(locations.lists()))
```
LautaroPetaccio, line 359: What if we show the users a blank page with an error if this happens instead of redirecting the user here? This is kind of related to what Flor asked.
It might be a bit confusing for the user to be redirected to the lists page and there might be different issues that could appear as well that we should consider and show.
If we would like to remove these lines, we could implement the error page in a different PR.

webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -1,52 +1,146 @@
...
+                  className={classNames(styles.iconContainer, styles.share)}
+                  inverted
+                  compact
+                  onClick={onShareList && (() => onShareList(list))}
+                  disabled={list.isPrivate}
+                  data-testid={SHARE_LIST_BUTTON_TEST_ID}
+                >
+                  <Icon name="share alternate" />
+                </Button>
+                {/* TODO: should we change the icons to the ones used in the figma? */}
```
LautaroPetaccio, line 89: I think that they're pretty close to the Figma, there shouldn't be a problem with them.

LautaroPetaccio (changes requested):
Great job! I've left some comments that should be addressed before merging.

webapp/src/components/ListPage/ListPage.module.css
```diff
@@ -1,29 +1,100 @@
...
 }
 
-.header:global(.ui.header) {
-  margin-left: 24px;
+.header:global(.ui.header.large) {
+  margin-bottom: 7px;
+}
+
+.header:global(.ui.sub.header) {
+  /* TODO: to align the name with the description, the number of items now seems weird */
```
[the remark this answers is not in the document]
kevinszuchet: My bad, nothing to worry about. I'll remove the comment.

webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -1,52 +1,146 @@
...
+  listId,
+  list,
+  isLoading,
+  onFetchList,
+  onBack,
+  onEditList,
+  onDeleteList,
+  onShareList
+}: Props) => {
+  // TODO: what is pushing the page up?
```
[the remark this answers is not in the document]
kevinszuchet: Yes, it was fixed with the `min-height: 100vh`, I'll try to change it so we can keep the fix and be aligned with your comment.

webapp/src/components/ListPage/ListPage.tsx
kevinszuchet: Same as [here](https://github.com/decentraland/marketplace/pull/1763#discussion_r1221325291)

LautaroPetaccio (approved):
🚀

flobarreto (approved):

PR #1765, opened 2023-06-06
feat: Add the CreateOrEditListModal

### LINKED ISSUE #1720: Create the Edit List modal
Similar to the Create List modal, create a new modal Edit List to update the name, description, and visibility of the list.
The validations and required inputs will be the same as the other modal.

After saving the changes, the user should see the updates in the detail page.

**The default list cannot get its name modified.**

LautaroPetaccio:
This PR modifies the `CreateListModal` to be able to edit lists.
The modal has been changed to receive a list, populating it with the list's data to later perform an update using the `onListEdit` prop function.

Screenshot 2023-06-06 at 15 54 09

Closes #1720


webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditListModal.tsx
```diff
@@ -75,22 +84,31 @@ const CreateListModal = ({
   }, [isLoading, onClose])
 
   const isNameDuplicatedError = error?.includes(DUPLICATED_ERROR)
+  const listChanged = list
```
kevinszuchet, line 87: Why not using the `useState` hook for this and modifying this flag on the input (name, description, isPrivate) handlers?

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditList.spec.tsx
```diff
@@ -0,0 +1,426 @@
...
+    fireEvent.focusOut(
+      getByTestId(CREATE_OR_EDIT_LIST_NAME_DATA_TEST_ID).children[0]
+    )
+    expect(
+      getByTestId(CREATE_OR_EDIT_LIST_NAME_DATA_TEST_ID).nextSibling
+    ).not.toHaveTextContent('List names can contain up to 32 characters')
+  })
+})
+
+describe('when the describe input is focused', () => {
```
kevinszuchet, line 296:
```suggestion
describe('when the description input is focused', () => {
```

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditListModal.tsx
```diff
@@ -9,38 +9,47 @@ import {
...
-  'create-list-accept-button'
+export const CREATE_OR_EDIT_LIST_NAME_DATA_TEST_ID = 'create-or-edit-list-name'
+export const CREATE_OR_EDIT_LIST_DESCRIPTION_DATA_TEST_ID =
+  'create-or-edit-list-description'
+export const CREATE_OR_EDIT_LIST_PRIVATE_DATA_TEST_ID =
+  'create-or-edit-list-private'
+export const CREATE_OR_EDIT_LIST_CANCEL_BUTTON_DATA_TEST_ID =
+  'create-or-edit-list-cancel-button'
+export const CREATE_OR_EDIT_LIST_ACCEPT_BUTTON_DATA_TEST_ID =
+  'create-or-edit-list-accept-button'
```
flobarreto, line 26: what about moving this constants to a separate file ?

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditListModal.tsx
LautaroPetaccio: There were two reasons why:
1. It would require adding a `setIsChanged` to each of the input changes.
2. The most important thing about this check is to see if there's a difference between the current value and the value being proposed to be updated to prevent an unnecessary update, so we should check that it's different on each change.
I could wrap it around a `useMemo` hook, but the only things that could re-trigger a render are those values.

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditList.spec.tsx
LautaroPetaccio: Fixed it!

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditList.spec.tsx
```diff
@@ -0,0 +1,426 @@
...
+  it('should render the max length info message', () => {
+    const { getByTestId } = renderedModal
+    fireEvent.focusIn(getByTestId(CREATE_OR_EDIT_LIST_DESCRIPTION_DATA_TEST_ID))
+    expect(
+      getByTestId(CREATE_OR_EDIT_LIST_DESCRIPTION_DATA_TEST_ID).nextSibling
+    ).toHaveTextContent('List descriptions can contain up to 100 characters')
+  })
+})
+
+describe('when the describe input is blurred', () => {
```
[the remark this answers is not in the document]
LautaroPetaccio: Fixed it!

webapp/src/components/Modals/CreateOrEditListModal/CreateOrEditListModal.tsx
LautaroPetaccio: I've created the file and placed all the constants inside of it.

flobarreto (approved):

kevinszuchet (approved):

PR #1768, opened 2023-06-07
feat: When creating a list in the pick in bulk process, add the item to the created list

### LINKED ISSUE #1766: Handle creating a list to pick and item
## 🎉 Description

In order to fulfill the requirement of creating a list to pick an item through the `SaveToListModal`, we must code the behavior to pick the item after creating the new list.

Two of many options can be chosen for this case:
- Create a new sagas that will open the `CreateOrEditListModal` and wait for a successful creation to later launch the item picking.
- Modify or change the `CreateOrEditListModal` to accept an item and pick it after successfully creating the list.

* [ ] 🐛 This is a bug report.

### LINKED ISSUE #1805: Sometimes, the list detail page displays no results when the request to the catalog succeeds

kevinszuchet:
Closes #1766 
Closes #1805


flobarreto (approved):

flobarreto (approved):

webapp/src/modules/favorites/reducer.spec.ts
```diff
@@ -336,6 +341,52 @@ describe('when reducing the action of canceling a pick item as favorite', () =>
...
+      requestAction = bulkPickUnpickStart(item)
+      cancelAction = bulkPickUnpickCancel(item, error)
+
+      initialState = {
+        ...initialState,
+        loading: loadingReducer([], requestAction)
+      }
+    })
+
+    it('should return a state with an empty loading state', () => {
```
LautaroPetaccio, line 380:
```suggestion
    it('should return a state with an empty loading state and the error', () => {
```

webapp/src/modules/favorites/sagas.ts
```diff
@@ -400,6 +414,72 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
...
+      yield call(getAccountIdentity)
+
+      yield put(openModal('SaveToListModal', { item }))
+
+      const {
+        listCreationSuccess
+      }: {
+        listCreationSuccess: CreateListSuccessAction
+        listCreationFailure: CreateListFailureAction
+        modalClosed: CloseModalAction
```
LautaroPetaccio, line 457: What happens when the `CreateOrEditListModal` gets opened? Are both modals opened at the same time?
I'm wondering this because if I open the `SaveToListModal`, then click on the `Create list` button, keeping the `SaveToListModal` modal open to later close the `CreateOrEditListModal`, the `take(CLOSE_MODAL)` will win the `race` condition and exit without closing the `SaveToListModal`. In this case, if the user opens again the `CreateOrEditListModal`, the saga would have been already completed, preventing the item from being added if the user finishes the flow.
A fix for this would be to close the `SaveToListModal` when the `CreateOrEditListModal` gets opened.

webapp/src/modules/favorites/sagas.spec.ts
```diff
@@ -1271,6 +1278,101 @@ describe('when handling the request for creating a list', () => {
...
+          [call(getAccountIdentity), Promise.reject(error)]
+        ])
+        .put(bulkPickUnpickCancel(item, error.message))
+        .dispatch(bulkPickUnpickStart(item))
+        .run({ silenceTimeout: true })
+    })
+  })
+
+  describe('and the user tries to create a new list', () => {
+    describe('and the creation of the list succeeds', () => {
```
LautaroPetaccio, line 1356: Would you mind adding the cases where the creation of the lists fail or get handled by a close modal?
We can assert that the item will not be picked in these cases.

LautaroPetaccio (commented):
The PR looks great! I've left a couple of comments for us to review.

webapp/src/modules/favorites/sagas.ts
kevinszuchet: Nice catch, thanks! Solved it with your suggestion.

LautaroPetaccio (approved):
🍟

LautaroPetaccio (approved):
🍟  x 2

PR #1773, opened 2023-06-08
feat: Show errors when get lists fails

### LINKED ISSUE #1771: Show a view with an error message and a try again button if the GET `/lists` fails
When fetching the detail of a list, if the request fails with a status different to a 404, the user should see the following view:

kevinszuchet:
Closes #1771


webapp/src/components/ListsPage/ListsPage.tsx
```diff
@@ -30,18 +32,38 @@ const ListsPage = ({
...
+      if (!count && page !== 1) {
+        skip = 0
+        // Load all pages up to the last one
+        first = page * PAGE_SIZE
+      }
+      onFetchLists({ page, first, skip, sortBy: selectedSortBy })
+    },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [first, page, selectedSortBy, onFetchLists])
+    [count, page, onFetchLists, first, selectedSortBy]
```
LautaroPetaccio, line 48: If fetchLists changes each time the count changes, we'll re-trigger a second request every time a get lists request is performed and the count saved in the state. The fetch should get triggered if the `first`, `page`, `selectedSortBy` or `onFetchLists` change, which would re-trigger a fetch.

LautaroPetaccio (changes requested):
Looks great! 👍  We need to review a bit the `useEffect` and it should be good to go

flobarreto (dismissed):

LautaroPetaccio (approved):
🍟

PR #1778, opened 2023-06-09
feat: List detail public view

### LINKED ISSUE #1723: Allow users to see the detail page of a public list
If the user enters an URL pointing to a list that is public, the marketplace should render the List Detail Page without showing the actions allowed only for the owner of the list (share, edit, delete).
This logic should be implemented using the `userAddress` to check if the user owns the list and the `permissions` property to see if the user viewing the list has VIEW or EDIT permissions.
If the list is not public (the user is not the owner and does not have VIEW or EDIT permissions) the user should be redirected to a blank page.

The Navbar of the marketplace should be hidden, and the user shouldn't be able to go back.

kevinszuchet:
Closes #1723 

Public view with an item selected by the owner but not by the connected user:
image


flobarreto (approved):

LautaroPetaccio (approved):
Looks great! 👍 
Send it! ✈️

PR #1784, opened 2023-06-13
Feat/create save to list modal bulk

LautaroPetaccio:


webapp/src/components/Modals/SaveToListModal/SaveToListModal.module.css
```diff
@@ -0,0 +1,77 @@
...
+
+.checkbox {
+  margin-right: 15px;
+  height: 24px;
+  width: 24px;
+}
+
+.checkbox input {
+  height: 23px !important;
+  width: 23px !important;
```
kevinszuchet, line 44: Is there a way to avoid this `!important`s using a more specific selector?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.module.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.checkbox:global(.ui.checkbox) :global(label:before) {
+  width: 23px;
+  left: 1px;
+  height: 23px;
+}
+
+.checkbox:global(.ui.checkbox.checked:not(.toggle)) :global(input[type='checkbox'].hidden + label:after) {
```
kevinszuchet, line 53: This rule is already covered by the rule next to this in line 60. Would you mind removing it?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.spec.tsx
```diff
@@ -0,0 +1,196 @@
...
+  CREATE_LIST_BUTTON_DATA_TEST_ID,
+  SAVE_BUTTON_DATA_TEST_ID,
+  LISTS_LOADER_DATA_TEST_ID,
+  LIST_CHECKBOX,
+  LIST_NAME,
+  LIST_PRIVATE
+} from './SaveToListModal'
+import { Props } from './SaveToListModal.types'
+import { act } from 'react-dom/test-utils'
+import { fireEvent } from '@testing-library/react'
```
kevinszuchet, line 16: Would you mind moving these imports up?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+export const CREATE_LIST_BUTTON_DATA_TEST_ID = 'save-to-list-create-list-button'
+export const SAVE_BUTTON_DATA_TEST_ID = 'save-to-list-save-button'
+export const LIST_CHECKBOX = 'save-to-list-data-checkbox-'
+export const LIST_NAME = 'save-to-list-data-name-'
+export const LIST_ITEMS_COUNT = 'save-to-list-data-count-'
+export const LIST_PRIVATE = 'save-to-list-data-private-'
+
+const ITEM_HEIGHT = 60
+const DEFAULT_LIST_HEIGHT = 300
+const DEFAULT_LIST_WIDTH = 650
```
kevinszuchet, line 38: WDYT about moving these constants to a new file `constants.ts` to keep the consistency with the last changes in the feature?

webapp/src/modules/translation/locales/zh.json
```diff
@@ -1380,5 +1414,10 @@
     "title": "删除 <b>{name}</b> 列表",
     "message": "您确定要删除此列表吗？ 里面的所有项目将被永久删除，无法撤消。",
     "confirm": "永久删除"
+  },
+  "save_to_list_modal": {
+    "title": "保存到列表",
+    "items_count": "{count} {count, plural, one {项} other {项目}}",
+    "create_list": "创建列表"
```
kevinszuchet, line 1421: Some of the translations in the `en.js` file are missing here. Would you mind adding them?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+        } else {
+          setPicks({
+            ...picks,
+            pickFor: picks.pickFor.concat(lists.data[index])
+          })
+        }
+      }
+    },
+    [lists.data, picks]
+  )
```
kevinszuchet, line 127: Do you think we can write a function `pickOrUnpickFrom` (maybe) that receives the index, the key of the state to update, and an array (`unpickFrom` or `pickFor`) and reuse it in both cases of the if-else? The logic is the same but for different properties of the `picks` object.

webapp/src/modules/analytics/track.ts
```diff
@@ -389,3 +391,13 @@ track<UndoUnpickingItemAsFavoriteFailureAction>(
...
+
+track<BulkPickUnpickSuccessAction>(
+  BULK_PICK_SUCCESS,
+  events.BULK_PICK,
+  ({ payload: { item, pickedFor, unpickedFrom } }) => ({
+    item,
+    pickedFor,
+    unpickedFrom
+  })
+)
```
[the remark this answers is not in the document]
kevinszuchet: What do you think about "unwinding" the lists to emit an event for each one? Would that be possible?

webapp/src/modules/favorites/selectors.ts
```diff
@@ -2,20 +2,23 @@ import { createMatchSelector } from 'connected-react-router'
...
   UNDO_UNPICKING_ITEM_AS_FAVORITE_REQUEST,
   FETCH_LISTS_REQUEST,
   CREATE_LIST_REQUEST,
-  UPDATE_LIST_REQUEST,
-  DELETE_LIST_REQUEST
+  DELETE_LIST_REQUEST,
+  BULK_PICK_REQUEST,
+  UPDATE_LIST_REQUEST
 } from './actions'
+import { DEFAULT_FAVORITES_LIST_ID } from '../vendor/decentraland/favorites'
```
kevinszuchet, line 21: Would you mind moving this up?

webapp/src/modules/toast/toasts.tsx
```diff
@@ -281,3 +283,125 @@ export function getDeleteListFailureToast(list: List): Omit<Toast, 'id'> {
...
+      <div className="list-flow-toast">
+        <p>
+          {pickedFor.length > 0
+            ? buildBulkPickItemBodyMessage(
+                BulkPickUnpickMessageType.ADD,
+                BulkPickUnpickSuccessOrFailureType.SUCCESS,
+                item,
+                pickedFor
+              )
+            : undefined}{' '}
```
kevinszuchet, line 347: Tiny tiny comment. The space after the message should be displayed only if the length of the `pickedFor` is gt 0, right?

webapp/src/index.tsx
```diff
@@ -23,16 +23,16 @@ async function main() {
...
               <ModalProvider components={modals}>
                 <ScrollToTop />
                 <Routes />
               </ModalProvider>
-            </ConnectedRouter>
-          </WalletProvider>
-        </ToastProvider>
+            </ToastProvider>
+          </ConnectedRouter>
+        </WalletProvider>
```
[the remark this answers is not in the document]
LautaroPetaccio: It was needed to be able to use the `Link` component in the toasts.

webapp/src/modules/toast/toasts.tsx
```diff
@@ -281,3 +283,125 @@ export function getDeleteListFailureToast(list: List): Omit<Toast, 'id'> {
     icon: <Icon size="large" name="exclamation circle" />
   }
 }
+
+enum BulkPickUnpickMessageType {
```
flobarreto, line 287: what about moving this to types file? same to the one below

webapp/src/components/Modals/SaveToListModal/SaveToListModal.module.css
LautaroPetaccio: Done!

webapp/src/components/Modals/SaveToListModal/SaveToListModal.module.css
LautaroPetaccio: Done!

webapp/src/components/Modals/SaveToListModal/SaveToListModal.spec.tsx
LautaroPetaccio: Moved them up.

webapp/src/modules/translation/locales/zh.json
LautaroPetaccio: Great catch, I completely forgot to add them.

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
LautaroPetaccio: Moved them to a constants file.

webapp/src/modules/favorites/selectors.ts
LautaroPetaccio: Moved it!

webapp/src/modules/toast/toasts.tsx
LautaroPetaccio: You're right! I've changed it to show the whitespace only when both texts are used.

webapp/src/modules/toast/toasts.tsx
LautaroPetaccio: Moved them to the types file!

webapp/src/modules/analytics/track.ts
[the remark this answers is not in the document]
LautaroPetaccio: Added the tracking for the failure of the operation.
I theory it would be possible, but we would need to create a custom middleware or modify the existing one. The [current mechanism](https://github.com/decentraland/decentraland-dapps/blob/c711dfe7fdd59c33481442ffbbb79887a708ebb4/src/modules/analytics/utils.ts#L27) allows one event and payload per action.

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
LautaroPetaccio: Done!

webapp/src/components/Modals/SaveToListModal/SaveToListModal.module.css
```diff
@@ -0,0 +1,76 @@
...
+
+.loading span {
+  margin-left: 20px;
+}
+
+.actions {
+  display: flex;
+  flex-direction: column-reverse;
+  gap: 10px;
+}
```
kevinszuchet, line 76: A tiny one 🙏 : missing the EOF here.

webapp/src/modules/analytics/track.ts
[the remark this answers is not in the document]
kevinszuchet: I'll create a task to work on that later.

kevinszuchet (dismissed):
Wow 🐶

webapp/src/components/Modals/SaveToListModal/SaveToListModal.spec.tsx
```diff
@@ -0,0 +1,258 @@
+import { Item } from '@dcl/schemas'
+import { act } from 'react-dom/test-utils'
+import { fireEvent, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
```
kevinszuchet, line 4: This import is not being used, would you mind removing it?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.container.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const mapDispatch = (
+  dispatch: Dispatch,
+  ownProps: OwnProps
+): MapDispatchProps =>
+  bindActionCreators(
+    {
+      onSavePicks: (picksFor: List[], unpickFrom: List[]) =>
+        bulkPickUnpickRequest(ownProps.metadata.item, picksFor, unpickFrom),
+      onCreateList: () => undefined
```
kevinszuchet, line 31: Would you mind adding the real behavior here?

webapp/src/modules/toast/toasts.tsx
```diff
@@ -298,3 +304,117 @@ export function getDeleteListFailureToast(list: List): Omit<Toast, 'id'> {
...
+  const count = lists.length === 1 ? 'one' : lists.length === 2 ? 'two' : 'many'
+  return (
+    <T
+      id={`toast.bulk_pick.${successOrFailure}.body.${addOrRemove}.${count}`}
+      values={{
+        item_name: <b>{getAssetName(item)}</b>,
+        count: lists.length,
+        first_list_name: (
+          <b>
+            <Link to={locations.list(lists[0]?.id ?? '')}>
```
kevinszuchet, line 323: Would you mind adding the default browse options so the items will be displayed correctly in the list detail page?

webapp/src/modules/toast/toasts.tsx
```diff
@@ -298,3 +304,117 @@ export function getDeleteListFailureToast(list: List): Omit<Toast, 'id'> {
...
+              {lists[0]?.name ?? ''}
+            </Link>
+          </b>
+        ),
+        second_list_name: (
+          <b>
+            <Link to={locations.list(lists[1]?.id ?? '')}>
```
kevinszuchet, line 330: Same here about the default params.

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+      const hasNextPage = lists.data.length < lists.total
+      return !hasNextPage || index < lists.data.length
+    },
+    [lists]
+  )
+
+  const Row = useCallback(
+    ({ index, style }: { index: number; style: object }) => {
+      const isPicked =
+        (lists.data[index].isItemInList &&
```
kevinszuchet, line 182: Sometimes when you scroll fast, this throws an error because the `lists[index]` is not defined. Would you tal?

kevinszuchet:
I found an issue when unpicking an item that belongs to the list which I am seeing. IMO it's related to when you unpick an item, but it is still saved in another list.

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+    }
+  }, [picks.pickFor, picks.unpickFrom])
+
+  const handleSavePicks = useCallback(() => {
+    onSavePicks(picks.pickFor, picks.unpickFrom)
+  }, [onSavePicks, picks.pickFor, picks.unpickFrom])
+
+  const handleClose = useCallback(
+    () => (!isLoadingLists ? onClose : undefined),
+    [isLoadingLists, onClose]
```
kevinszuchet, line 96: The logic seems ok, but it's not possible to close the modal. Would you tal 🙏 ?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
LautaroPetaccio: Wow, I totally missed it, great catch!

webapp/src/modules/toast/toasts.tsx
LautaroPetaccio: Done, great catch!

webapp/src/modules/toast/toasts.tsx
LautaroPetaccio: Done, great catch!

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
LautaroPetaccio: Done!

kevinszuchet (approved):

PR #1786, opened 2023-06-14
fix: Empty list wording for different views

kevinszuchet:


webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -211,13 +213,15 @@ const ListPage = ({
...
-                    </Button>
-                  </div>
+                  <h1>{t(`list_page.empty.${privacyView}.title`)}</h1>
+                  <p>{t(`list_page.empty.${privacyView}.subtitle`)}</p>
+                  {!isPublicView && (
+                    <div className={styles.emptyActions}>
+                      <Button primary as={Link} to={locations.browse()}>
+                        {t(`list_page.empty.${privacyView}.action`)}
+                      </Button>
+                    </div>
```
LautaroPetaccio, line 223: Would you mind adding a test to assert that the action button exists or not depending on if it's public or not?

LautaroPetaccio (commented):
The PR looks great! I've just added a small comment on a test that could be useful

LautaroPetaccio (approved):
🍟

PR #1797, opened 2023-06-15
fix: Avoid loading lists before the ListsPage header

### LINKED ISSUE #1788: After going back from the detail to the lists page, the header with the filters and the create list button takes longer than the cards.
https://www.loom.com/share/839929d2d8bc4efc99fb1ae7e94d6762?sid=598b3655-d93e-4ce9-83cb-f4f23624cd2c

kevinszuchet:
Closes #1788

### DIFF
--- a/webapp/src/components/ListsPage/ListsPage.tsx
+++ b/webapp/src/components/ListsPage/ListsPage.tsx
@@ -77,10 +77,10 @@ const ListsPage = ({
       <Header className={styles.header} size="large">
         {t('lists_page.title')}
       </Header>
-      {!error && !isLoading && (
+      {!error && (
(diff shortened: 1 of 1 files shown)


webapp/src/components/ListsPage/ListsPage.tsx
```diff
@@ -121,7 +121,7 @@ const ListsPage = ({
           </div>
         </div>
       )}
-      {isLoading && lists.length === 0 ? (
+      {isLoading ? (
```
LautaroPetaccio, line 124: If we remove the condition of the lists length being zero, wouldn't be hiding the current items when loading more?

LautaroPetaccio (approved):
🚀

PR #1803, opened 2023-06-15
fix: Pagination not storing the current page

LautaroPetaccio:
This PR fixes the lists pagination that was caused because of the store not saving the current pagination.


kevinszuchet:
It seems ok. I saw a visual issue while sorting the lists. The pagination works great.

image

flobarreto (approved):

PR #1806, opened 2023-06-16
fix: create list modal button styles

### LINKED ISSUE #1802: Unespaced buttons in the CreateOrEditListModal
## 🎉 Description

The action buttons don't have any kind of separation between them.

flobarreto:
image

closes #1802


kevinszuchet (commented):
Nice, would you mind changing their order? We should also check what would happen in mobile.

PR #1825, opened 2023-06-19
fix: tab on my lists not selected on public view

flobarreto:


webapp/src/components/ListPage/ListPage.tsx
```diff
@@ -145,7 +145,9 @@ const ListPage = ({
   }
 
   return (
-    <PageLayout activeTab={NavigationTab.MY_LISTS}>
+    <PageLayout
+      activeTab={isPublicView ? NavigationTab.BROWSE : NavigationTab.MY_LISTS}
```
kevinszuchet, line 149: Why the `BROWSE` tab, wdyt about not focusing any of the tabs in the navigation bar?

kevinszuchet (commented):
Touche 🎩

webapp/src/components/ListPage/ListPage.tsx
flobarreto: we cannot focus on any on the tabs using page layout, as browse has not a tab I've chosen it

kevinszuchet: I think we can make the `PageLayout` component accept an optional `activeTab` as the `Navigation` component does.

Instead of the `NavigationTab.BROWSE`, we could send `undefined` and see the same result. WDYT?

PR #1863, opened 2023-06-21
fix: Redirect to list after creating one

### LINKED ISSUE #1862: Create a new list from the lists page doesn't redirect to the newly created list
## 🎉 Description

When creating a new list, the user is not redirected to the created list. Fix this by making the creation of lists redirect the user to the page of the created list.

LautaroPetaccio:
Closes #1862

### DIFF
--- a/webapp/src/modules/favorites/sagas.ts
+++ b/webapp/src/modules/favorites/sagas.ts
@@ -404,6 +404,7 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
   function* handleCreateListRequest(action: CreateListRequestAction) {
     const { name, isPrivate, description } = action.payload
     try {
+      const { pathname } = yield select(getLocation)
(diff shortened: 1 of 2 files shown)


kevinszuchet (approved):
Touche 🖨️

webapp/src/modules/favorites/sagas.ts
```diff
@@ -414,6 +415,9 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
         description
       })
       yield put(createListSuccess(list))
+      if (pathname === locations.lists()) {
+        yield put(push(locations.list(list.id)))
+      }
```
kevinszuchet, line 420: With this, we are losing the toast that appeared after creating a new list. How could we keep both mechanisms?

kevinszuchet: My bad, the toast never appeared before

PR #1878, opened 2023-06-22
fix: make cards enable actions

### LINKED ISSUE #1850: The card in the ShareListModal must be view only
## 🎉 Description

The card in the ShareListModal has hover and the options menu which allows the user to edit or delete the list. This should not be the case.
To complete this task, change the `ListCard` to have a `readOnly` prop which will remove the hover animation and the card's options (three dotted buttons).

flobarreto:
closes #1850


kevinszuchet (commented):
We could call the new prop `viewOnly` and go for an optimistic comparison, but it is up to you.

Would you add some tests asserting that the dropdown exists or not? 🙏

kevinszuchet (approved):
❤️

PR #1879, opened 2023-06-22
fix: Avoid calling the catalog without trending items

kevinszuchet:


webapp/src/modules/item/sagas.ts
```diff
@@ -121,6 +121,12 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
         size
       )
       const ids = data.map(item => item.id)
+
+      if (!ids.length) {
```
cyaiox, line 125: could it be items without id? 

wdyt about validating with `data.length`?

PR #1880, opened 2023-06-22
fix: End bulk sagas when closing the modal or failing the creation of a list

### LINKED ISSUE #1868: When removing an item from a list and adding another one to a new list, the new one contains all the items from the state
## 🎉 Description

https://images.zenhubusercontent.com/325377244/b1d8037e-a421-48d2-b058-2767a1af754d/screen_recording_2023_06_21_at_120319_pxzjesvu.mp4

LautaroPetaccio:
This PR changes the actions that were being raced in the `handleBulkPickStart` saga.
The race now handles the `BULK_PICK_REQUEST` action that would occur when the users picked some lists without creating one, ending the saga. Without handling this action, the saga would hang and, whenever any of the other actions occurred, it would continue all of them.

Closes #1868

### DIFF
--- a/webapp/src/modules/favorites/sagas.ts
+++ b/webapp/src/modules/favorites/sagas.ts
@@ -459,10 +459,12 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
         listCreationSuccess
       }: {
         listCreationSuccess: CreateListSuccessAction
+        picksInBulkRequest: BulkPickUnpickRequestAction
         modalClosed: CloseModalAction
       } = yield race({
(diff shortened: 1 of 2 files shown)


webapp/src/modules/favorites/sagas.ts
```diff
@@ -484,6 +481,7 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
 
         yield put(bulkPickUnpickRequest(item, [pickedList], []))
       }
+      console.log('end sagas')
```
kevinszuchet, line 484: Would you remove it?

flobarreto (approved):

PR #1882, opened 2023-06-22
fix: remove hover transition on viewonly cards

flobarreto:


webapp/src/components/ListsPage/ListCard/ListCard.tsx
```diff
@@ -43,7 +43,7 @@ const ListCard = (props: Props) => {
         vendor: VendorName.DECENTRALAND,
         sortBy: SortBy.NEWEST
       })}
-      className={styles.card}
+      className={`${styles.ListCard} ${isViewOnly ? styles.viewOnly : ''}`}
```
LautaroPetaccio, line 46: Would you mind using `classnames` here?

LautaroPetaccio (commented):
The PR looks great! It would be great if we could use the dependency I've marked in a comment.

LautaroPetaccio (approved):

PR #1883, opened 2023-06-23
feat: Use pagination with collections

LautaroPetaccio:
This PR makes the `CollectionList` component work with the `usePagination` hook, removing its code from the browse function.


flobarreto (dismissed):

webapp/src/components/CollectionList/CollectionList.container.ts
```diff
@@ -23,14 +21,15 @@ import {
...
-  sortBy: getSortBy(state),
-  page: getPageNumber(state)
+  isLoading: isLoadingType(getLoading(state), FETCH_COLLECTIONS_REQUEST)
 })
 
-const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onBrowse: options => dispatch(browse(options))
-})
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps =>
+  bindActionCreators(
```
cyaiox, line 28: sq: what is the benefit of using `bindActionCreators` here?

webapp/src/lib/pagination.ts
```diff
@@ -64,23 +81,38 @@ export function usePagination(
...
+      }
       params.set(filter, value)
       push(`${pathname}?${params.toString()}`)
     },
-    [pathname, push, search]
+    [pathname, push, search, sortBy]
   )
 
+  const pages = options?.count
+    ? Math.ceil(options?.count / (options?.pageSize ?? PAGE_SIZE))
```
cyaiox, line 102: could `options?.count` be undefined?

webapp/src/components/AccountSidebar/CurrentAccountSidebar/CurrentAccountSidebar.tsx
```diff
@@ -26,70 +27,76 @@ const {
...
+  return (
+    <>
+      <Menu>
+        <Header sub>{t('account_sidebar.assets')}</Header>
+        <MenuItem
+          key={COLLECTIONS}
+          value={COLLECTIONS}
+          currentValue={section}
+          onClick={section =>
+            changeFilter('section', section, { clearOldFilters: true })
```
cyaiox, line 42: should this `'section'` string be an enum in `lib/pagination`:

```
export enum FILTERS {
  SECTION = 'section',
  FILTER = 'filter'
}
```

cyaiox (dismissed):
🚀 🚀 🚀

webapp/src/components/CollectionList/CollectionList.container.ts
LautaroPetaccio: The benefit is to replace the boilerplate where we define the same parameters of the action and wrap it in the dispatch. It's a great addition to the repository!

webapp/src/components/AccountSidebar/CurrentAccountSidebar/CurrentAccountSidebar.tsx
LautaroPetaccio: I think that all filters available should be set either by the module that is called upon changing the filter or by the API client that makes the request to the server as the idea is to group them by resource. The problem with the section filter is that is used as a "browse option", that is, filters in the browse operation are not handled as enums.

cyaiox (approved):

PR #1896, opened 2023-06-29
chore: Remove lists v1 ff

### LINKED ISSUE #1892: Remove Lists feature flags
## 🎉 Description

This PR removes the lists feature flags and all the feature flagged code.

kevinszuchet:
Closes #1892


webapp/src/components/FavoritesCounter/FavoritesCounter.spec.tsx
```diff
@@ -146,58 +140,6 @@ describe('FavoritesCounter', () => {
     })
   })
 
-  describe('when the user clicks the component', () => {
-    describe('and the item is unpicked', () => {
```
LautaroPetaccio, line 150: If I'm not wrong, we're removing all the tests that exercise that the onClick is being called. What do you think of keeping two, the last one (testing that the onClick was not called twice) and another one to test that the onClick function was called when clicked.

LautaroPetaccio (commented):
The PR looks great! 👍  Just a minor comment on a test

webapp/src/components/FavoritesCounter/FavoritesCounter.spec.tsx
kevinszuchet: Nice catch, thanks!

LautaroPetaccio (approved):
The best PRs are the ones that remove code 💣

PR #1901, opened 2023-07-03
feat: make create list modal adaptable when opening it inside save list modal

### LINKED ISSUE #1871: When removing an item from inside a list, we might want to skip the list selector and assume users want to remove the item from the specific list they are triggering the action.

flobarreto:
closes #1871

### DIFF
--- a/webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
+++ b/webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
@@ -19,6 +19,7 @@ import {
 } from '../../../modules/vendor/decentraland/favorites'
 import { retryParams } from '../../../modules/vendor/decentraland/utils'
 import { CreateListParameters } from '../../../modules/favorites/types'
+import * as events from '../../../utils/events'
 import { isErrorWithMessage } from '../../../lib/error'
 import { PrivateTag } from '../../PrivateTag'
 import {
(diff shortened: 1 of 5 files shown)


webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -298,7 +335,10 @@ const SaveToListModal = (props: Props) => {
           secondary
           disabled={isLoadingLists || isSavingPicks}
           data-testid={CREATE_LIST_BUTTON_DATA_TEST_ID}
-          onClick={onCreateList}
+          onClick={handleOnCreateListClick}
+          // (name: string, description: string, isPrivate: boolean) =>
+          // createList(name, description, isPrivate)
+          // }
```
LautaroPetaccio, line 341: Would you mind removing these comments?

webapp/src/components/Modals/SaveToListModal/SaveToListModal.tsx
```diff
@@ -162,6 +164,41 @@ const SaveToListModal = (props: Props) => {
...
+          stateLists.splice(1, 0, {
+            ...response,
+            itemsCount: 0,
+            previewOfItemIds: []
+          })
+          setLists({
+            total: lists.total++,
+            data: stateLists
+          })
+          onFinishListCreation()
```
LautaroPetaccio, line 183: I haven't realized, but we need to replicate the tracking we've in the track module for the list creation and failure here. We need to do that because we're not using actions.

LautaroPetaccio (changes requested):
Great work 👍  We need to replicate the tracking actions and we'll be ready to go!

LautaroPetaccio (approved):

PR #1914, opened 2023-07-07
feat: add sepolia support

lauti7:
This PR: 
- add support for Sepolia

### DIFF
--- a/webapp/src/config/env/dev.json
+++ b/webapp/src/config/env/dev.json
@@ -1,6 +1,6 @@
 {
-  "NETWORK": "sepolia",
-  "CHAIN_ID": "11155111",
+  "NETWORK": "goerli",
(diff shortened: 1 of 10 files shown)


webapp/src/config/env/dev.json
```diff
@@ -1,6 +1,6 @@
 {
-  "NETWORK": "goerli",
-  "CHAIN_ID": "5",
+  "NETWORK": "sepolia",
```
fzavalia, line 2: I think it would be better to make this change on a different PR so we can merge this one and prevent any conflicts while other things are developed

fzavalia (changes requested):

fzavalia (dismissed):

fzavalia (approved):

PR #1915, opened 2023-07-10
fix: Remove lists ftu from mobile

kevinszuchet:


webapp/src/components/Modals/ListsLaunchModal/ListsLaunchModal.tsx
```diff
@@ -28,14 +33,16 @@ export const ListsLaunchModal = ({
     }
   }, [isLoadingFeatureFlags])
 
+  const isMobile = useMobileMediaQuery()
```
juanmahidalgo, line 36: How does it look in tables? Do we want to remove it there as well? 
If so, we can use `const isMobileOrTable = useTabletAndBelowMediaQuery()` here.

juanmahidalgo (dismissed):
approving anyways if tablet is fine.

juanmahidalgo (approved):

PR #1918, opened 2023-07-10
feat: replace goerli with sepolia

lauti7:
This shouldn't be merged until the whole migration is done


webapp/src/config/env/dev.json
fzavalia, line 1: Change the peer url to the one in sepolia

fzavalia (changes requested):

fzavalia (approved):

PR #1936, opened 2023-07-17
fix: Create orders in seconds

LautaroPetaccio:
This PR fixes an issue where orders were created in milliseconds instead of seconds.


webapp/src/modules/order/actions.ts
```diff
@@ -20,6 +20,7 @@ export const CREATE_ORDER_FAILURE = '[Failure] Create Order'
 export const createOrderRequest = (
   nft: NFT,
   price: number,
+  /* The expiration date in seconds */
```
fzavalia, line 23: The comments says it is in seconds but the conversion to seconds is being made in the OrderService https://github.com/decentraland/marketplace/pull/1936/files#diff-25ef054ef50c658b3d517b843c58008ebdb4d9601e94852bc080a4961e9e5f0cR44. Maybe just remove the comment to avoid confusion.

fzavalia (changes requested):

fzavalia (approved):

PR #1941, opened 2023-07-20
feat: hide the legacy order from the Item and NFT detail pages

juanmahidalgo:


webapp/src/components/AssetPage/BuyNFTBox/BuyNFTBox.tsx
```diff
@@ -97,54 +99,69 @@ const BuyNFTBox = ({ nft, address }: Props) => {
...
-                >
-                  {formatWeiToAssetCard(listing.order.price)}
-                </Mana>
-              </div>
-              {+listing.order.price > 0 && (
-                <div className={styles.informationText}>
-                  {'('}
-                  <ManaToFiat mana={listing.order.price} />
-                  {')'}
+        {!isOrderExpired || (isLegacyOrder(listing.order) && isOwner) ? (
```
juanmahidalgo, line 106: If it's the owner, they should see that the order is active somehow and see the action buttons

PR #1962, opened 2023-08-03
feat: Upgrade decentraland ui and dapps packages

cyaiox:
This PR upgrades the decentraland ui and dapps packages.

Old Menu style:
image

New Menu style:
image


flobarreto (approved):

webapp/src/components/LinkedProfile/LinkedProfile.tsx
```diff
@@ -9,7 +9,7 @@ export const LinkedProfile = ({ isProfileEnabled, ...props }: Props) => {
   const redirectionProps: RedirectionProps = isProfileEnabled
     ? {
         as: 'a',
-        href: `${profileUrl}/${address}`
+        href: `${profileUrl}${locations.account(address)}`
```
LautaroPetaccio, line 12:
```suggestion
        href: `${profileUrl}/accounts/${address}`
```
What do you think of hardcoding the path instead of using locations? I'm a bit concerned about the location function being for locations inside of the Marketplace. If we eventually change or delete the account path in the Marketplace, we'll be be changing the path to the profile url.

cyaiox: good catch! let me update it

LautaroPetaccio (approved):

PR #1963, opened 2023-08-04
feat: Single Sign On

fzavalia:


webapp/package.json
```diff
@@ -113,5 +114,10 @@
       "last 1 safari version"
     ]
   },
-  "homepage": ""
+  "homepage": "",
+  "jest": {
+    "moduleNameMapper": {
+      "@dcl/single-sign-on-client": "identity-obj-proxy"
```
fzavalia, line 120: For tests it replaces this module with this stub so they don't break. That is because the sso client is an ESM and this old dapps don't support them

webapp/config-overrides.js
```diff
@@ -6,6 +6,7 @@ module.exports = override(
     path.resolve(__dirname, 'src'),
     path.resolve(__dirname, 'node_modules/react-virtualized-auto-sizer'),
     path.resolve(__dirname, 'node_modules/decentraland-connect/node_modules/@walletconnect'),
-    path.resolve(__dirname, 'node_modules/@walletconnect')
+    path.resolve(__dirname, 'node_modules/@walletconnect'),
+    path.resolve(__dirname, 'node_modules/@dcl/single-sign-on-client'),
```
fzavalia, line 10: Package has some unsupported operators by the bundler version used here like `??`. So it has to be added in the overrides.

webapp/package.json
fzavalia, line 120: As the module is an ESM and not CommonJs, jest doesn't like it. Doing this will replace it in tests with a mockable stub.

webapp/src/modules/store.ts
```diff
@@ -49,13 +48,11 @@ export function initStore() {
     paths: [
       ['ui', 'archivedBidIds'],
       ['ui', 'preview', 'isTryingOn'],
-      ['identity', 'data']
```
fzavalia, line 52: By removing this, the identity is not persisted in local storage anymore. It is now persisted in the iframe so there is no reason for it.

cyaiox (approved):

PR #1967, opened 2023-08-10
feat: introduce logic to fetch catalog from the marketplace-server

juanmahidalgo:
This PR enables the marketplace-server as the catalog provider based on its feature flag status.


webapp/src/config/env/dev.json
```diff
@@ -1,42 +1,43 @@
 {
-  "NETWORK": "sepolia",
-  "CHAIN_ID": "11155111",
-  "NFT_SERVER_URL": "https://nft-api.decentraland.zone/v1",
-  "BUILDER_SERVER_URL": "https://builder-api.decentraland.zone/v1",
-  "ATLAS_SERVER_URL": "https://api.decentraland.zone",
-  "PEER_URL": "https://peer-ue-2.decentraland.zone",
-  "SIGNATURES_SERVER_URL": "https://signatures-api.decentraland.zone/v1",
-  "MARKETPLACE_FAVORITES_SERVER_URL": "https://marketplace-favorites-api.decentraland.zone/v1",
+  "NETWORK": "mainnet",
```
juanmahidalgo, line 2: this changes will be rollbacked before merging

webapp/src/modules/favorites/sagas.spec.ts
```diff
@@ -1027,13 +1086,14 @@ describe('when handling the request for getting a list', () => {
     })
 
     describe('and non of the preview items are already in the state', () => {
-      it('should dispatch an action signaling the success of the handled action with the received list and items', () => {
+      it('should dispatch an action signaling the success of the handled action with the recieved list and items', () => {
```
kevinszuchet, line 1089:
```suggestion
      it('should dispatch an action signaling the success of the handled action with the received list and items', () => {
```

webapp/src/modules/features/selectors.ts
```diff
@@ -195,3 +195,14 @@ export const getIsSmartWearablesFTUEnabled = (state: RootState) => {
...
+export const getIsMarketplaceServerEnabled = (state: RootState) => {
+  if (hasLoadedInitialFlags(state)) {
+    return getIsFeatureEnabled(
+      state,
+      ApplicationName.MARKETPLACE,
+      FeatureName.MARKETPLACE_SERVER
+    )
+  }
+  return false
+}
```
kevinszuchet, line 208: Missing EOF here

webapp/src/modules/favorites/sagas.ts
```diff
@@ -173,9 +181,16 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
         filters: optionsFilters
       }
 
+      const isMarketplaceServerEnabled: boolean = yield select(
+        getIsMarketplaceServerEnabled
+      )
+      const api = isMarketplaceServerEnabled
+        ? marketplaceServerCatalogAPI
+        : catalogAPI
```
kevinszuchet, line 189: WDYT about moving the `select` and the `ternary operator` into a new function called `getCatalogAPI` or `getItemsAPI` (something like that)?

webapp/src/modules/features/utils.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  FETCH_APPLICATION_FEATURES_FAILURE,
+  FETCH_APPLICATION_FEATURES_SUCCESS
+} from 'decentraland-dapps/dist/modules/features/actions'
+import { isLoadingFeatureFlags } from 'decentraland-dapps/dist/modules/features/selectors'
+import { race, select, take } from 'redux-saga/effects'
```
kevinszuchet, line 6: Would you mind moving this up 🙏 😅 ?

webapp/src/modules/features/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function* waitForFeatureFlagsToBeLoaded() {
+  const isFetchingFeatureFlags: boolean = yield select(isLoadingFeatureFlags)
+  if (isFetchingFeatureFlags) {
+    yield race({
+      success: take(FETCH_APPLICATION_FEATURES_SUCCESS),
+      failure: take(FETCH_APPLICATION_FEATURES_FAILURE)
+    })
+  }
+}
```
kevinszuchet, line 16: Should we test this new function alone? It's already tested in the item sagas test file.
It's up to you.

webapp/src/modules/item/sagas.ts
```diff
@@ -172,7 +185,14 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
...
 
     try {
-      const api = isCatalogView(view) ? catalogAPI : itemAPI
+      yield call(waitForFeatureFlagsToBeLoaded)
+      const isMarketplaceServerEnabled: boolean = yield select(
+        getIsMarketplaceServerEnabled
+      )
+      const catalogViewAPI = isMarketplaceServerEnabled
+        ? marketplaceServerCatalogAPI
+        : catalogAPI
```
kevinszuchet, line 194: Here we can use the same function I suggested before (also in the line 143 where we also need this logic).

webapp/src/modules/favorites/sagas.ts
juanmahidalgo: I think it's a complicated abstraction because the selector will need both API instances and they are created within the saga, so it's even more complicated to test since the matches won't have access to the instances that the saga creates. Do you have a way to solve that? 
E.g:

Selector:
`const api: CatalogAPI = yield select(
        getCatalogAPI,
        catalogAPI,
        marketplaceServerCatalogAPI
      )`


test that won't work:

```
[
                    select(
                      getCatalogAPI,
                      new CatalogAPI('marketplaceAPI'),
                      new CatalogAPI('nftAPI')
                    ),
                    isMarketplaceFFOn
                  ],
```                  
since the catalogAPI instances won't match

kevinszuchet: I meant moving the select and the if condition inside a new function, not necessarily a selector.
We could have a function inside the context of the saga, so this new function won't need to receive the APIs in the parameters but will be able to use it from the context.

Something like:
```ts
function* getCatalogAPI() {
  const isMarketplaceServerEnabled: boolean = yield select(
    getIsMarketplaceServerEnabled
  );
  return isMarketplaceServerEnabled
    ? marketplaceServerCatalogAPI
    : catalogAPI;
}
```

webapp/src/modules/features/utils.ts
juanmahidalgo: done!

webapp/src/modules/favorites/sagas.ts
juanmahidalgo: oh ok, that makes sense, updated!

webapp/src/modules/features/selectors.ts
juanmahidalgo: added!

webapp/src/modules/favorites/sagas.ts
```diff
@@ -132,12 +139,7 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
           ids: previewListsItemIds
         }
 
-        const isMarketplaceServerEnabled: boolean = yield select(
-          getIsMarketplaceServerEnabled
-        )
-        const api = isMarketplaceServerEnabled
-          ? marketplaceServerCatalogAPI
-          : catalogAPI
+        const api: CatalogAPI = yield getCatalogAPI()
```
kevinszuchet, line 142:
```suggestion
        const api: CatalogAPI = yield call(getCatalogAPI)
```
WDYY?

webapp/src/modules/favorites/sagas.ts
```diff
@@ -181,12 +183,7 @@ export function* favoritesSaga(getIdentity: () => AuthIdentity | undefined) {
         filters: optionsFilters
       }
 
-      const isMarketplaceServerEnabled: boolean = yield select(
-        getIsMarketplaceServerEnabled
-      )
-      const api = isMarketplaceServerEnabled
-        ? marketplaceServerCatalogAPI
-        : catalogAPI
+      const api: CatalogAPI = yield getCatalogAPI()
```
kevinszuchet, line 186: Same as previous comment 🙏

webapp/src/modules/features/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expectSaga } from 'redux-saga-test-plan'
+import { select } from 'redux-saga/effects'
+import { isLoadingFeatureFlags } from 'decentraland-dapps/dist/modules/features/selectors'
+import {
+  FETCH_APPLICATION_FEATURES_FAILURE,
+  FETCH_APPLICATION_FEATURES_SUCCESS
+} from 'decentraland-dapps/dist/modules/features/actions'
+import { waitForFeatureFlagsToBeLoaded } from './utils' // adjust the path
+
+describe('waitForFeatureFlagsToBeLoaded Saga', () => {
```
kevinszuchet, line 10:
```suggestion
describe('waitForFeatureFlagsToBeLoaded Util', () => {
```

webapp/src/modules/features/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expectSaga } from 'redux-saga-test-plan'
+import { select } from 'redux-saga/effects'
+import { isLoadingFeatureFlags } from 'decentraland-dapps/dist/modules/features/selectors'
+import {
+  FETCH_APPLICATION_FEATURES_FAILURE,
+  FETCH_APPLICATION_FEATURES_SUCCESS
+} from 'decentraland-dapps/dist/modules/features/actions'
+import { waitForFeatureFlagsToBeLoaded } from './utils' // adjust the path
```
kevinszuchet, line 8: What do you mean by the comment "adjust the path"?

kevinszuchet (approved):
Just minor comments. LGTM 🚀

webapp/src/modules/features/utils.spec.ts
juanmahidalgo: just copilot messing things up 😂

PR #1981, opened 2023-08-17
fix: Autofix eslint errors

cyaiox:
This PR auto-fixes eslint errors and prettier formats


webapp/.eslintrc.js
```diff
@@ -4,20 +4,5 @@ module.exports = {
   parserOptions: {
     project: ['tsconfig.json']
   },
-  overrides: [
```
cyaiox, line 7: update prettier config

PR #2005, opened 2023-09-07
fix: wait for identity success action before fetching catalog (#1974)

### LINKED ISSUE #2004: Fix race condition not getting the identity before fetching Catalog
## 🎉 Description

The catalog request is going unauthenticated at the moment due to a race condition when getting the identity. This was introduced after the Single Sign On changed how the identity was fetched.

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #2004

### DIFF
--- a/webapp/src/modules/item/sagas.ts
+++ b/webapp/src/modules/item/sagas.ts
@@ -192,8 +192,9 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
     // If the wallet is getting connected, wait until it finishes to fetch the items so it can fetch them with authentication
     yield call(waitForWalletConnectionIfConnecting)
     const wallet: Wallet | undefined = yield select(getWallet)
+    const identity: AuthIdentity | undefined = yield select(getIdentity)
 
-    if (wallet) {
+    if (wallet && !identity) {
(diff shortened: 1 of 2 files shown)


webapp/src/modules/item/sagas.ts
```diff
@@ -163,13 +168,34 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
...
     action: FetchItemsRequestAction
   ): SagaIterator {
     const { filters, view } = action.payload
 
     // If the wallet is getting connected, wait until it finishes to fetch the items so it can fetch them with authentication
     yield call(waitForWalletConnectionIfConnecting)
+    const wallet: Wallet | undefined = yield select(getWallet)
+
+    if (wallet) {
+      yield call(waitForIdentityToBeGenerated, action)
```
fzavalia, line 197: If the identity was already generated, wouldn't it get stuck waiting for a GENERATE_IDENTITY_SUCCESS that will never happen?

juanmahidalgo: absolutely correct, updated!

fzavalia (approved):

PR #2007, opened 2023-09-08
Feat/emote sound geomtry

meelrossi:
Add `with sound` and `with props` filter to marketplace for new emotes 2.0


webapp/src/components/AssetFilters/EmoteAttributesFilter/EmoteAttributesFilter.container.ts
```diff
@@ -0,0 +1,13 @@
+import { connect } from 'react-redux'
+import { RootState } from '../../../modules/reducer'
+import {
+  MapStateProps
+} from './EmoteAttributesFilter.types'
+import { EmoteAttributesFilter } from './EmoteAttributesFilter'
+import { getIsEmotesV2Enabled } from '../../../modules/features/selectors'
```
kevinszuchet, line 7: Would you mind sorting these imports?

webapp/src/components/AssetFilters/EmoteAttributesFilter/EmoteAttributesFilter.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const handlePlayModeChange = useCallback(
+    (values: string[]) =>
+      onChange({ emotePlayMode: values as EmotePlayMode[] }),
+    [onChange]
+  )
+
+  const handleEmoteAttributesChange = useCallback(
+    (values: string[]) =>
+      onChange({
+        emoteHasSound: values.includes(WITH_SOUND_VALUE) ? true : undefined,
```
kevinszuchet, line 62: Not mandatory. I think the `true` is redundant. this could be:

```suggestion
        emoteHasSound: values.includes(WITH_SOUND_VALUE) || undefined,
```

webapp/src/components/AssetFilters/EmoteAttributesFilter/EmoteAttributesFilter.tsx
```diff
@@ -0,0 +1,122 @@
...
+        emoteHasGeometry: values.includes(WITH_GEOMETRY_VALUE)
```
kevinszuchet, line 63: Same as before this could be
```ts
emoteHasGeometry: values.includes(WITH_GEOMETRY_VALUE) || undefined
```

kevinszuchet (approved):

PR #2043, opened 2023-10-27
fix: remove search dropdown on account views

meelrossi:
Remove search dropdown on account views to use simple search


juanmahidalgo (approved):
LGTM!!

PR #2044, opened 2023-10-27
fix: empty search catalog

meelrossi:
- Fix issue when clearing search input it would not reload the page with all items
- Fix double call to catalog endpoint when using search dropdown. It was being called once with the previous searchUUID and once with the correct searchUUID


juanmahidalgo (approved):
LGTM! thanks for the fix @meelrossi 👏

PR #2047, opened 2023-11-08
Fix/fingerprint calculation

meelrossi:
We are currently having a cache problem causing the information coming from the atlas-server to be outdated. For this, in several parts of the ui (atlas, estate detaill, etc) we are showing information that is not up to date misleading the user to what they are actually seeing/buying.

To solve some of the consequences of this, this PR adds:
- Fingerprint calculation for estates in the ui and compare it with the one generated with the contract. If this fingerprints are not the same then the information is not up to date and we prevent the user from buying the land/estate.

- Add warning messages in the atlas and in the buy and bid page to make the users aware of the missmatch

image
image


webapp/src/components/Atlas/Atlas.tsx
```diff
@@ -17,6 +17,7 @@ import { VendorName } from '../../modules/vendor'
 import { NFT } from '../../modules/nft/types'
 import Popup from './Popup'
 import './Atlas.css'
+import ErrorBanner from '../ErrorBanner'
```
fzavalia, line 20: Move up

webapp/src/components/BidPage/BidModal/BidModal.tsx
```diff
@@ -28,6 +28,7 @@ import { ConfirmInputValueModal } from '../../ConfirmInputValueModal'
 import { Mana } from '../../Mana'
 import { Props } from './BidModal.types'
 import './BidModal.css'
+import ErrorBanner from '../../ErrorBanner'
```
fzavalia, line 31: Move up

webapp/src/components/BuyPage/BuyNFTModal/BuyNFTModal.tsx
```diff
@@ -33,6 +33,7 @@ import { PartiallySupportedNetworkCard } from '../PartiallySupportedNetworkCard'
 import { NotEnoughMana } from '../NotEnoughMana'
 import { PriceHasChanged } from '../PriceHasChanged'
 import { Props } from './BuyNFTModal.types'
+import ErrorBanner from '../../ErrorBanner'
```
fzavalia, line 36: Move up

webapp/src/modules/nft/estate/utils.ts
```diff
@@ -21,6 +23,55 @@ export const getCenter = (selection: { x: number; y: number }[]) => {
...
+      }
+    ],
+    new ethers.providers.Web3Provider(provider)
+  )
+
+  const estateTokenIds = []
+
+  for (const parcel of parcels) {
+    estateTokenIds.push(
+      (await contract.encodeTokenId(parcel.x, parcel.y)).toString()
```
fzavalia, line 55: Is this toString necessary?
The land tokenId is then used in ethers.utils.solidityKeccak256 so I believe that big number should be handled without the need of stringifying.

fzavalia:
Should the message cover the 100% of the width?

Screenshot 2023-11-08 at 11 30 54

PR #2069, opened 2023-12-05
feat: add notifications to marketplace

lauti7:


package-lock.json
```diff
@@ -0,0 +1,12 @@
+{
```
meelrossi, line 1: i think this is not needed .. you probably run `npm i` in the base folder

webapp/src/components/Navbar/Navbar.container.ts
```diff
@@ -8,14 +8,16 @@ import { isConnected } from '../../modules/wallet/selectors'
...
 
 const mapState = (state: RootState): MapStateProps => ({
   isConnected: isConnected(state),
   location: getLocation(state),
   hasPendingTransactions: getTransactions(state).some(tx =>
     isPending(tx.status)
   ),
-  isNewNavbarDropdownEnabled: getIsNewNavbarDropdownEnabled(state)
+  isNewNavbarDropdownEnabled: getIsNewNavbarDropdownEnabled(state),
+  identity: getCurrentIdentity(state) || undefined
```
meelrossi, line 20: is the `|| undefined` needed? it returns null?

package-lock.json
lauti7: leak :(

meelrossi (approved):

PR #2087, opened 2024-01-05
feat: udpate decentraland dapps

meelrossi:
Update decentraland dapps with new translation
REmove UserMenu as it was removed from decentraland-ui


juanmahidalgo (approved):
LGTM!

PR #2099, opened 2024-01-11
fix: name taken link fix

juanmahidalgo:


webapp/src/components/NamesPage/ClaimNamePage/ClaimNamePage.tsx
```diff
@@ -41,6 +42,9 @@ import { Props } from './ClaimNamePage.types'
 import styles from './ClaimNamePage.module.css'
 import { PageLayout } from '../../PageLayout'
 
+export const MARKETPLACE_SERVER_URL = config.get('MARKETPLACE_SERVER_URL')!
```
fzavalia, line 45: Is the export necessary?

webapp/src/components/NamesPage/ClaimNamePage/ClaimNamePage.tsx
```diff
@@ -41,6 +42,9 @@ import { Props } from './ClaimNamePage.types'
 import styles from './ClaimNamePage.module.css'
 import { PageLayout } from '../../PageLayout'
```
fzavalia, line 43: Move import up

fzavalia (changes requested):
🦊

webapp/src/components/NamesPage/ClaimNamePage/ClaimNamePage.tsx
juanmahidalgo: it's not a change from this PR 😂 moved above anyways 👍

webapp/src/components/NamesPage/ClaimNamePage/ClaimNamePage.tsx
juanmahidalgo: no, and the variable it's not needed either 😂 removed (copy and pasted from my ongoing task and I didn't notice)

fzavalia (approved):

PR #2102, opened 2024-01-12
feat: Add the new FIAT Flow to the ClaimName feature

juanmahidalgo:
image

### FF not enabled

image

### FF enabled

image


webapp/src/modules/ens/sagas.ts
```diff
@@ -29,6 +34,56 @@ export const REGISTRAR_ADDRESS = config.get('REGISTRAR_CONTRACT_ADDRESS', '')
...
+  ) {
+    const data = action.payload[TRANSACTION_ACTION_FLAG]
+    const {
+      hash,
+      payload: { subdomain, address }
+    } = data
+
+    const from = address
+
+    try {
```
LautaroPetaccio, line 53: Would you mind adding tests for this new handler?

LautaroPetaccio (commented):
Nice work! It looks great. I've left some minor comments.
There's an issue that I've encountered wile trying the FIAT option:
```
payload:  {error: 'No identity found'}
type:  "[Failure] Open FIAT Gateway Widget"
```
There might be an issue with the retrieval of the identity in the `decentraland-dapps` repo.

webapp/src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.tsx
```diff
@@ -17,56 +36,189 @@ export const CONTROLLER_V2_ADDRESS = config.get(
...
+                options.tx_id as string
+              )
+            }
+          }
+        }
+      )
     }
+  }, [wallet, ENSName, analytics, onOpenFiatGateway, onClaimTxSubmitted])
+
+  const isLoading = useMemo(() => isClaiming || isLoadingFIATWidget, [])
```
LautaroPetaccio, line 152: If I'm not wrong, as there's no dependencies in this `useMemo`, `isLoading` will be computed only once. As this is a boolean comparison, what do you think about removing the `useMemo` hook?
```suggestion
  const isLoading = isClaiming || isLoadingFIATWidget
```

webapp/src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.spec.tsx
```diff
@@ -57,29 +85,108 @@ describe('ClaimNameFatFingerModal', () => {
...
+        await fireEvent.change(inputField, { target: { value: name } })
+        const claimButton = getByText(t('global.confirm'))
+        await waitFor(() => {
+          expect(claimButton).not.toBeDisabled()
+        })
+        fireEvent.click(claimButton)
+
+        await waitFor(() => {
+          expect(baseProps.onOpenFiatGateway).toHaveBeenCalled()
+        })
```
LautaroPetaccio, line 188: I think the test is agnostic on if the user has or not MANA, right? Would you mind moving it outside the `and does not have enough MANA balance to pay`context?
```suggestion
  describe('when clicking the claim button', () => {
      it('should open FIAT gateway widget', async () => {
        const { getByRole, getByText } = renderWithProviders(
          <ClaimNameFatFingerModal
            {...baseProps}
            currentMana={currentMana}
            getContract={jest.fn().mockResolvedValue({
              address: '0x0' // mana contract mock
            })}
            onAuthorizedAction={jest
              .fn()
              .mockImplementation(({ onAuthorized }: any) => {
                onAuthorized()
              })}
          />
        )

        const inputField = getByRole('textbox')
        await fireEvent.change(inputField, { target: { value: name } })
        const claimButton = getByText(t('global.confirm'))
        await waitFor(() => {
          expect(claimButton).not.toBeDisabled()
        })
        fireEvent.click(claimButton)

        await waitFor(() => {
          expect(baseProps.onOpenFiatGateway).toHaveBeenCalled()
        })
```

LautaroPetaccio (approved):

LautaroPetaccio (approved):
🚀

PR #2105, opened 2024-01-18
feat: Use vite to build site

LautaroPetaccio:
This PR changes the way we're building the Marketplace to use Vite.


webapp/src/modules/vendor/decentraland/index.ts
```diff
@@ -20,13 +20,6 @@ export * from './BidService'
 export * from './ContractService'
```
LautaroPetaccio, line 20: I'm removing the services export as we're not longer using the vendors idea and we have a cyclic dependency later when instantiating them.

webapp/src/modules/features/selectors.ts
```diff
@@ -208,7 +208,9 @@ export const getIsMarketplaceServerEnabled = (state: RootState) => {
 }
 
 export const getIsBuyCrossChainEnabled = (state: RootState) => {
-  if (hasLoadedInitialFlags(state)) {
+  const hasLoadedInitialFlagsx = hasLoadedInitialFlags(state)
+  console.log('Has loaded initial feature flags?', hasLoadedInitialFlagsx)
```
cyaiox, line 212: wild `console.log` appeared!

LautaroPetaccio: Looool, good catch, removed it.

webapp/src/index.tsx
```diff
@@ -1,13 +1,14 @@
...
+
 import { ConnectedRouter } from 'connected-react-router'
 import * as SingleSignOn from '@dcl/single-sign-on-client'
 import WalletProvider from 'decentraland-dapps/dist/providers/WalletProvider'
 import ToastProvider from 'decentraland-dapps/dist/providers/ToastProvider'
 import TranslationProvider from 'decentraland-dapps/dist/providers/TranslationProvider'
 import ModalProvider from 'decentraland-dapps/dist/providers/ModalProvider'
 
-import './setup'
+// import './setup'
```
cyaiox, line 11:
```suggestion

```

LautaroPetaccio: Fixed!

cyaiox (approved):

PR #2123, opened 2024-02-01
fix: Remove unused FFs

LautaroPetaccio:
This PR removes the following FFs:
- IsPriceFilter
- IsEstateSizeFilter
- IsCreatorsFilter
- IsLocationFilter
- IsRentalPeriodFilter
- IsMapViewFilters
- IsRentalPriceFilterChart
- IsHandsCategory
- isEmotesV2
- buy-crosschain


webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -185,19 +182,6 @@ const AssetBrowse = (props: Props) => {
 
   let right: ReactNode
 
-  const mapTopbar = isMapViewFiltersEnabled ? (
-    <MapTopbar
```
meelrossi, line 189: is this ok? if teh FF is enabled we should have the first part not the second

LautaroPetaccio: Great catch! I've restored it

webapp/src/components/AssetBrowse/AssetBrowse.tsx
```diff
@@ -240,9 +219,16 @@ const AssetBrowse = (props: Props) => {
...
+          {isMap && isFullscreen ? (
+            <MapTopbar
+              showOwned={showOwnedLandOnMap}
+              onShowOwnedChange={(show: boolean) => setShowOwnedLandOnMap(show)}
+            />
+          ) : (
+            <AssetTopbar />
+          )}
           {isMap ? (
-            <MapBrowse showOwned={showOwnedLandOnMap} />
```
meelrossi, line 245: i think you need this too

LautaroPetaccio: You're right!

meelrossi (approved):

PR #2130, opened 2024-02-13
fix: Refactor buy with crypto modal

LautaroPetaccio:
This PR refactors the `BuyWithCryptoModal` by:
- Extracting most of the logic of computing the gas cost and the route from the component into props, which are then loaded through their specific components (`MintNftWithCryptoModal` and `BuyNftWithCryptoModal`). This logic has been moved and refactored into hooks.
- Extracting some of the code that generates the modal into two components to reduce the complexity of the modal. These two new componentes are the `PaymentSelector` and the `PurchaseTotal`


fzavalia:
Uploading Screen Recording 2024-02-20 at 10.52.52.mov…

Estimations are not updated when selecting 0 balance tokens. Not important, as it is just a view issue.

fzavalia:
- [x] Mint NFT with Polygon MANA
- [x] Mint NFT with Polygon DAI

fzavalia:
There is an issue when selecting 0 balance tokens, the estimations are not being updated. This will not affect the user given that with 0 balance they cannot proceed, just as a reminder.

fzavalia (approved):
Great idea of separating the PaymentSelector, PurchaseTotal and hooks out of the main component.
I tried some operations and it worked correctly. 
Also committed some formatting and import fixes.

juanmahidalgo (approved):
looks awesome, so much organized than before. Left a few comments of how I think we could improve it further

PR #2132, opened 2024-02-14
fix: Don't set ens filter to on sale by default on the current accoun…

### LINKED ISSUE #2129: [BUG] My Assets page fetching assets
## 🎉 Description

There is a bug when navigating the section: `My Assets` that's adding the param: `onlyOnSale` by default but then removing it when clicking on the Assets Category Component.

* [x] 🐛 This is a bug report.

## 📝 Details

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzVIQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f36dfe65aed3cfc44c27cec38f39fa50fc82ca3f/Screen%20Recording%202024-02-09%20at%2012.26.04.mov)

## 🔢 Steps To Reproduce Issue [ Good To Have ]

Please remember that with baby steps it's easier to reproduce the bug and it's much faster to fix it.

```
- Log in to the marketplace
- Go to My Assets section
- Click on the `Names` category
  - it should show `0 items`
- Click on another category
- Click again on `Names` category
  - it should show all your names
```

fzavalia:
…t page

Closes #2129

### DIFF
--- a/webapp/src/modules/routing/sagas.ts
+++ b/webapp/src/modules/routing/sagas.ts
@@ -696,10 +696,15 @@ function* deriveCurrentOptions(
       // for ENS, if the previous page had `onlyOnSale` as `undefined` like wearables or emotes, it defaults to `true`, otherwise use the current value
       newOptions = {
         ...newOptions,
-        assetType: AssetType.NFT,
-        onlyOnSale:
+        assetType: AssetType.NFT
+      }
+
+      // Only if the user is not in their own page, show ens on sale by default.
+      if (window.location.pathname !== locations.currentAccount()) {
+        newOptions.onlyOnSale =
           previous.onlyOnSale === undefined ? true : current.onlyOnSale
       }
+
       break
     }
     default: {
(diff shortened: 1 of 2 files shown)


webapp/src/modules/routing/sagas.ts
```diff
@@ -696,10 +696,15 @@ function* deriveCurrentOptions(
...
-        assetType: AssetType.NFT,
-        onlyOnSale:
+        assetType: AssetType.NFT
+      }
+
+      // Only if the user is not in their own page, show ens on sale by default.
+      if (window.location.pathname !== locations.currentAccount()) {
+        newOptions.onlyOnSale =
           previous.onlyOnSale === undefined ? true : current.onlyOnSale
       }
```
fzavalia, line 706: What was happening is that on your own account page, it was being set to only on sale, which then forced the only on sale on the other tabs.

cyaiox (approved):

PR #2142, opened 2024-02-21
fix: Fetch smart-wearables permissions

cyaiox:
This PR avoids executing the `fetchSmartWearableRequiredPermissionsRequest` action when it is not required.


juanmahidalgo (approved):
LGTM! we could update the tests to check the smart wearable case and that it puts the action if you want!

PR #2145, opened 2024-02-23
fix: Various mint name with crypto fixes

LautaroPetaccio:
This PR fixes the following:
- Forces the Mana icon in the `BuyWithCryptoButton` to be set as Matic so the white logo is shown. The red logo was fusing with the red color of the button, preventing it from being seen.
- Changes the `ChainButton` when buying with card to be just a simple button.
- Disables the Buy now button when doing a native purchase.
- Closes the `FatFinger` modal when closing the `BuyWithCryptoModal` modal when paying for the minting of a name.
- Adds popups to the icons of the tokens so the token the user is using is displayed in a clearer way.
- Removes the token from the `InfoTooltip` that talked about paying the gas fee with the native token to remove the need to name all native tokens.

### DIFF
--- a/webapp/src/components/Modals/BuyWithCryptoModal/PaymentSelector/PaymentSelector.module.css
+++ b/webapp/src/components/Modals/BuyWithCryptoModal/PaymentSelector/PaymentSelector.module.css
@@ -169,11 +169,11 @@
 }
 
 @media (max-width: 767px) {
-  .buyWithCryptoModal .dropdownContainer {
+  .dropdownContainer {
     flex-direction: column;
   }
 
-  .buyWithCryptoModal .tokenDropdownContainer {
+  .tokenDropdownContainer {
     align-self: flex-start;
     margin-left: 0;
     width: 100%;


webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyWithCryptoButton/BuyWithCryptoButton.types.ts
```diff
@@ -1,6 +1,3 @@
-import type { Network } from "@dcl/schemas"
 import type { ButtonProps } from "decentraland-ui/dist/components/Button/Button"
 
-export type Props = ButtonProps & {
-  assetNetwork: Network
-}
+export type Props = ButtonProps
```
juanmahidalgo: can we add EOF here?

webapp/src/components/Modals/BuyWithCryptoModal/MintNameWithCryptoModal/MintNameWithCryptoModal.tsx
```diff
@@ -103,7 +104,12 @@ const MintNameWithCryptoModalHOC = (props: Props) => {
   const onGoBack = useCallback(() => {
     onClose()
     onOpenFatFingerModal()
-  }, [])
+  }, [onOpenFatFingerModal, onClose])
+
+  const onCloseModal = useCallback(() => {
+    onCloseFatFingerModal()
+    return onClose()
```
juanmahidalgo, line 111:
```suggestion
    onClose()
```
just to keep it consistent with the one above

webapp/src/components/Modals/BuyWithCryptoModal/PurchaseTotal/PurchaseTotal.tsx
```diff
@@ -94,12 +114,31 @@ const PurchaseTotal = (props: Props) => {
...
+                          alt={gasCost.token.name}
+                        />
+                      }
+                    />
                     {formatPrice(Number(gasCost.total), gasCost.token)}
                     <span> + </span>
                   </>
                 ) : null}
-                <img src={selectedToken?.logoURI} alt={selectedToken?.name} />
+                <Popup
```
juanmahidalgo, line 133: Just a thought... it would be nice to abstract this in a component (or even just a function) that recieves the token and returns the wrapped one. (to avoid repetetition of the props `on`, `position`, `style`, etc, that are all the same in this cases)

juanmahidalgo (approved):
a few comments but approving anyways, looks good :clap:

webapp/src/components/Modals/BuyWithCryptoModal/MintNameWithCryptoModal/MintNameWithCryptoModal.tsx
LautaroPetaccio: As we're replacing the `onClose` method from the original one to this one, the `BuyWithCryptoModal` expects the type of the method to return the same type as the original `onClose`. I could either keep the return there or change the type in the `BuyWithCryptoModal` what do you find best?

webapp/src/components/Modals/BuyWithCryptoModal/PurchaseTotal/PurchaseTotal.tsx
LautaroPetaccio: Done! Created the `TokenIcon` component.

webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyWithCryptoButton/BuyWithCryptoButton.types.ts
LautaroPetaccio: Done!

PR #2148, opened 2024-02-23
fix: fix the formatPrice fn to show at least 1 relevant decimal

juanmahidalgo:
## before the fix

image

## after the fix

image

### DIFF
--- a/webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
+++ b/webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
@@ -67,14 +67,21 @@ function truncateToDecimals(num: number, dec = 2) {
 }
 
 export function formatPrice(price: string | number, token: Token): number {
-  // Determine the number of decimals based on the USD price
-  let decimalsToShow: number
+  // Convert price to a number if it's a string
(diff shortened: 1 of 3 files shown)


webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
```diff
@@ -67,14 +67,21 @@ function truncateToDecimals(num: number, dec = 2) {
...
+  let decimalsToShow: number
+  if (token.usdPrice && token.usdPrice >= 1.5) {
+    decimalsToShow = 4 // For tokens priced over $1.5, show up to 4 decimals
   } else {
-    decimalsToShow = 2 // Show 2 decimals for other tokens or higher-value fractions
+    decimalsToShow = 2 // Default to 2 decimals
+  }
+
+  // Ensure small values are displayed with at least two significant digits
+  if (numericPrice !== 0 && numericPrice < 0.01) {
```
LautaroPetaccio, line 82: Is `numericPrice < 0.01` contemplating not being 0?
```suggestion
  if (numericPrice < 0.01) {
```

LautaroPetaccio (approved):

PR #2152, opened 2024-02-23
feat: Show new ENS images

LautaroPetaccio:
This PR adds auto-generated ENS images to the `AssetImage` component.

Screenshot 2024-02-23 at 18 02 18
Screenshot 2024-02-23 at 18 01 27


webapp/src/components/AssetImage/EnsImage/EnsImage.tsx
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useRef } from 'react'
```
juanmahidalgo, line 1: No need to make the change now but I believe it would be nice to move this to the UI so we can re-use it in the builder and other dapps, right?

juanmahidalgo (approved):
LGTM!

webapp/src/components/AssetImage/EnsImage/EnsImage.tsx
LautaroPetaccio: 100% correct. We need to move it to the UI and change then use the AssetImage from the UI in the Marketplace as well

PR #2156, opened 2024-02-27
fix: select the first token from the chain selected

juanmahidalgo:


webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
```diff
@@ -796,7 +799,9 @@ export const BuyWithCryptoModal = (props: Props) => {
                   })}
                 </span>
               ) : null}
-              {!canBuyAsset && !isFetchingBalance && !isFetchingRoute ? (
+              {canBuyAsset === false &&
```
juanmahidalgo, line 802: `canBuyAsset` must be computed, it starts at `undefined`, at that value, we don't know if it has funds or not for the tx

LautaroPetaccio (approved):

PR #2160, opened 2024-02-29
fix: Add gas cost and token balance hooks tests and simplify functions

LautaroPetaccio:
This PR does the following:
- Adds a fix to the `AssetImage` component where it was fetching the item of parcel, estates or ens NFTs.
- Changes the way we load the default chains in the development environment. It now loads the default chains using the `getDefaultChains` function which discriminates between the development and the production environment.
- Simplifies the gas hooks by removing the `providerTokens` variable which was used to get the chain native token. The chain native token is now computed in the modal.
- Simplifies the gas hooks by removing the `selectedChain` variable, as it can be retrieved from the `selectedToken`.
- Simplifies the `useShouldUseCrossChainProvider` by removing the `selectedChain` variable that can be retrieved from the `selectedToken`.
- Splits the `estimateTransactionGas` function into `estimateBuyNftGas` and `estimateMintNftGas` to be used in each hook separately.
- Adds tests for the gas and token balance hooks. 
- Fixes the estimation of minting names.

### DIFF
--- a/webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
+++ b/webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
@@ -1,11 +1,7 @@
-import { ChainId, Item, Network, Order } from '@dcl/schemas'
+import { ChainId, Network, Order, Item } from '@dcl/schemas'
+import { BigNumber, ethers } from 'ethers'
 import { Env } from '@dcl/ui-env'
-import {
-  ChainData,
-  CrossChainProvider,
-  Route,
-  Token
-} from 'decentraland-transactions/crossChain'
+import { ChainData, Token } from 'decentraland-transactions/crossChain'
 import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
 import {
   ContractName,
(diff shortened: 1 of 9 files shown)


webapp/src/components/Modals/BuyWithCryptoModal/utils.ts
```diff
@@ -207,81 +201,50 @@ export const getDefaultChains = () => {
...
+): Promise<BigNumber> => {
   const networkProvider = await getNetworkProvider(selectedChain)
   const provider = new ethers.providers.Web3Provider(networkProvider)
-
-  const contract = getContract(ContractName.DCLRegistrar, selectedChain)
+  const contract = getContract(ContractName.DCLControllerV2, selectedChain)
   const c = new ethers.Contract(contract.address, contract.abi, provider)
-  const estimation = await c.estimateGas.register(ownerAddress, name)
-  return estimation
+  return c.estimateGas.register(name, ownerAddress)
```
juanmahidalgo, line 249:
```suggestion
  return c.estimateGas.register(ownerAddress, name)
```
I belive this should remaing like this, is there any reason why we're changing it?

LautaroPetaccio: I've changed the contract, from the `DCLRegistrar` to the `DCLControllverV2` because it was wrong and the `register` method to have [the correct parameters according to its ABI](https://github.com/decentraland/decentraland-transactions/blob/6d4a6d59b3cd5f0961308f9492e62ecdcc12d88e/src/abis/DCLControllerV2.ts#L134) which have the name first and then the address of the beneficiary.

juanmahidalgo (approved):
LGTM!

PR #2167, opened 2024-03-05
feat: Use new cross chain transaction payload action

LautaroPetaccio:
This PR uses the new version of the transactions module, which fetches and waits for pending cross chain transactions.
It removes all the logic created to work with cross chain transactions once they were launched.

### DIFF
--- a/webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
+++ b/webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
@@ -378,7 +378,9 @@ export const BuyWithCryptoModal = (props: Props) => {
           data-testid={GET_MANA_BUTTON_TEST_ID}
           loading={isFetchingBalance || isBuyingAsset}
           onClick={() => {
-            onGetMana()
+            // onGetMana()
+            console.log("Getting MANA")
+            handleCrossChainBuy()
             onClose()
           }}
         >
@@ -405,6 +407,7 @@ export const BuyWithCryptoModal = (props: Props) => {
     asset.chainId,
     isLoadingAuthorization,
     onBuyWithCard,
+    handleCrossChainBuy,
     handleBuyWithCard,
     onGetMana,
     onClose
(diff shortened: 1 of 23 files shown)


webapp/src/modules/item/sagas.ts
```diff
@@ -296,67 +287,10 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
...
-          )
-        )
-      }
-    }
-  }
-
   function* handleBuyItemCrossChain(action: BuyItemCrossChainRequestAction) {
     const { item, route, order } = action.payload
     try {
+      console.log('Buying item cross chain')
```
juanmahidalgo, line 293: whoops

webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
```diff
@@ -378,7 +378,9 @@ export const BuyWithCryptoModal = (props: Props) => {
           data-testid={GET_MANA_BUTTON_TEST_ID}
           loading={isFetchingBalance || isBuyingAsset}
           onClick={() => {
-            onGetMana()
+            // onGetMana()
+            console.log("Getting MANA")
```
juanmahidalgo, line 382:
```suggestion
            onGetMana()
```

webapp/src/modules/item/sagas.ts
LautaroPetaccio: Removed it!

webapp/src/config/env/dev.json
```diff
@@ -44,6 +44,7 @@
   "SUBGRAPH_WORKER": "https://subgraph.decentraland.zone",
   "SENTRY_DSN": "https://1dc401149e1c819b8477565c9cdd9b70@o4504361728212992.ingest.sentry.io/4505743351676928",
   "SQUID_API_URL": "https://testnet.v2.api.squidrouter.com/",
+  "SQUID_RETRY_DELAY": "3000",
```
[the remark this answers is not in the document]
LautaroPetaccio: Sure! Changed it to 2 secs.

PR #2169, opened 2024-03-07
feat: Add cross chain payments for LANDs

juanmahidalgo:
image

image
(note: buy now button hardcoded, it should render the BUY MANA one since I don't have enough balance)

### DIFF
--- a/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
+++ b/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
@@ -102,11 +102,11 @@
 
 .saleButtons {
   display: flex;
-  flex-direction: row;
+  flex-direction: column;
 }
 
 .buy + .fullWidth > .bid {
-  margin-left: 8px !important;
+  margin-top: 8px !important;
 }
 
 .bid {
--- a/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
+++ b/webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
@@ -105,7 +105,7 @@
   flex-direction: column;
 }
 
-.buy + .fullWidth > .bid {
+.fullWidth > .bid {
   margin-top: 8px !important;
 }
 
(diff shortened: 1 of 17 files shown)


webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -283,20 +285,31 @@ const SaleRentActionBox = ({
...
-                      )}
-                      disabled={!hasEnoughManaToBuy}
-                      className={styles.buy}
-                      primary
-                      fluid
-                    >
-                      {t('asset_page.actions.buy')}
-                    </Button>
+                    isCrossChainLandEnabled ? (
+                      <Button
```
LautaroPetaccio, line 289: We have the `BuyWithCryptoButton` component, would it fit here if we used it?

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
```diff
@@ -102,11 +102,11 @@
 
 .saleButtons {
   display: flex;
-  flex-direction: row;
+  flex-direction: column;
```
LautaroPetaccio, line 105: Will changing this break the style of the sales buttons when the FF is disabled?

LautaroPetaccio (commented):
Great work, looks great 👍

LautaroPetaccio (approved):

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.module.css
juanmahidalgo: Yeah, I thought about this, but I didn't want to add a new class so it later gets deleted with the FF
It would look like this, it won't necessarly be broken
image

wdyt?

LautaroPetaccio: It does look good, let's keep it

webapp/src/components/AssetPage/SaleRentActionBox/SaleRentActionBox.tsx
```diff
@@ -283,20 +286,24 @@ const SaleRentActionBox = ({
...
-                      )}
-                      disabled={!hasEnoughManaToBuy}
-                      className={styles.buy}
-                      primary
-                      fluid
-                    >
-                      {t('asset_page.actions.buy')}
-                    </Button>
+                    isCrossChainLandEnabled ? (
+                      <BuyWithCryptoButton onClick={() => onBuyWithCrypto()} />
```
LautaroPetaccio, line 290:
```suggestion
                      <BuyWithCryptoButton onClick={onBuyWithCrypto} />
```

PR #2196, opened 2024-03-19
fix: Linting rules (no-unsafe-argument)

LautaroPetaccio:
This PR fixes the `no-unsafe-argument` linting rule.


webapp/src/modules/item/sagas.ts
```diff
@@ -91,7 +91,7 @@ export function* itemSaga(getIdentity: () => AuthIdentity | undefined) {
 
   // to avoid race conditions, just one fetch items request is handled at once in the browse page
   function* takeLatestByPath(actionType: string, path: string): SagaIterator {
-    let task
+    let task: Task<any> | undefined
```
meelrossi, line 94: can we set this a unknown?

meelrossi (approved):

PR #2197, opened 2024-03-20
fix: LANDs section race condition

juanmahidalgo:


webapp/src/components/AssetList/AssetList.container.ts
```diff
@@ -25,18 +26,22 @@ const mapState = (state: RootState): MapStateProps => {
   const section = getSection(state)
   const page = getPageNumber(state)
   const assetType = getAssetType(state)
+  const view = getView(state)
+  const loadingState = getLoadingNFTs(state).filter(loading => loading.payload.options.view === view)
+  const assets = getBrowseAssets(state, section, assetType)
+  const loadingStateFilteredByView = loadingState.filter(ls => ls.payload.options.view === View.MARKET)
```
LautaroPetaccio, line 32: As the `AssetList` component is used in a lot of places, will this restriction affect other places that don't share the Market view?

LautaroPetaccio (approved):
Looks great, send it 🚀

PR #2222, opened 2024-04-17
fix: fingerprint estates issue

juanmahidalgo:


webapp/src/components/Modals/BuyWithCryptoModal/hooks.ts
```diff
@@ -172,9 +173,13 @@ export const useBuyNftGasCost = (
 ): GasCost => {
   const chainId = parseInt(selectedToken.chainId) as ChainId
 
+  const [fingerprint] = useFingerprint(nft)
```
LautaroPetaccio, line 176: Instead of re-computing the fingerprint here, we could receive it as an optional parameter and fill it in the `BuyWithCryptoModal` component, wdyt?

LautaroPetaccio (approved):

PR #2239, opened 2024-04-26
feat: change my assets default order and section

### LINKED ISSUE #670: Change MyStore label and default section
When entering the Marketplace and starting browsing as a user, I can see a tab labeled "My Store". When I go to that tab, the default section is "Collections". This is not so intuitive because I'm looking for putting my Assets for sale in my store, not looking for my store to put my assets for sale. Sounds similar but the thought process of users is inverse. 
On the other hand, if the user doesn't have a collection created by themselves, in other words, it is not a creator, the default section is most of the time empty. 
To fix these 2 UX issues, the following changes should be made:

1. Change the tab label to "My Assets"
2. Change the default option to "Wearables"
3. Change the order of the options on the Assets menu to make it consistent with item 2. When a user enters My Assets, it should look this way:

### LINKED ISSUE #1348: NFT not found page is not aligned
## 🎉 Description

when accessing the NFT detail page and the request to the server fails (you can try locally) `/contracts/0xffc5043d9a00865d089d5eefa5b3d1625aec6763/tokens/99`, the not found page is not aligned

* [X] 🐛 This is a bug report.

### LINKED ISSUE #1550: New order for My Assets sidebar navigation
We need to re-arrange the sidebar based on the product team investigation and what users use the most.

This is the new order:

Note: The navbar should set the `Wearables` tab as the default section instead of the `Collections` as it is right now.

### LINKED ISSUE #1599: Change default section in My Assets
## 🎉 Description

When any user opens the My Assets section in the Marketplace, they are redirected to the Collections section. Current default view:

The problem here is that most users have never created a collection, so this section will always be empty. Instead, the default section should be Wearables, as most users will have at least one in there. 

So when any user enters the My Assets tab, by default has to be redirected to the Wearables section: 

* [ ] 🐛 This is a bug report.

## 📝 Details

Describe the problem you have been experiencing in more detail. Include as much information as you think is relevant. Keep in mind that transactions can fail for many reasons; context is key here. Links to tx, screenshots are welcome.
Remember to not share sensible information

## 🔢 Steps To Reproduce Issue [ Good To Have ]

Please remember that with baby steps it's easier to reproduce the bug and it's much faster to fix it.

```
- Step 1
- Step 2
- Step 3
...
```

meelrossi:
Closes #1599 
Closes #1550 
Closes #670 
Closes #1348


LautaroPetaccio (approved):
Send it 🚀

PR #2240, opened 2024-04-29
fix: add timeout when MM active method hangs

juanmahidalgo:
When MM is blocked, this saga keeps waiting for ever since it won't succed or failed. This PR adds a timeout to avoid that

### DIFF
--- a/webapp/src/modules/wallet/utils.spec.ts
+++ b/webapp/src/modules/wallet/utils.spec.ts
@@ -64,9 +64,8 @@ describe('when waiting for the wallet to connect', () => {
       return expectSaga(waitForWalletConnectionAndIdentityIfConnecting)
         .provide([
           [select(isConnecting), true],
-          [delay(WAIT_FOR_WALLET_CONNECTION_TIMEOUT + 1), void 0]
+          [delay(WAIT_FOR_WALLET_CONNECTION_TIMEOUT), void 0]
         ])
-        .delay(WAIT_FOR_WALLET_CONNECTION_TIMEOUT + 1) // Simulate a delay longer than the timeout
         .run()
     })
   })
--- a/webapp/src/modules/wallet/utils.spec.ts
+++ b/webapp/src/modules/wallet/utils.spec.ts
@@ -58,4 +58,13 @@ describe('when waiting for the wallet to connect', () => {
         .run()
     })
   })
+
+  describe('and the wallet connection times out', () => {
+    it('should finish waiting after the timeout', () => {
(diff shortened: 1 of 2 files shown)


webapp/src/modules/wallet/utils.ts
```diff
@@ -47,10 +48,12 @@ export function formatBalance(balance: number) {
...
   const isConnectingToWallet = (yield select(isConnecting)) as ReturnType<typeof isConnecting>
   if (isConnectingToWallet) {
-    const { success } = (yield race({
+    const { success, timeout } = (yield race({
       success: take(CONNECT_WALLET_SUCCESS),
-      failure: take(CONNECT_WALLET_FAILURE)
-    })) as { success: ConnectWalletSuccessAction; failure: ConnectWalletFailureAction }
+      failure: take(CONNECT_WALLET_FAILURE),
+      timeout: delay(WAIT_FOR_WALLET_CONNECTION_TIMEOUT) // 10 seconds timeout
+    })) as { success: ConnectWalletSuccessAction; failure: ConnectWalletFailureAction; timeout: any }
```
LautaroPetaccio, line 55: Would you mind typing the return of delay? We don't have the linter rule available for it yet, but typing it would help on the changes we need to do:
```suggestion
    })) as { success: ConnectWalletSuccessAction; failure: ConnectWalletFailureAction; timeout: undefined }
```
If I'm not wrong, it should return undefined

webapp/src/modules/wallet/utils.ts
```diff
@@ -47,10 +48,12 @@ export function formatBalance(balance: number) {
...
+    console.log('timeout: ', timeout)
```
LautaroPetaccio, line 56: Would you mind removing the console log?

webapp/src/modules/wallet/utils.spec.ts
```diff
@@ -58,4 +58,13 @@ describe('when waiting for the wallet to connect', () => {
         .run()
     })
   })
+
+  describe('and the wallet connection times out', () => {
+    it('should finish waiting after the timeout', () => {
+      return expectSaga(waitForWalletConnectionAndIdentityIfConnecting)
+        .provide([[select(isConnecting), true]])
+        .delay(10000) // Simulate a delay longer than the timeout
```
LautaroPetaccio, line 66: This should make the test take at least 10 seconds. What if we mock in the provide the delay so it executes instantly?

LautaroPetaccio (approved):

PR #2257, opened 2024-06-05
feat: remove connected react router from sagas

meelrossi:
Replace connected-react-router library with history to handle app redirections
Add history object in saga context to be able to redirect once a saga is done


LautaroPetaccio (approved):
Great job 👍

PR #2265, opened 2024-07-08
feat: create bids trades

meelrossi:


webapp/src/modules/bid/sagas.ts
```diff
@@ -33,135 +38,152 @@ import {
   fetchBidsByNFTFailure,
   acceptBidtransactionSubmitted
 } from './actions'
+import * as bidUtils from './utils'
 
-export function* bidSaga() {
+export function* bidSaga(getIdentity: () => AuthIdentity | undefined) {
```
LautaroPetaccio, line 43: What if we make it possible for the saga to receive an instance of the MarketplaceAPI instead of the identity? In that way, we could simplify the tests by removing the need of using jest.

webapp/src/modules/bid/sagas.spec.ts
```diff
@@ -1,20 +1,167 @@
...
+              contractAddress: '0x1234',
+              tokenId: '1',
+              extra: '',
+              beneficiary: wallet.address
+            }
+          ]
+        }
+      })
+
+      it('should create bid successfully', () => {
```
LautaroPetaccio, line 90: What do you think about adding a test to excercise when the trade creation fails?

webapp/src/modules/bid/utils.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+        })
+      })
+    })
+
+    describe('and a fingerprint is not provided', () => {
+      beforeEach(() => {
+        fingerprint = ''
+      })
+
+      it('should return the trade object', async () => {
```
LautaroPetaccio, line 139:
```suggestion
      it('should return the trade object without the fingerpint', async () => {
```

webapp/src/modules/bid/utils.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    beforeEach(() => {
+      asset = { chainId: ChainId.ETHEREUM_SEPOLIA, network: Network.ETHEREUM, tokenId: 'token-id', contractAddress: '0xcontract' } as Asset
+    })
+
+    describe('and a fingerprint is provided', () => {
+      beforeEach(() => {
+        fingerprint = 'fingerprint'
+      })
+
+      it('should return the trade object', async () => {
```
LautaroPetaccio, line 94:
```suggestion
      it('should return the trade object with the fingerprint', async () => {
```

webapp/src/modules/vendor/decentraland/marketplace/api.ts
```diff
@@ -19,8 +19,19 @@ export class MarketplaceAPI extends BaseAPI {
       [ChainId.ARBITRUM_MAINNET]: 'arbitrum-mainnet',
       [ChainId.FANTOM_MAINNET]: 'fantom-mainnet'
     } as Record<ChainId, string>
-    return this.request('get', `/${chainIdToChainName[chain]}/address/${wallet}/balance`) as Promise<Balance[]>
+    return this.fetch(`/${chainIdToChainName[chain]}/address/${wallet}/balance`, { method: 'GET' })
+  }
+
+  addTrade = async (trade: TradeCreation) => {
```
LautaroPetaccio, line 25: Would you mind adding a test for this client request?

LautaroPetaccio (approved):
Looks amazing 🔥 
I've left some comments about some tests, but the PR is good to be merged.

PR #2266, opened 2024-07-10
fix: Names search showing other assets

LautaroPetaccio:
This PR fixes an issue where:
1. When in the name's browse site, other assets would be loaded when loading more names.
2. The search bar dropdown showed wearables instead of names.

By:
1. Setting the correct section when the user is in the names browse section so the browse sagas can set the correct category (ENS).
2. Adding the capability to **not** show the search bar dropdown for names, as the feature does not work with names for the time being.


juanmahidalgo (approved):
LGTM! great job 👏

PR #2268, opened 2024-07-16
feat: add accept trade flow

meelrossi:
Add flow to accept bids from offchain marketplace


webapp/src/components/Bid/Bid.tsx
```diff
@@ -67,17 +124,19 @@ const Bid = (props: Props) => {
               <div className="actions">
                 {isBidder ? (
                   <>
-                    <Button primary onClick={() => history.push(locations.bid(bid.contractAddress, bid.tokenId))}>
-                      {t('global.update')}
-                    </Button>
+                    {'bidAddress' in bid && (
```
cyaiox, line 127: should we show a disabled state/message when the bidAddress is not present in the bid?

cyaiox (approved):

webapp/src/components/Bid/Bid.tsx
meelrossi: if the bidaddress is not in the bid then it means that is one of the "new" bids created by a trade. This new bids can't be updated. The process requires the cancellation of the previous trade and hte addition of the new one. That is why the button should not even appear.

webapp/src/utils/trades.ts
```diff
@@ -111,7 +98,44 @@ export async function getTradeSignature(trade: Omit<TradeCreation, 'signature'>)
       beneficiary: asset.beneficiary
     }))
   }
+}
+
+export function getTradeToAccept(trade: Trade) {
```
LautaroPetaccio, line 103: It would be importante to test these function to be sure that the trade is being built correctly, would you mind adding tests for them?

webapp/src/modules/bid/sagas.ts
```diff
@@ -147,20 +178,27 @@ export function* bidSaga(marketplaceAPI: MarketplaceAPI) {
...
-        const { bidService } = VendorFactory.build(vendorName)
-        if (bidService === undefined) {
-          continue
-        }
+      const isBidsOffchainEnabled: boolean = yield select(getIsBidsOffChainEnabled)
+      if (isBidsOffchainEnabled) {
+        sellerBids = ((yield call([marketplaceAPI, 'fetchBids'])) as Awaited<ReturnType<typeof marketplaceAPI.fetchBids>>).results // TODO: add seller filter
+        bidderBids = (
+          (yield call([marketplaceAPI, 'fetchBids'], { bidder: address })) as Awaited<ReturnType<typeof marketplaceAPI.fetchBids>>
+        ).results
```
LautaroPetaccio, line 186: What if we make this request using yield all as the one below is doing? That should speed up the loading of the bids.

webapp/src/modules/bid/sagas.spec.ts
```diff
@@ -221,157 +224,278 @@ describe('when handling the accepting a bid action', () => {
...
+          name: 'OffChainMarketplace',
+          version: '1.0.0'
+        }
+
+        txHash = '0x12312412'
+
+        marketplaceAPIMock = { fetchTrade: jest.fn().mockResolvedValue(trade) } as unknown as jest.Mocked<MarketplaceAPI>
+      })
+
+      it.only('should dispatch an action signaling the success of the action handling', () => {
```
LautaroPetaccio, line 481: A wild only has appeared

webapp/src/modules/bid/sagas.ts
```diff
@@ -83,24 +95,43 @@ export function* bidSaga(marketplaceAPI: MarketplaceAPI) {
...
+      if ('tradeId' in bid) {
+        if (isBidsOffchainEnabled) {
+          const trade: Trade = yield call([marketplaceAPI, 'fetchTrade'], bid.tradeId)
+          const tradeToAccept = tradeUtils.getTradeToAccept(trade)
+          const offchainMarketplaceContract: ContractData = yield call(getDCLContract, ContractName.OffChainMarketplace, trade.chainId)
+          txHash = yield call(
+            sendTransaction as (contract: ContractData, contractMethodName: string, ...contractArguments: any[]) => Promise<string>,
+            offchainMarketplaceContract,
+            'function accept(Trade[] calldata _trades) external;',
+            [tradeToAccept]
```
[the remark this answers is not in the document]
meelrossi: Added one task to the backlog to handle this

LautaroPetaccio (approved):

PR #2270, opened 2024-07-19
feat: show correctly bids on nfts

meelrossi:
- Rename fetchBidsByNFT action to fetchBidsByAssett to accept also fetching all bids for an item.


webapp/src/modules/bid/reducer.spec.ts
```diff
@@ -16,3 +18,39 @@ describe('when clear bid error action is received', () => {
...
+describe('when fetching bids by asset request action is received', () => {
+  it('should set the action as loading', () => {
+    const request = fetchBidsByAssetRequest({} as Asset)
+    expect(bidReducer(state, request)).toEqual(expect.objectContaining({ loading: expect.arrayContaining([request]) }))
+  })
+})
+
+describe('when fetching bids by asset failure action is received', () => {
+  it('should add the error to the state', () => {
+    const error = 'Some test error'
```
LautaroPetaccio, line 31: Would you mind asserting that the request action is removed from the loading state?

webapp/src/modules/bid/sagas.ts
```diff
@@ -207,19 +207,29 @@ export function* bidSaga(marketplaceAPI: MarketplaceAPI) {
...
+    const { asset } = action.payload
+    let bids: Bid[] = []
     try {
-      const { bidService } = VendorFactory.build(nft.vendor)
-      if (!bidService) {
-        throw new Error("Couldn't find a valid bid service for vendor")
-      }
+      const isBidsOffchainEnabled: boolean = yield select(getIsBidsOffChainEnabled)
+      if (isBidsOffchainEnabled) {
+        const response: Awaited<ReturnType<typeof marketplaceAPI.fetchBids>> = yield call([marketplaceAPI, 'fetchBids'], {
```
LautaroPetaccio, line 216: It would be great if we could move this interaction into the services eventually, would you mind taking notes on that?

LautaroPetaccio (approved):

PR #2275, opened 2024-07-24
fix: address null

meelrossi:
Fix address coming as null in `addressEquals` helpers. Do not allow to accept a trade when there is no more available stock

### DIFF
--- a/webapp/src/modules/wallet/utils.ts
+++ b/webapp/src/modules/wallet/utils.ts
@@ -22,7 +22,7 @@ export function shortenAddress(address: string) {
 }
 
 export function addressEquals(address1?: string, address2?: string) {
-  return !!address1 && !!address2 && address1.toLowerCase() === address2.toLowerCase()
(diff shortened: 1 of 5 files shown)


webapp/src/modules/wallet/utils.ts
```diff
@@ -22,7 +22,7 @@ export function shortenAddress(address: string) {
 }
 
 export function addressEquals(address1?: string, address2?: string) {
-  return address1 !== undefined && address2 !== undefined && address1.toLowerCase() === address2.toLowerCase()
+  return !!address1 && !!address2 && address1.toLowerCase() === address2.toLowerCase()
```
LautaroPetaccio, line 25: What do you think of doing:
```suggestion
  return address1?.toLowerCase() === address2?.toLowerCase()
```
It would allow us to compare two null or undefined strings as well

LautaroPetaccio (approved):

Issue #590, opened 2022-02-07
Show full name when hovering over Lands and Wearables


nicorajco:

nicosantangelo:
I think I'd use a `title` prop for starters. We can check how it looks with tooltip but I think it'd be more annoying than useful in both cases

ogabor82:
@nicorajco @nicosantangelo 
Created a PR about the fix:
https://github.com/decentraland/marketplace/pull/2290

PR #2290, opened 2024-08-23
Fix: Show full name of Lands and Wearables

ogabor82:
image

image


ogabor82:
related issue:
https://github.com/decentraland/marketplace/issues/590

PR #2319, opened 2024-10-31
feat: add cancel listing step to LANDs flow

juanmahidalgo:
Step added: 

image

### DIFF
--- a/webapp/src/components/Modals/SellModal/SellModal.tsx
+++ b/webapp/src/components/Modals/SellModal/SellModal.tsx
@@ -18,6 +18,7 @@ import { getAssetName, isOwnedBy } from '../../../modules/asset/utils'
 import { getDefaultExpirationDate, INPUT_FORMAT } from '../../../modules/order/utils'
 import { locations } from '../../../modules/routing/locations'
 import { getContractNames, VendorFactory } from '../../../modules/vendor'
+import ErrorBanner from '../../ErrorBanner'
 import { ManaField } from '../../ManaField'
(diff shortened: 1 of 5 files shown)


webapp/src/components/Modals/SellModal/SellModal.tsx
```diff
@@ -318,33 +321,45 @@ const SellModal = ({
...
           <Button disabled={isCancelling} onClick={onClose}>
             {t('global.cancel')}
           </Button>
-          <Button type="submit" primary disabled={isCancelling} loading={isCancelling} onClick={() => onCancelOrder(order!, nft)}>
+          <Button
+            type="submit"
+            primary
+            disabled={isCancelling}
+            loading={isCancelling}
+            onClick={() => !!order && onCancelOrder(order, nft, shouldRemoveOffchainListing)}
```
LautaroPetaccio, line 361: Although the type of `onCancelOrder` says that it can be called with a third parameter, the [SellModal.container.ts](https://github.com/decentraland/marketplace/blob/52b7b4e6f9d0156d89a7f44f834841de160d2ef6/webapp/src/components/Modals/SellModal/SellModal.container.ts) file, which defines the `onCancelOrder` only has two.

LautaroPetaccio (approved):
There's just a small thing to change, but it looks good!

webapp/src/components/Modals/SellModal/SellModal.tsx
juanmahidalgo: good catch, params updated in the container to pass the 3rd parameter!

PR #2344, opened 2024-12-29
fix: Multiple off chain issues

LautaroPetaccio:
This PR fixes the following issues:
- The cancel sell page was not disabling the cancel button when cancelling an order.
- The ConfirmInputValue, which is used to confirm the process of placing an order was not correctly disabling the inputs when loading.
- The sell modal was not preventing users from closing it when cancelling the order.
- When creating an order, the site reloaded the NFT to get information about the order, but, as the request is now cached, the user wasn't seeing the NFT on sale for the duration of the cache.
- When cancelling an order, specially on lands, the modal was not closed after finishing the process.
- (Not an off chain issue) Removes the owner div on L1 items, thus aligning the components to look nicer.


juanmahidalgo (approved):
great stuff, LGTM!

PR #2354, opened 2025-01-24
fix: Campaign page without items

LautaroPetaccio:
This PR fixes the case where the user loaded a campaign tab that did not have items tagged as the campaign was designed to.

### DIFF
--- a/webapp/src/modules/routing/sagas.ts
+++ b/webapp/src/modules/routing/sagas.ts
@@ -1,8 +1,10 @@
+import { ethers } from 'ethers'
 import { History } from 'history'
 import { AnyAction } from 'redux'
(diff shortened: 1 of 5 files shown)


webapp/src/modules/routing/sagas.ts
```diff
@@ -158,6 +165,7 @@ export function* fetchAssetsFromRoute(options: BrowseOptions) {
     emoteHasGeometry,
     emoteHasSound
   } = options
+  console.log('Sagas: Fetching assets from route', contracts)
```
cyaiox, line 168:
```suggestion
```

cyaiox (approved):

PR #2372, opened 2025-02-07
feat: Add magic auto authorization feature

LautaroPetaccio:
This PR adds the Magic Auto Authorization feature. This implies that Magic users will not be prompted to authorize contracts on most of the operations on the Marketplace. There's a `SellModal` that still has authorizations, but other operations should not show the authorization modal.
The new changes imply that we must propagate the error of the authorizations to the components. This is done in several components, most of them which use the `ConfirmInputValueModal` which now handles errors.

### DIFF
--- a/webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
+++ b/webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
@@ -28,6 +28,7 @@ const ConfirmRentModal = ({
   onAuthorizedAction,
   getContract,
   isLoadingAuthorization,
+  authorizationError,
   onClearRentalErrors,
   error
 }: Props) => {
@@ -56,7 +57,7 @@ const ConfirmRentModal = ({
     if (isUserTheOperatorAddress) {
       setOperatorAddress('')
     } else {
-      setOperatorAddress(wallet!.address)
+      setOperatorAddress(wallet.address)
     }
     setIsUserTheOperatorAddress(!isUserTheOperatorAddress)
   }, [isUserTheOperatorAddress, wallet, setIsUserTheOperatorAddress])
@@ -131,7 +132,9 @@ const ConfirmRentModal = ({
           </div>
         </div>
 
-        {error ? <Message error size="tiny" visible content={error} header={t('global.error')} /> : null}
+        {error || authorizationError ? (
+          <Message error size="tiny" visible content={error || authorizationError} header={t('global.error')} />
+        ) : null}
       </Modal.Content>
       <Modal.Actions className={styles.actions}>
         <Button className={styles.cancel} secondary disabled={isLoading} onClick={onClose}>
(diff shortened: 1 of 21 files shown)


webapp/src/components/Modals/ConfirmRentModal/ConfirmRentModal.tsx
```diff
@@ -56,7 +57,7 @@ const ConfirmRentModal = ({
     if (isUserTheOperatorAddress) {
       setOperatorAddress('')
     } else {
-      setOperatorAddress(wallet!.address)
+      setOperatorAddress(wallet?.address)
```
cyaiox, line 60: should we show an error message when the wallet is undefined?

webapp/src/components/Modals/SellModal/SellModal.tsx
```diff
@@ -77,57 +74,47 @@ const SellModal = ({
...
-    contractName:
-      (nft.category === NFTCategory.WEARABLE || nft.category === NFTCategory.EMOTE) && nft.network === Network.MATIC
-        ? ContractName.ERC721CollectionV2
-        : ContractName.ERC721,
-    chainId: nft.chainId,
-    type: AuthorizationType.APPROVAL
-  }
-
-  const [isLoadingAuthorizations, isAuthorized] = useAuthorization(authorization, onFetchAuthorizations)
+  const authorizedContract = !!offchainOrdersContract && isOffchainPublicNFTOrdersEnabled ? offchainOrdersContract : marketplaceContract
```
cyaiox, line 84:
```suggestion
  const authorizedContract = offchainOrdersContract || marketplaceContract
```

webapp/src/components/Modals/SellModal/SellModal.tsx
```diff
@@ -77,57 +74,47 @@ const SellModal = ({
...
   const [fingerprint] = useFingerprint(nft)
 
   if (!wallet) {
     return null
   }
 
-  const contract = getContract({
-    address: authorization.authorizedAddress
-  })
+  const handleCreateOrder = () => onCreateOrder(nft, parseMANANumber(price), new Date(`${expiresAt} 00:00:00`).getTime(), fingerprint)
```
cyaiox, line 91:
```suggestion
  const handleCreateOrder = useCallback(() => onCreateOrder(nft, parseMANANumber(price), new Date(`${expiresAt} 00:00:00`).getTime(), fingerprint), [expiresAt, fingerprint, nft, price, onCreateOrder])
```

LautaroPetaccio: Changed it!

webapp/src/components/Modals/SellModal/SellModal.tsx
LautaroPetaccio: Great catch! Changed it

cyaiox (approved):

PR #2389, opened 2025-03-18
feat: introduce the Credits feature

juanmahidalgo:


webapp/src/components/AssetPage/PriceComponent/PriceComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ManaToFiat } from '../../ManaToFiat'
+import { Props } from './PriceComponent.types'
+import styles from './PriceComponent.module.css'
+
+const PriceComponent = ({ price, network, useCredits, credits, className }: Props) => {
+  const getAdjustedPrice = (originalPrice: string) => {
+    if (!useCredits || !credits) return originalPrice
+    const bigIntPrice = BigInt(originalPrice) - BigInt(credits.totalCredits)
+    return bigIntPrice > 0 ? bigIntPrice.toString() : '0'
+  }
```
LautaroPetaccio, line 14: What do you think about using an `useCallback` or `useMemo` hooks here?

webapp/src/components/AssetPage/PriceComponent/PriceComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (useCredits && credits) {
+    const adjustedPrice = getAdjustedPrice(price)
+    return (
+      <div className={classNames(styles.PriceContainer, className)}>
+        <div className={styles.manaContainer}>
+          <Mana withTooltip size="large" network={network} />
+          <div className={classNames(styles.prices, styles.price)}>
```
LautaroPetaccio, line 22: As this is a brand new component, would you mind adding tests for it?

webapp/src/components/AssetPage/SaleActionBox/UseCreditsToggle/UseCreditsToggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Props } from './UseCreditsToggle.types'
+import styles from './UseCreditsToggle.module.css'
+
+const UseCreditsToggle = ({ assetPrice, credits, isOwner, useCredits, onUseCredits }: Props) => {
+  const handleToggleCredits = useCallback(() => {
+    onUseCredits(!useCredits)
+  }, [useCredits, onUseCredits])
+
+  // convert credits to eth, they are stored in wei
+  const creditsInEth = useMemo(() => {
```
LautaroPetaccio, line 16: What if we rename this variable to `creditsToUseInEth`?

webapp/src/components/Modals/BuyWithCryptoModal/BuyNftWithCryptoModal/BuyNftWithCryptoModal.tsx
```diff
@@ -51,19 +52,34 @@ const BuyNftWithCryptoModalHOC = (props: Props) => {
 
     const offchainMarketplace =
       isOffchainPublicNFTOrdersEnabled && order?.tradeId && getDCLContract(ContractName.OffChainMarketplace, nft.chainId)
+    const creditsManager = getDCLContract(ContractName.CreditsManager, nft.chainId)
+
+    const areCreditsEnoughToBuy = useCredits && credits && BigInt(credits.totalCredits) >= BigInt(order.price)
+    const needsToAuthorizeCredits = useCredits && !areCreditsEnoughToBuy
```
LautaroPetaccio, line 58: Shouldn't we authorize the credits contract even if the amount of credits is not enough to get the NFT for free? For example, shouldn't I be able to buy an NFT that costs 5 MANA with 3 credits to pay for 2 MANA?
The same applies to the Mint version of the modal.

webapp/src/components/Modals/BuyWithCryptoModal/MintNftWithCryptoModal/MintNftWithCryptoModal.tsx
```diff
@@ -78,9 +92,16 @@ const MintNftWithCryptoModalHOC = (props: Props) => {
     [item]
   )
 
+  const price = useMemo(() => {
```
LautaroPetaccio, line 95: This price conversion is only done here, do we also need it in the other modals?
We might be able to abstract it if it's the same function.

webapp/src/components/Modals/BuyWithCryptoModal/BuyNftWithCryptoModal/BuyNftWithCryptoModal.tsx
```diff
@@ -51,19 +52,34 @@ const BuyNftWithCryptoModalHOC = (props: Props) => {
...
-      authorizedContractLabel: offchainMarketplace ? offchainMarketplace.name : marketplace.label || marketplace.name,
+      authorizedContractLabel,
       requiredAllowanceInWei: order.price,
-      onAuthorized: (alreadyAuthorized: boolean) => onExecuteOrder(order, nft, fingerprint, !alreadyAuthorized) // undefined as fingerprint
+      onAuthorized: (alreadyAuthorized: boolean) => onExecuteOrder(order, nft, fingerprint, !alreadyAuthorized, useCredits) // undefined as fingerprint
     })
-  }, [nft, order, fingerprint, getContract, onAuthorizedAction, onExecuteOrder])
+  }, [nft, order, fingerprint, getContract, onAuthorizedAction, onExecuteOrder, useCredits, credits])
 
   const onBuyWithCard = useCallback(() => {
```
LautaroPetaccio, line 84: We have the buy with card option here, should we add the credits logic here as well?

webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyNFTButtons.tsx
```diff
@@ -62,10 +76,29 @@ const BuyNFTButtons = ({
           if (asset && shouldOpenBuyWithCryptoModal) {
             onBuyWithCrypto(asset, order)
           }
+          const isItemFree = !isNFT(asset) && asset.price === '0'
+          const isBuyingEntirelyWithCredits =
+            useCredits &&
+            ((!isNFT(asset) && BigInt(credits.totalCredits) >= BigInt(asset.price)) ||
+              (isNFT(asset) && order && BigInt(credits.totalCredits) >= BigInt(order.price)))
```
LautaroPetaccio, line 83: Would that work as well?
```suggestion
            !isNFT(asset) ? BigInt(credits.totalCredits) >= BigInt(asset.price)) : order && BigInt(credits.totalCredits) >= BigInt(order.price)))
```

webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyNFTButtons.tsx
```diff
@@ -37,21 +43,29 @@ const BuyNFTButtons = ({
...
-      !isNFT(asset) ? onBuyItemWithCard(asset) : onExecuteOrderWithCard(asset, order)
+      !isNFT(asset) ? onBuyItemWithCard(asset, useCredits) : onExecuteOrderWithCard(asset, order, useCredits)
     },
-    [analytics, onBuyItemWithCard, onExecuteOrderWithCard]
+    [analytics, onBuyItemWithCard, onExecuteOrderWithCard, useCredits]
   )
 
   const handleBuyWithCrypto = useCallback(
-    (asset: Asset, order: Order | null) => {
+    (asset: Asset, order: Order | null, useCredits: boolean) => {
```
LautaroPetaccio, line 52: The `handleBuyWithCrypto` receives the `useCredits` as a parameter and the `handleBuyWithCard` gets it from the state. What do you think about making bot functions get the value from the state?

webapp/src/components/AssetPage/BestBuyingOption/BestBuyingOption.tsx
```diff
@@ -63,16 +69,34 @@ const BestBuyingOption = ({ asset, tableRef, isOffchainPublicNFTOrdersEnabled }:
...
+                    status: ListingStatus.OPEN,
+                    sortBy: BidSortBy.MOST_EXPENSIVE,
+                    limit: 1
+                  })
+                  .then(({ results }) => {
+                    setIsLoading(false)
+                    setMostExpensiveBid(results[0])
+                  })
+                  .catch(error => {
+                    console.error(error)
```
LautaroPetaccio, line 86: We're missing here the `setIsLoading(false)`, would you mind adding it?
```suggestion
                    console.error(error)
                    setIsLoading(false)
```

LautaroPetaccio (commented):
Great work 🥇 ! I've left some comments to review.

webapp/src/modules/order/sagas.ts
```diff
@@ -128,6 +132,11 @@ export function* orderSaga(tradeService: TradeService) {
       yield call(waitForFeatureFlagsToBeLoaded)
       const isOffchainPublicNFTOrdersEnabled: boolean = yield select(getIsOffchainPublicNFTOrdersEnabled)
       const wallet = (yield select(getWallet)) as ReturnType<typeof getWallet>
+
+      if (!wallet) {
+        throw new Error('A defined wallet is required to buy an item')
+      }
+
```
LautaroPetaccio, line 139: Maybe if we move these checks over here, we could remove the duplicated code that checks for the credits to be available, what do you think?
```suggestion
if (useCredits) {
    const isCreditsEnabled: boolean = yield select(getIsCreditsEnabled)
    if (!isCreditsEnabled) {
      throw new Error('Credits are not enabled')
    }
    const credits: CreditsResponse = yield select(getCredits, wallet?.address || '')
    if (!credits || credits.totalCredits <= 0) {
      throw new Error('No credits available')
    }
}
```
As another option, we could create a function called `getCredits` which can do this and re-use it in the transak sagas file, which does the same checks.

webapp/src/modules/transak/sagas.ts
```diff
@@ -65,81 +75,188 @@ function* handleOpenTransak(action: OpenTransakAction) {
...
+          throw new Error('Credits are not enabled')
+        }
+        contractId = CreditsManagerContractIds[Network.MATIC][asset.chainId]
+        if (!contractId) {
+          throw new Error(`Credits manager contract not found for network ${asset.network} and chainId ${asset.chainId}`)
+        }
+        const credits: CreditsResponse = yield select(getCredits, wallet?.address || '')
+        if (!credits || credits.totalCredits <= 0) {
+          throw new Error('No credits available')
+        }
```
LautaroPetaccio, line 159: This check is only done here, should we do it in the case on line `180: !isNFT(asset)`?

webapp/src/modules/transak/sagas.ts
```diff
@@ -65,81 +75,188 @@ function* handleOpenTransak(action: OpenTransakAction) {
...
+      // Legacy Marketplace
+      contractId = MarketplaceV2ContractIds[asset.network]?.[asset.chainId]
+      if (!contractId) {
+        throw new Error(`Marketplace contract not found for network ${asset.network} and chainId ${asset.chainId}`)
+      }
+      const contractName = getContractName(order.marketplaceAddress)
+      const contract = getContract(contractName, order.chainId)
+
+      if (useCredits) {
+        const isCreditsEnabled: boolean = yield select(getIsCreditsEnabled)
```
LautaroPetaccio, line 148: Following the comment on the item sagas file, this can be replaced by a common function, what do you think about it?

LautaroPetaccio (approved):
Great job! 💪 I've left just a couple of comments to review that might improve the code, but it looks great

PR #2393, opened 2025-04-07
fix: Retrieve user balance in the BuyWithCryptoModal

LautaroPetaccio:
This PR changes the BuyWithCryptoModal to use the RPCs to get the user's token balance instead of using the old API.
To do so, it uses the `InView` component from  `react-intersection-observer`, which will tell us when each token in the list is shown for the first time so we can trigger the token balance fetch.
There are two differences between the API solution:
1. We don't have the price in USD anymore => we don't get that from the balance contract call.
2. We can't order the tokens by the amount the user owns => We don't query the balance of all of the tokens at once.


juanmahidalgo (approved):
LGTM!! great job 🚀

PR #2400, opened 2025-04-20
feat: add credits wallets variant support

juanmahidalgo:


kevinszuchet (approved):

webapp/src/modules/features/selectors.ts
```diff
@@ -85,10 +91,9 @@ export const getIsOffchainPublicItemOrdersEnabled = (state: RootState) => {
 }
 
 export const getIsCreditsEnabled = (state: RootState) => {
-  if (hasLoadedInitialFlags(state)) {
-    return getIsFeatureEnabled(state, ApplicationName.MARKETPLACE, FeatureName.CREDITS)
-  }
-  return false
+  const wallet = getWallet(state)
```
kevinszuchet, line 94: I'd return false if the wallet is undefined or the address is falsy. WDYT?

PR #2421, opened 2025-06-10
[WIP] feat: Use Unity Wearable Preview

cyaiox:


cyaiox:
This approach increases the CPU and memory usage over time. We are closing this PR because we implemented a [basic but functional version](https://github.com/decentraland/marketplace/pull/2426) and optimized the loading times in the WearablePreview project.

PR #2426, opened 2025-07-08
feat: Add Unity wearable preview

cyaiox:


webapp/package.json
```diff
@@ -107,7 +107,7 @@
     "preview": "vite preview",
     "prebuild": "node scripts/prebuild.cjs",
     "build": "tsc && node --max_old_space_size=8192 ./node_modules/vite/bin/vite.js build",
-    "test": "jest",
+    "test": "jest --maxWorkers=2",
```
LautaroPetaccio, line 110: Would you mind reverting this to use all the workers?

webapp/src/modules/features/selectors.ts
```diff
@@ -105,3 +105,10 @@ export const getIsCreditsSecondarySalesEnabled = (state: RootState) => {
   }
   return false
 }
+
+export const getIsUnityWearablePreviewEnabled = (state: RootState) => {
```
LautaroPetaccio, line 109: Would you mind adding a test for this selector?

LautaroPetaccio (approved):

webapp/src/modules/features/selectors.ts
cyaiox: sure!

PR #2437, opened 2025-07-30
feat: Remove connected router from current NFT and items

LautaroPetaccio:
This PR removes the need of using the connected react router when getting the nfts and items related to the current path.


kevinszuchet (approved):
Great work 🚀

PR #2506, opened 2025-10-03
fix: min price MANA when credits toggle is ON

juanmahidalgo:
image


kevinszuchet:
Closes https://github.com/decentraland/core-team/issues/97

PR #2525, opened 2025-11-05
fix: Remove off chain feature flag from the second set of components

LautaroPetaccio:
This PR removes the off chain feature flag from the SaleRentActionBox, TransactionHistory, YourOffer, SearchBarDropdown, Bid, BidModal and BidPage components.


juanmahidalgo (approved):
LGTM!

PR #2527, opened 2025-11-06
fix: Remove off chain feature flags from analytics, bid, favorites, item and nft

LautaroPetaccio:
This PR removes the off chain feature flags from the analytics, bid, favorites, item and nft modules.

### DIFF
--- a/webapp/src/modules/bid/sagas.spec.ts
+++ b/webapp/src/modules/bid/sagas.spec.ts
@@ -495,7 +495,6 @@ describe('when handling the fetching of bids by asset', () => {
 
       it('should dispatch an action signaling the success of the action handling', () => {
         return expectSaga(bidSaga, bidServiceMock, tradeService)
-          .provide([])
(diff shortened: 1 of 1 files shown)


webapp/src/modules/bid/sagas.spec.ts
```diff
@@ -554,58 +512,7 @@ describe('when handling the fetching of bids by asset', () => {
...
-      })
-
-      it('should dispatch an action signaling the failure of the action handling', () => {
-        return expectSaga(bidSaga, bidService, tradeService)
-          .provide([
-            [select(getIsBidsOffChainEnabled), false],
-            [call([VendorFactory, 'build'], asset.vendor), vendor],
-            [call([vendor.bidService!, 'fetchByNFT'], asset), Promise.reject(error)]
-          ])
+          .provide([])
```
juanmahidalgo: can we remove this?

LautaroPetaccio: Done!

PR #2543, opened 2025-12-04
feat: introduce use of credits for NAMEs

juanmahidalgo:
image

### DIFF
--- a/webapp/src/modules/ens/sagas.spec.ts
+++ b/webapp/src/modules/ens/sagas.spec.ts
@@ -5,6 +5,7 @@ import * as matchers from 'redux-saga-test-plan/matchers'
 import { ChainId } from '@dcl/schemas'
 import { CreditsService } from 'decentraland-dapps/dist/lib/credits'
 import { getSigner, getConnectedProvider } from 'decentraland-dapps/dist/lib/eth'
+import { CreditsClient } from 'decentraland-dapps/dist/modules/credits/CreditsClient'
 import { getCredits } from 'decentraland-dapps/dist/modules/credits/selectors'
 import { CreditsResponse } from 'decentraland-dapps/dist/modules/credits/types'
 import { closeModal } from 'decentraland-dapps/dist/modules/modal/actions'
@@ -13,7 +14,7 @@ import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
 import { sendTransaction } from 'decentraland-dapps/dist/modules/wallet/utils'
 import { Route, AxelarProvider } from 'decentraland-transactions/crossChain'
 import { Provider } from 'decentraland-connect'
-import signedFetch, { AuthIdentity } from 'decentraland-crypto-fetch'
+import { AuthIdentity } from 'decentraland-crypto-fetch'
 import { ContractData, ContractName, getContract } from 'decentraland-transactions'
 import { config } from '../../config'
 import { DCLRegistrar__factory } from '../../contracts/factories'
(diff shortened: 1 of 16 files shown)


webapp/src/modules/ens/sagas.ts
```diff
@@ -148,4 +161,120 @@ export function* ensSaga() {
...
+        throw new Error('No credits available')
+      }
+
+      // Get route from backend using signed fetch
+      const identity: AuthIdentity | null = yield select(getCurrentIdentity)
+      if (!identity) {
+        throw new Error('No identity available for signed fetch')
+      }
+
+      const creditsServerUrl = config.get('CREDITS_SERVER_URL')
```
LautaroPetaccio: We have a `CreditsClient` in `decentraland-dapps` would you mind eventually adding this request to it?

webapp/src/modules/ens/sagas.ts
```diff
@@ -148,4 +161,120 @@ export function* ensSaga() {
...
+
+      const response: Response = yield call(() =>
+        signedFetch(url, {
+          method: 'GET',
+          identity: identity
+        })
+      )
+
+      if (!response.ok) {
+        const errorData: { error: string; message: string } = yield call(() => response.json())
```
LautaroPetaccio: Following the above comment, the correct way to perform a call would be to do it over the response object:
```suggestion
        const errorData: { error: string; message: string } = yield call([response, 'json')
```

webapp/src/modules/ens/sagas.spec.ts
```diff
@@ -316,4 +320,67 @@ describe('ENS Saga', () => {
...
+          .provide([
+            [select(getWallet), mockWallet],
+            [select(getCredits, mockWallet.address), mockCredits]
+          ])
+          .put(claimNameWithCreditsFailure(mockName, 'No credits available'))
+          .dispatch(claimNameWithCreditsRequest(mockName))
+          .silentRun()
+      })
+    })
+
```
LautaroPetaccio: Would you mind adding more tests that exercise the full path of the new saga?

webapp/src/modules/ens/sagas.ts
```diff
@@ -148,4 +161,120 @@ export function* ensSaga() {
...
+        PRICE_IN_WEI,
+        credits.credits,
+        ChainId.MATIC_MAINNET,
+        routeData.externalCall,
+        routeData.customExternalCallSignature
+      )
+
+      // Poll Squid Router API to wait for cross-chain completion
+      // Dispatch transaction submitted action and redirect to SuccessPage (loading state)
+      yield put(claimNameWithCreditsTransactionSubmitted(name, wallet.address, ChainId.MATIC_MAINNET, txHash))
```
LautaroPetaccio: If I'm not wrong, the `claimNameWithCreditsSuccess` is redirecting the user to the success page. This action should only store the activity of the transaction. Is that what was expected from it?

webapp/src/components/Modals/BuyWithCryptoModal/hooks.ts
```diff
@@ -380,7 +393,6 @@ const useCrossChainRoute = (
         setRoute(route)
       }
     } catch (error) {
-      console.error('Error while getting Route: ', error)
```
LautaroPetaccio: Can we keep this log? Just in case there's an error we want to debug

LautaroPetaccio (approved):

PR #2550, opened 2025-12-15
feat: add auth modal in mint name with credits flow

juanmahidalgo:


webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
```diff
@@ -340,10 +340,16 @@ export const BuyWithCryptoModal = (props: Props) => {
     )
   }, [isFetchingBalance, isBuyingAsset, asset.chainId, isLoadingAuthorization, onBuyWithCard, handleBuyWithCard, onGetMana, onClose])
 
+  const onPayWithCredits = useCallback(() => {
+    if (onBuyWithCredits) {
+      onBuyWithCredits(BigInt(finalPrice))
+    }
+  }, [onBuyWithCredits, finalPrice, credits])
```
LautaroPetaccio: Credits isn't being used here, should we keep it as a dependency?
```suggestion
  }, [onBuyWithCredits, finalPrice])
```

LautaroPetaccio (approved):
Great catch! 🎖️

webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
juanmahidalgo: good catch!

PR #2578, opened 2026-01-30
feat: add emote preview to catalog card

juanmahidalgo:


nachomazzara (commented):
## Review: feat: add emote preview to catalog card

**CI:** All checks passing ✅ | Vercel preview deployed ✅

Overall nice architecture — context provider + portal is clean, `memo` on `PreviewContent` prevents WearablePreview remounts, and the optional hook pattern for gradual adoption is smart.

### Issues to address:

**1. Console.log statements in production code** — There are ~10 `console.log` calls across `EmoteHoverPreviewContext.tsx` and `AssetCard.tsx` that look like debug leftovers. These should be removed before merging.

**2. Race condition with inner setTimeout in `hidePreview`** — The inner `setTimeout` (150ms delay to clear `activeAsset`) is NOT tracked in a ref. If `showPreview` fires between the outer timeout and this inner one, it will wipe the newly set asset. Fix: store it in a ref and clear it in `clearTimeouts`.

**3. `contextValue` not memoized** — The context value object is recreated every render, causing all consumers to re-render unnecessarily. Wrap in `useMemo`.

**4. Dead CSS** — `transition: none;` immediately followed by `transition: height 0.3s !important;` in the `.emote-preview-active .AssetImage.catalog` block. First line is dead code.

**5. Duplicated logic in `handleMouseLeave`** — It re-implements `isAssetBeingPreviewed` inline instead of calling the context function.

### Config changes bundled in (worth noting):

- `vercel.json`: env var renamed `VITE_REACT_APP_DCL_DEFAULT_ENV` → `VITE_DCL_DEFAULT_ENV` AND value changed from `dev` → `prod` — is this intentional for production builds?
- `vite.config.ts`: `manualChunks` removed — could impact bundle sizes
- `NODE_OPTIONS=--max-old-space-size=8192` added — looks like a separate build fix
- `sourcemap` now also disabled when `VERCEL` env var is set

These config changes seem like separate concerns from the emote preview feature — consider splitting into a separate PR for cleaner history.

Requested by Ignacio Mazzara via Slack

PR #2647, opened 2026-06-09
feat: simplified mobile IAP view for Godot explorer integration

sdilauro:
## Summary

This PR introduces a simplified marketplace view designed for in-app purchases (IAP) from the Decentraland Godot explorer mobile client. The view is activated via the `?view=mobile-iap` query parameter and strips down the marketplace UI to a focused, purchase-only experience.

**All changes are gated behind the `view=mobile-iap` query param — the standard marketplace is completely unaffected.**

### What changes

**Navigation & Layout**
- Only the Collectibles tab is visible (Land, Names, My Assets, My Lists, Activity are hidden)
- Navbar shows only the DCL logo (non-clickable), credits balance, and profile picture (non-interactive)
- Footer shows only language selector and copyright (no nav links, no social links)
- Intercom chat widget is hidden

**Filters**
- Network → All networks (locked)
- Status → On Sale (locked)
- Special Filters → "Get with Credits" forced ON (locked)
- Price → All prices (locked)
- Other filters (Rarity, Collection, Creators, Body Shape, etc.) remain fully interactive

**Pricing Display**
- Cards and asset detail page show prices with the credits icon instead of MANA
- No USD/fiat conversion shown in IAP mode
- Color: #a0abff (DCL Periwinkle)

**Purchase Flow**
- Single "Checkout" button paying with credits only
- Button disabled when credits balance is insufficient
- Confirm purchase modal shows credits-only pricing (no network selector, no token selector, no fees)
- "Make Offer" (bid) is hidden
- Buy with Card and Buy with MANA buttons are hidden

**Navigation Restrictions**
- The `view=mobile-iap` param persists across all navigation automatically
- `withCredits=true` and `status=on_sale` are forced in every URL
- Profile/account links are disabled (no navigation outside the marketplace flow)
- Only `/browse`, `/contracts/`, `/collections/`, and `/buy` routes are allowed

### Technical approach

- `useIsIAP()` hook reads the `view` query param via `useLocation()`
- History wrapper in `store.ts` injects params and restricts navigation
- CSS overrides for library components (navbar nav links, footer links) use proper specificity or `!important` only where emotion CSS-in-JS requires it
- No new dependencies added

### How to test

1. Normal marketplace: `http://localhost:5174/` — should be completely unchanged
2. IAP view: `http://localhost:5174/?view=mobile-iap` — simplified view with credits-only flow

## Test plan

- [ ] Verify normal marketplace (`/`) is unaffected by changes
- [ ] Verify IAP view (`/?view=mobile-iap`) redirects to `/browse`
- [ ] Verify only Collectibles tab is visible in IAP
- [ ] Verify navbar: logo non-clickable, no nav links, profile non-interactive
- [ ] Verify footer: only language selector and copyright
- [ ] Verify locked filters (Network, Status, Price, Special) are visible but disabled
- [ ] Verify interactive filters (Rarity, Collection, etc.) work normally
- [ ] Verify prices show credits icon instead of MANA
- [ ] Verify checkout button disabled when insufficient credits
- [ ] Verify confirm purchase modal shows credits-only pricing
- [ ] Verify no navigation to account/profile pages
- [ ] Verify `view=mobile-iap` param persists across navigation
- [ ] CI tests pass


sdilauro:
**Note on CI lint failure:** The `import/order` error in `SelectedFilters.spec.tsx` is pre-existing on master and unrelated to this PR — we did not modify that file. You can verify with:

```
git diff master -- webapp/src/components/AssetTopbar/SelectedFilters/SelectedFilters.spec.tsx
```

webapp/src/modules/store.ts
```diff
@@ -42,6 +42,61 @@ export const createHistory = () => {
...
+      }
+      return { ...path, search: `?${params.toString()}` }
+    }
+
+    const originalPush = history.push.bind(history)
+    const originalReplace = history.replace.bind(history)
+
+    // In IAP mode, only allow navigation to browse, item detail, and collection pages
+    const isAllowedIAPRoute = (path: string | Location): boolean => {
+      const pathname = typeof path === 'string' ? path.split('?')[0] : path.pathname
```
regenesis-claw, line 73: 🔴 **BLOCKER** — The regex matches `contracts/` but not `collections/`. The PR description lists `/collections/` as an allowed route, and `locations.collection()` returns `/collections/{addr}`. Navigating to a collection page in IAP mode will be silently blocked.

```suggestion
      return /^\/(browse|contracts\/|collections\/|buy|success)/.test(pathname) || pathname === '/'
```

webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
```diff
@@ -393,6 +399,21 @@ export const BuyWithCryptoModal = (props: Props) => {
...
   const renderMainActionButton = useCallback(() => {
+    if (isIAP) {
+      const isBuying = isBuyingAsset || isLoadingAuthorization
+      return (
+        <Button
+          fluid
+          primary
+          disabled={!wallet || isBuying}
+          loading={isBuying}
+          onClick={onBuyWithCredits ? onPayWithCredits : onBuyNatively}
```
regenesis-claw, line 424: 🟡 **RISK** — `onBuyWithCredits` is only provided for ENS names (via `MintNameWithCryptoModal`). For wearable/emote purchases, this is `undefined`, so the fallback is `onBuyNatively`. This currently works because `onBuyNatively` respects `useCredits=true` through the authorization flow, but the implicit fallback is fragile.

Consider adding a comment here explaining the intent, e.g.:
```js
// In IAP mode, onBuyWithCredits is only available for ENS names.
// For items/NFTs, onBuyNatively handles credits via the useCredits flag.
```

webapp/src/index.css
```diff
@@ -25,6 +25,61 @@ code {
...
+  z-index: 0 !important;
+}
+.iap-navbar button[aria-label$="credits"] {
+  position: relative;
+  z-index: 1 !important;
+}
+
+/* IAP: hide unsupported network modal and its dimmer (auto-switch handles it silently).
+   Target tiny modals without the buyWithCryptoModal class to avoid hiding the checkout modal. */
+body:has(.iap-navbar) .modals.dimmer:has(.ui.tiny.modal:not([class*='buyWithCryptoModal'])) {
```
regenesis-claw, line 55: 🟡 **RISK** — This hides *all* tiny modals except BuyWithCryptoModal in IAP mode. Any future tiny modal would be silently suppressed. Consider targeting the specific unsupported-network modal class (e.g. `[class*='UnsupportedNetworkModal']`) instead of a blanket exclusion.

webapp/src/components/Modals/BuyWithCryptoModal/BuyWithCryptoModal.tsx
```diff
@@ -81,6 +83,10 @@ export const BuyWithCryptoModal = (props: Props) => {
   const manaAddressOnAssetChain = getContract(ContractName.MANAToken, asset.chainId).address
   const abortControllerRef = useRef(new AbortController())
 
+  const isIAP = useIsIAP()
+  // In IAP mode, show the original asset price (not adjusted by credits)
+  const displayPrice = isIAP && 'price' in asset ? asset.price : price
```
regenesis-claw: 🟢 **NIT** — For NFTs, `'price' in asset` is false, so `displayPrice` falls back to `price` (the credits-adjusted value, possibly '0'). If secondary NFT sales are expected in IAP mode, the total would show "0" instead of the original order price. If secondary sales aren't expected, consider adding a comment noting that.

regenesis-claw (changes requested):
## IAP Gate Audit — Thorough Review

I've reviewed every file in this PR against the invariant: **nothing leaks into the default flow when `view=mobile-iap` is not active**. Below is the exhaustive assessment.

---

### ✅ Gate mechanism is sound

The two-layer approach is well designed:
1. **`store.ts` history wrapper** — injects `view`, `withCredits`, `status` params and restricts navigation at the router level. This is the hard gate.
2. **`useIsIAP()` hook** — reads `view` param from `useLocation()` for component-level conditional rendering. Properly memoized.

Every UI component change correctly gates behind `isIAP`. The hook is consistently used across all 17 modified component files. In non-IAP mode, all code paths remain identical to master.

---

### Findings

#### 🔴 BLOCKER — Route allowlist missing `/collections/` path

`store.ts` line 80:
```js
return /^\/(browse|contracts\/|buy|success)/.test(pathname) || pathname === '/'
```

The PR description explicitly lists `/collections/` as an allowed route, but the regex only matches `contracts/` (which covers item/NFT detail pages like `/contracts/{addr}/items/{id}`). Collection browsing pages at `/collections/{addr}` (generated by `locations.collection()`) are **silently blocked** — navigation is swallowed with no feedback.

**Fix:** Add `collections\/` to the regex:
```js
return /^\/(browse|contracts\/|collections\/|buy|success)/.test(pathname) || pathname === '/'
```

---

#### 🟡 RISK — `BuyWithCryptoModal` IAP button falls through to `onBuyNatively` when `onBuyWithCredits` is absent

`BuyWithCryptoModal.tsx` line ~410:
```js
onClick={onBuyWithCredits ? onPayWithCredits : onBuyNatively}
```

For wearable/emote purchases (both primary via `MintNftWithCryptoModal` and secondary via `BuyNftWithCryptoModal`), `onBuyWithCredits` is not passed — it's only provided for ENS names. The fallback to `onBuyNatively` **does** correctly handle credits through the authorization flow (`onBuyItem(item, useCredits=true)` / `onExecuteOrder(..., useCredits=true)`), so this works. However, the implicit fallback is fragile — a future refactor that changes `onBuyNatively`'s credits handling would silently break IAP purchases with no test to catch it.

**Suggestion:** Add a comment explaining the fallback is intentional, or explicitly pass `onBuyWithCredits` for all IAP purchase types.

---

#### 🟡 RISK — CSS modal hiding is overly broad

`index.css`:
```css
body:has(.iap-navbar) .ui.tiny.modal:not([class*='buyWithCryptoModal']) {
  display: none !important;
}
```

This hides **all** tiny modals except BuyWithCryptoModal when IAP is active. Any future tiny modal (error dialogs, confirmations) would be silently suppressed. Consider targeting the specific unsupported-network modal class instead of using a blanket exclusion.

---

#### 🟡 RISK — `hideSocialLinks` prop may not exist on `FooterProps`

`PageLayout.tsx` passes `hideSocialLinks={isIAP}` to `Footer`, but the component types come from `FooterProps` (decentraland-ui v7.1.0). If this prop doesn't exist on the interface, React silently ignores it. The CSS fallback (`.iap-footer .main-footer .links { display: none }`) works regardless — confirm the prop exists or remove it and rely on CSS alone.

---

#### 🟡 RISK — No test coverage for IAP-specific code paths

Coverage dropped 0.4%. Zero new tests cover the IAP checkout button, IAP sign-in button, success page deep link, modal simplified rendering, or history wrapper route blocking. These are high-value integration paths that should have unit tests.

---

#### 🟢 NIT — `displayPrice` type narrowing gap for NFTs

`BuyWithCryptoModal.tsx` line 88:
```js
const displayPrice = isIAP && 'price' in asset ? asset.price : price
```

For NFTs (no `.price` property), this falls back to `price` (the credits-adjusted value, possibly '0'). If secondary NFT sales are expected in IAP mode, the modal total would show "0" instead of the order price.

---

### Leak analysis — non-IAP flow

| Area | Leaks? | Notes |
|------|--------|-------|
| Routes/sections | ✅ No | `!isIAP &&` guards only suppress when active |
| ENS category | ✅ No | Hidden via `!isIAP && shouldRenderSection(...)` |
| Modal price display | ✅ No | `isIAP` ternary, non-IAP path unchanged |
| Network switching | ✅ No | `useIAPAutoSwitchNetwork` exits early when `!isIAP` |
| Success page | ✅ No | Standard buttons wrapped in `{!isIAP && ...}` |
| Bid buttons | ✅ No | `!isIAP` guard in BuyNFTBox, ItemSaleActions |
| Navbar | ✅ No | IAP returns early; non-IAP falls through to original |
| Footer | ✅ No | `iap-footer` class only applied when `isIAP` |
| Filters | ✅ No | `isIAP ? locked : original` throughout |
| Asset cards | ✅ No | `isIAP ? CreditsPrice : Mana` ternary |
| History wrapper | ✅ No | Only installed when `viewParam` exists |
| Favorites counter | ✅ No | `!isIAP &&` guard |
| Detail tables | ✅ No | Offers/Bids/History/Listings hidden behind `!isIAP` |

**The non-IAP experience stays completely unchanged.** Every gate is properly scoped.

---

### CI Status
- Lint ✅, Test ✅, Audit ✅, Title convention ✅
- Vercel preview ❌ FAILED (pre-existing issue per author's comment)

Requested by Lean via Slack

sdilauro (commented):
Thanks for the thorough review! Addressing each finding:

**P1 — `/collections/` in route allowlist**: Intentionally excluded. In IAP mode we want users focused on browse → item detail → purchase. Collection pages are outside the purchase flow. The PR description was outdated — updated the comment in `store.ts` to clarify.

**P1 — `onBuyNatively` fallback**: Comment already added in previous commit explaining the intent. `onBuyNatively` correctly handles credits via `useCredits` flag through the authorization flow.

**P1 — `displayPrice` for NFTs**: Already fixed — reconstructs original price by adding credits back to the adjusted price.

**P2 — NFT without order**: Added explicit guard — disabled checkout button when NFT has no active order.

**P2 — `hasSwitched` ref preventing retry**: Removed the ref entirely. The effect now dispatches `switchNetworkRequest` whenever `chainId !== expectedChainId`, naturally stopping when the switch succeeds.

**P2 — `viewParam` scope too broad**: Tightened the guard from `if (viewParam)` to `if (isIAP)`. The entire history wrapper now only activates for `mobile-iap`.

**P2 — Hardcoded color, duplicated patterns, fragile CSS**: Noted for future cleanup. Keeping changes minimal since this is not our project.

**`hideSocialLinks` prop**: Confirmed it exists in `decentraland-ui` Footer types (v7.1.0, line 24 of Footer.d.ts).

**Tests**: Existing PriceComponent tests updated to work with the Router context. Keeping new test additions minimal per team agreement.

regenesis-claw (approved):
## Re-review — All previous findings addressed

Re-reviewed the 4 new commits (`1d42521a`, `caf58888`, `0151463f`, `9d3896a7`). Here's the status of every item from my previous review:

---

### Previously flagged items

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 BLOCKER | Route allowlist missing `/collections/` | ✅ Resolved — clarified as *intentional* design decision (keep users in purchase flow). Comment added. PR description was misleading; code is correct. |
| 2 | 🟡 RISK | `onBuyNatively` fallback undocumented | ✅ Addressed — clarifying comment added at the `onClick` handler |
| 3 | 🟡 RISK | CSS modal hiding overly broad | ✅ Acknowledged — NOTE comment added warning future devs |
| 4 | 🟡 RISK | `hideSocialLinks` prop may not exist | ℹ️ Unchanged — CSS fallback works regardless, low concern |
| 5 | 🟡 RISK | No test coverage for IAP paths | ℹ️ Unchanged — noted but not blocking |
| 6 | 🟢 NIT | `displayPrice` wrong for NFTs | ✅ Fixed — proper `useMemo` reconstruction adding credits back to adjusted price |

### New changes review

#### `store.ts` — guard tightened ✅
Wrapper now activates on `if (isIAP)` instead of `if (viewParam)`. Cleaner — only `mobile-iap` triggers param injection and route restrictions. The `injectParams` function also simplified: hardcodes `'mobile-iap'` instead of using the variable, removing the generic `viewParam` abstraction. Good.

#### `PageLayout.tsx` — `hasSwitched` ref removed ✅
The auto-switch effect now retries on each dep change without a one-shot guard. Since `useEffect` deps are `[isIAP, walletConnected, chainId, dispatch]`, it naturally stops once `chainId` matches. If the user rejects the wallet prompt, `chainId` stays wrong but deps don't change, so no infinite loop. On navigation (remount), it retries — reasonable behavior.

#### `BuyNFTButtons.tsx` — NFT without order guard ✅
New explicit disabled checkout button when `isNFT(asset) && !order`. Good edge case — prevents crash on `order!.price` and gives clear UX. The non-null assertion `order!.price` after the guard is correct.

#### `index.css` — credits tooltip hidden ✅
Credits button gets `pointer-events: none` and `.credits-tooltip` hidden. Prevents tooltip interaction in mobile WebView.

---

### Leak analysis — re-confirmed

All non-IAP paths remain unchanged. The tighter `if (isIAP)` guard in `store.ts` actually *reduces* the surface area compared to the previous `if (viewParam)` version. No new leak vectors introduced.

### CI Status
- Audit ✅, Title convention ✅
- Lint, Test: pending (just pushed)
- Vercel preview: ❌ (pre-existing)

Approving. The remaining items (test coverage, `hideSocialLinks` prop) are improvements, not blockers.

Requested by Lean via Slack

webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyNFTButtons.module.css
leanmendoza, line 1: revert this file?

leanmendoza (approved):
Tested and working as expected locally.
Looks good to me, great job 👏

juanmahidalgo (approved):
LGTM!

PR #2670, opened 2026-08-05
fix: accept usd-pegged trades instead of signing an empty value

juanmahidalgo:
Every USD-pegged listing is unbuyable from this app. Not mispriced — **unbuyable**.

`getValueForTradeAsset` had no case for `TradeAssetType.USD_PEGGED_MANA`, so it fell through to `default`,
logged "Invalid asset type" and returned `''`.

That value is not cosmetic. `generateTradeValues` feeds it into the EIP-712 struct, and `getOnChainTrade`
rebuilds that struct **client-side** to hand to the contract's `accept()`:

```
getOnChainTrade(trade, buyer) → generateTradeValues(trade) → getValueForTradeAsset(asset)   // utils/trades.ts:103,110
```

So the seller signs a trade whose received value is `600000000000000000`, and the buyer submits one whose
received value is `''`. Different struct, different trade hash, and the on-chain signature check rejects it.
`estimateTradeGas` runs through the same helper, so it fails first and surfaces as a generic error with nothing
pointing at the cause.

This is live right now and getting worse on its own: as creators migrate listings from MANA to credits, each
migrated listing becomes unbuyable here. The shop is unaffected — it carries its own extractor precisely
because of this gap, with a comment saying so.

## The fix

```ts
case TradeAssetType.USD_PEGGED_MANA:
  return asset.amount
```

The amount **verbatim**, exactly as for `ERC20`. It is denominated in USD wei rather than MANA wei, but the
conversion is the contract's business at settlement, not this function's — all that matters here is
reproducing the value the seller signed.

## Tests

Two, and **both fail without the fix** (verified by removing the case and re-running):

1. The unit case: a pegged asset returns its amount, not an empty value.
2. The one that matters: `getOnChainTrade` **preserves** the pegged amount when it rebuilds the struct.
   Asserted against the literal `600000000000000000` rather than through `getValueForTradeAsset` — the
   existing structural test builds its expectation by calling that same function, which is right for checking
   shape but means it would have passed no matter what the function returned. That is why this bug reached
   production with a green suite.

```
src/utils/trades.spec.ts — 8 passed, 2 skipped (pre-existing skips)
```

## Test plan

- [x] `tsc --noEmit`, prettier
- [x] `src/utils/trades.spec.ts`
- [ ] Full suite in CI
- [ ] **Buy a USD-pegged listing on `.zone` from this app** — the point of the change, and it has never
      succeeded, so there is no "still works" to regress
- [ ] Confirm a plain MANA (ERC20) listing still buys, since it shares the code path

### DIFF
--- a/webapp/src/utils/trades.spec.ts
+++ b/webapp/src/utils/trades.spec.ts
@@ -7,7 +7,8 @@ import {
   Trade,
   TradeAsset,
   TradeCreation,
-  TradeType
+  TradeType,
+  USDPeggedManaTradeAsset
 } from '@dcl/schemas/dist/dapps/trade'
 import * as ethUtils from 'decentraland-dapps/dist/lib/eth'
 import { TradeService } from 'decentraland-dapps/dist/modules/trades/TradeService'
@@ -67,11 +68,16 @@ describe('when getting the value for a trade asset', () => {
    */
   describe('and the asset is USD-PEGGED MANA', () => {
     beforeEach(() => {
-      asset = { assetType: TradeAssetType.USD_PEGGED_MANA, amount: '600000000000000000' } as ERC20TradeAsset
+      asset = {
+        assetType: TradeAssetType.USD_PEGGED_MANA,
+        contractAddress: '0xmana',
+        amount: '600000000000000000',
+        extra: '0x'
+      } as USDPeggedManaTradeAsset
     })
 
     it('should return the amount, not an empty value', () => {
-      expect(getValueForTradeAsset(asset)).toBe('600000000000000000')
+      expect(getValueForTradeAsset(asset)).toBe((asset as USDPeggedManaTradeAsset).amount)
     })
   })
 })
@@ -294,7 +300,6 @@ describe('when getting the trade to accept', () => {
     const onChainTrade = getOnChainTrade(peggedTrade, beneficiaryAddress)
 
     expect(onChainTrade.received[0].value).toBe('600000000000000000')
-    expect(onChainTrade.received[0].value).not.toBe('')
   })
 })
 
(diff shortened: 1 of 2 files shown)


juanmahidalgo:
Build fixed, plus the review's P2s.

## The build break was mine, and worth naming

`tsc` type-checks the spec files as part of `npm run build`, and my fixture cast a pegged asset as
`ERC20TradeAsset` — which does not overlap and is missing `contractAddress`/`extra`.

I did run `tsc --noEmit` before opening this, but **the marketplace had no `node_modules` in my checkout**, so
tsc resolved nothing and I filtered its output for `utils/trades` and saw a clean result. I filtered the output
of a typecheck that had not actually run. Deps installed, and everything below is verified against a real one.

## What changed

- **The fixture uses `USDPeggedManaTradeAsset`** — the dedicated type `@dcl/schemas` already exports, with the
  full `BaseTradeAsset` shape. Same conclusion as the review's P2, arrived at because the build forced it.
- **The switch is now exhaustive at compile time.** This is the part I would keep even if the fixture were
  fine: `USDPeggedManaTradeAsset` has been a member of the `TradeAsset` union all along, so this was a missing
  case in a closed union — and the `default` swallowed it instead of the compiler reporting it. A
  `const unhandled: never = asset` in the default makes the next union member a type error rather than another
  silent `''`. Runtime behaviour is unchanged on purpose: this function also receives API data, so a value
  outside the union must still degrade rather than throw.
- **Comment trimmed** to the two things a future editor needs — return the amount verbatim, and do not
  "convert" it, because the value has to reproduce what the seller signed. The narrative lives here and in the
  commit, per the P2.
- **Dropped the redundant `.not.toBe('')`.**

Verified with deps present: `tsc --noEmit` clean, prettier clean, `src/utils/trades.spec.ts` 8 passed / 2
pre-existing skips. Both new tests still fail without the fix (re-checked by removing the case).

## The remaining P2 is the one that outlives this PR

`decentraland-dapps` exports an identical `getValueForTradeAsset` with the same missing case, so **every other
consumer of the package has this bug** — the shop already carries its own extractor to route around it. That is
a separate repo and a separate release, so it belongs in its own change rather than here; filing it as a
follow-up so it does not stay folklore in three codebases.

PR #2671, opened 2026-08-05
fix: price usd-pegged listings in mana from the settlement oracle

juanmahidalgo:
#2669 stopped USD-pegged listings from rendering as fractional MANA, but it replaced them with **credits** — and
credits are the wrong unit for this app. This corrects that.

## Why credits were wrong here

1. **This app charges in MANA.** A `USD_PEGGED_MANA` trade denominates the item in USD and the marketplace
   contract converts to MANA at accept time. A buyer told "5 credits" has been informed of a unit they will
   never spend, and denied the one they will.
2. **"Credits" already means something else on this screen.** This repo has the DCL Credits *programme* — the
   weekly rewards, `icon-credits.svg`, `CreditsResponse`, the IAP branch in `AssetCard`. Two unrelated things
   under one word, one of them next to a Buy button.

So the shop-side vocabulary is gone from this repo: `lib/credits.ts`, `components/CreditsPrice`, the duplicated
peg constants and the copied glyph are all deleted. What replaces them converts to MANA.

## The oracle, and why it has to be this one

`modules/trade/manaRate.ts` reads the aggregator address **off the marketplace contract**
(`manaUsdAggregator()`), then that aggregator's `decimals()` + `latestRoundData()`.

That indirection is the point, not ceremony. `decentraland-transactions` also ships a `ChainlinkOracle` entry,
and it is a **different contract** from the one the marketplace settles with — measured on Polygon:

| | address | `getRate()` |
|---|---|---|
| `marketplace.manaUsdAggregator()` | `0xA1CbF3Fe43BC3501e3Fc4b573e822c70e76A7512` | absent (raw Chainlink: `latestRoundData()`) |
| package `ChainlinkOracle` | `0xe18B1361d41afC44658216F3Dc27e48c2336e3c2` | present |

Taking the convenient one-read option would have displayed a rate from a contract that does not settle the
trade — a number the buyer sees and a different number they pay, which is the whole class of bug this work has
been chasing. Reading it off the marketplace makes displayed and settlement rates the same by construction.

Cached 60s per chain (one read for a grid, not one per card), failures deliberately not cached, and a
non-positive answer throws rather than being divided by.

## What a buyer sees

`~7.48` behind the MANA glyph, for the $0.50 listing that used to read `0.5`, then `5 credits ($0.50)`.
At the live rate ($0.06686601/MANA) that is what $0.50 costs.

- **No tooltip.** There is no peg to explain — the price is in MANA now.
- **No dollar figure**, even where the caller passes `showFiat`. The USD amount is what the listing is pegged
  TO, not what the buyer pays, and showing both invites reading the wrong one as the price.
- **Approximate on purpose.** The rate moves between render and confirmation, so the `~` is honest: the exact
  MANA is whatever the oracle says at accept time, and if it moved far enough to fail, the buyer retries
  against a fresh quote. The conversion truncates rather than rounds up, so the shown figure never sits above
  the charge.
- **No rate, no number.** While the read is in flight or if the oracle is unreachable it says "price
  unavailable" rather than rendering a placeholder next to a Buy button.

## Also

`getChainIdByNetwork` reads app config and **throws** when that is not initialised. A leaf price component must
not be able to take a grid of ninety cards down with it, so that call is guarded and a failure resolves to the
unavailable state.

## Test plan

- [x] `tsc --noEmit`, `check:code` clean on every touched path, prettier clean
- [x] `npm test` — 159 suites / 1813 passing
- [x] `npm run build` — the step that broke on #2670, run here before pushing
- [x] `manaRate.spec.ts` (10 cases) pins: the aggregator read going through the marketplace; the TTL reuse; a
      failure not being cached; a non-positive rate rejected; and the conversion — including the production
      case ($0.50 → 7477640732563525175 wei) and $1M staying exact past the float range, asserted against
      independently computed values rather than the function's own output
- [x] `ListingPrice` / `PriceComponent` now assert `~7.48` / `~8.97`, no `$` anywhere in the pegged branch, and
      that a plain MANA listing keeps its fiat line
- [ ] A pegged listing checked by eye on `.zone`

## Not fixed here, and it needs the API

Browse **catalog cards** still render the raw amount. `/v1/catalog` rows carry neither a denomination flag nor a
`tradeId`, so a card cannot know it is looking at a pegged listing — and `minPrice` / the listing range are
cross-listing aggregates, so a mixed set cannot be one price in one unit at all. That is a marketplace-server
change (denominate the price, or send the `tradeId`), not a front-end one.


juanmahidalgo:
All four P2s applied in `d1ead91d`.

## The dev file was my mistake, and it is worth naming

`webapp/src/config/env/across.code-workspace` is not from this change — it is a local editor workspace that was
sitting untracked in the checkout, and I swept it in by running `git add webapp/src` on a **directory** instead
of explicit paths. That is exactly what the repo convention against directory-wide staging exists to prevent,
and it would have committed one developer's sibling-checkout paths for everyone.

Removed from the branch with `git rm --cached`, so the file stays on disk locally, and `*.code-workspace` is now
in `.gitignore` so the next `add` cannot pick it up either. This commit stages explicit paths only.

## The three code suggestions

**Stale rate on a chain switch** — applied. `setRate(null)` before the fetch, so the previous chain's rate can
never briefly price a listing under another network's glyph.

**The unchecked tuple cast** — applied, and I went a step further than the suggestion. Rather than casting
`round[1]`, it now reads the index defensively and **throws** when the tuple carries no rate:

```ts
const raw: unknown = Array.isArray(round) ? round[1] : undefined
if (raw === null || raw === undefined) {
  throw new Error('The MANA/USD oracle returned no rate')
}
const answer = BigInt(String(raw))
```

The reasoning: the result of this read *is a price*. If an ethers major or a different feed ever changes the
decode shape, failing loudly is right and a cast that quietly yields something unusable is not.

**The divisor guard in `usdWeiToManaWei`** — applied. You are right that the function being exported is what
makes it reachable: `fetchManaUsdRate` rejects a non-positive answer, but a hand-built rate would have thrown a
`RangeError` out of a render. It returns `null` now, which the component already renders as "price unavailable".

Both new guards have tests (a tuple with no rate at index 1; `answer: 0n` and `answer: -1n`), so they are
covered rather than merely present.

`tsc --noEmit` clean, lint clean on every touched path, **159 suites / 1815 tests** passing.

PR #2677, opened 2026-08-10
fix: stop offering credits the marketplace cannot spend [ON HOLD]

juanmahidalgo:
The navbar still shows a credits chip, and hovering it reads:

> Expiring in 0 days
> (1 Credit = 1 MANA in value)

Those are the **legacy MANA-denominated credits** — a programme this app no longer issues or spends. The figure is at best zero, and at worst an offer this surface cannot honour.

## Changes

`withCredits={false}` on both navbar variants (the standard one and the iOS in-app view, which took the same default).

The marketplace never set this prop, and `decentraland-dapps` defaults it to **true** — so the chip was not a decision anyone made here, it was a default nobody turned off.

## One thing to be aware of before merging

`withCredits` is the only switch dapps exposes, and it gates both balances:

```ts
creditsBalance:     withCredits ? creditsBalance : undefined,
shopCreditsBalance: withCredits ? shopCreditsBalance : undefined,
```

So this also hides the **Shop's USD credits** from this navbar. I believe that is right for the same reason — they are not spendable here either, and a balance shown where it cannot be used is the same false offer in a newer denomination.

If the intent is to keep Shop credits visible in the marketplace while dropping only the legacy ones, this flag cannot express that and the change belongs in `decentraland-dapps` as a separate prop instead. Say the word and I will take it there.

## Test plan

- [x] New `Navbar.spec.tsx` — the component had none — asserting `withCredits: false` reaches `Navbar2` in both the standard and in-app variants
- [x] **Mutation-checked**: dropping the prop from either variant fails its case. This matters more than usual here, because the dapps default is `true` — a future edit that removes the prop silently brings the chip back rather than erroring
- [x] typecheck clean; 2 tests pass
- [ ] full suite and build not run locally (machine thermals); CI is the first full run

### DIFF
--- a/webapp/src/components/Navbar/Navbar.spec.tsx
+++ b/webapp/src/components/Navbar/Navbar.spec.tsx
@@ -1,0 +1,48 @@
+import { render } from '@testing-library/react'
+import { MemoryRouter } from 'react-router-dom'
+import Navbar from './Navbar'
+
+const baseNavbar = jest.fn((_props: Record<string, unknown>) => null)
+jest.mock('decentraland-dapps/dist/containers/Navbar', () => ({
+  Navbar2: (props: Record<string, unknown>) => baseNavbar(props)
+}))
+
+const useIsIAP = jest.fn(() => false)
+jest.mock('../../modules/iap/useIAP', () => ({ useIsIAP: () => useIsIAP() }))
+
+function renderNavbar() {
+  return render(
+    <MemoryRouter>
+      <Navbar />
+    </MemoryRouter>
+  )
+}
+
+beforeEach(() => {
+  baseNavbar.mockClear()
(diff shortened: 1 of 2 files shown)


juanmahidalgo:
**ON HOLD — do not merge.**

There are still users holding the legacy MANA-denominated credits, so hiding the chip would take away the only place they can see a balance they actually own. Hiding it is right eventually; it is not right yet.

Parked as a draft rather than closed, so the work and its test survive.

**To resume, the open question is the one in the PR body:** `withCredits` gates the legacy balance and the Shop's USD balance together, so it cannot express "drop the old, keep the new". Whenever this is picked up, decide first whether legacy holders have been migrated or drained — and if the answer is "some still hold them", the change belongs in `decentraland-dapps` as a separate prop rather than here.

PR #2679, opened 2026-08-14
feat: load segment's analytics.js from a first party proxy in dev

nachomazzara:
## What

Loads Segment's `analytics.js` from our first party proxy instead of `cdn.segment.com`, **in dev only**.

- Bumps `decentraland-dapps` to `29.4.0`, which is where the `analyticsUrl` option comes from (decentraland/decentraland-dapps#814).
- Adds `SEGMENT_ANALYTICS_URL` to `webapp/src/config/env/dev.json` and passes it to `createAnalyticsMiddleware`.
- `stg` and `prod` have no URL configured, so they keep loading the bundle from Segment's CDN until they get their own proxy path.

## Why

Ad blockers drop the requests to `cdn.segment.com`, so those sessions send no events at all. The proxy serves the same bundle from a first party host with an obfuscated path.

## Notes

- The write key is untouched: each environment keeps writing to its own source through `SEGMENT_API_KEY`.
- `analytics.js` fetches its settings from the origin of the URL (`https://evs.e.decentraland.org`), which the dapps side derives on its own. No extra config needed here.
- Upstream ignores any value that is not a valid URL, or that is not HTTPS unless it belongs to the dapp's own origin, falling back to Segment's CDN. So a misconfigured environment degrades to today's behavior instead of loading a script from an untrusted origin.
- Worth tracking separately: whoever controls the proxy host controls script execution on the dapp, and the webapp has no CSP `script-src` allowlist today to constrain it.

## Test

`tsc`, eslint and the test suite pass against the bumped dependency. Once deployed to dev, the check is that the bundle request goes to `evs.e.decentraland.org` and the events land in the dev source.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -21,7 +21,7 @@
     "dcl-catalyst-commons": "^9.0.1",
     "decentraland-connect": "^12.0.0",
     "decentraland-crypto-fetch": "^2.0.1",
-    "decentraland-dapps": "^29.3.1-20260805170817.commit-f4ffa6a",
+    "decentraland-dapps": "^29.4.0",
     "decentraland-transactions": "^3.0.3",
     "decentraland-ui": "^7.1.0",
     "decentraland-ui2": "^3.8.0",
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -26,7 +26,7 @@
         "dcl-catalyst-commons": "^9.0.1",
         "decentraland-connect": "^12.0.0",
         "decentraland-crypto-fetch": "^2.0.1",
-        "decentraland-dapps": "^29.3.1-20260805170817.commit-f4ffa6a",
+        "decentraland-dapps": "^29.4.0",
         "decentraland-transactions": "^3.0.3",
         "decentraland-ui": "^7.1.0",
         "decentraland-ui2": "^3.8.0",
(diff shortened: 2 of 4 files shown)


nachomazzara:
Both findings match what the PR already describes, no changes needed here:

**P1** — Right, that's why this is a draft. The `analyticsUrl` option lands in decentraland/decentraland-dapps#814, and dapps only publishes from `master`, so there is no snapshot to bump to yet. Sequence is: merge that PR, wait for the published version, bump `decentraland-dapps` here, undraft. Verified locally against a build of that branch: `tsc` and eslint pass with this change.

**P2** — Intentional, dev only for now. stg and prod have no `SEGMENT_ANALYTICS_URL`, so they keep loading the bundle from Segment's CDN until they get their own proxy path.

The failing Vercel deployment is the same root cause as P1 and will clear with the bump.

nachomazzara:
Re-review addressed:

**P1** — Unchanged, this is the dependency sequencing already described: merge decentraland/decentraland-dapps#814, wait for the publish, bump `decentraland-dapps` here, undraft. Nothing to fix on this side.

**P2** — Fixed in dc8c7012, `analyticsUrl` is now `config.get('SEGMENT_ANALYTICS_URL', '') || undefined`, so staging and prod pass no URL at all instead of an empty string. The upstream side no longer depends on that anyway: it validates the value and falls back to Segment's CDN for anything that is not a trusted URL.

Security wise this PR adds one public config value pointing at a first party HTTPS host, no secrets. Worth tracking separately: whoever controls that host controls script execution on the dapp, and there is no CSP `script-src` allowlist in the webapp today to constrain it.

braianj (approved):
LGTM!
/just a reminder, remember to update decentraland-dapps first

nachomazzara:
Unblocked: decentraland/decentraland-dapps#814 is merged and published as `29.3.1-20260814150128.commit-bf4b95a`, bumped here in 6714946d. Out of draft.

`tsc` is clean, eslint has no errors, and the suite passes (159 suites, 1817 tests).

Correcting a note I left earlier in this PR: the `showManaBalancesInNavbar` type error I expected from the bump does not happen. It came from my local check, where I had overlaid a build of the dapps branch whose declarations resolved against a different `decentraland-ui2` than the one this repo installs. Against the published package there is nothing to sort out.

webapp/package.json
```diff
@@ -21,7 +21,7 @@
     "dcl-catalyst-commons": "^9.0.1",
     "decentraland-connect": "^12.0.0",
     "decentraland-crypto-fetch": "^2.0.1",
-    "decentraland-dapps": "^29.3.1-20260805170817.commit-f4ffa6a",
+    "decentraland-dapps": "^29.3.1-20260814150128.commit-bf4b95a",
```
braianj: We should noy use a commit version, better the full published version

nachomazzara:
Switched to the stable release: `decentraland-dapps@^29.4.0` instead of the commit snapshot (bc1beda1).

There was no stable release carrying the change, so I cut decentraland/decentraland-dapps@29.4.0 from master. Worth knowing for future releases there: the publish only lands on `latest` when the git tag is plain semver. The `v` prefixed ones (`v29.2.0`, `v29.3.1`) made oddish publish a `*-commit-<sha>` snapshot under a `tag-v29.x.y` dist-tag instead, which is why npm `latest` had been stuck on `29.3.0` since August 3rd. This release used `29.4.0` with no prefix and published cleanly.

`tsc` clean, eslint with no errors, 159 suites / 1817 tests passing against it.

Issue #2, opened 2019-12-09
Scaffold React project


fmiras:
Add:
- [ ] Navbar
- [ ] Footer
- [ ] Blockie
- [ ] Available MANA

PR #2692, opened 2026-08-26
feat: sign and authorize against the newest off-chain marketplace

LautaroPetaccio:
Moves the webapp onto `OffChainMarketplaceV3` where it is deployed, matching decentraland-dapps#824 and marketplace-server#395.

## Why both halves had to move together

The webapp pinned V2 in two places that must name the same contract:

- the **EIP-712 domain** a listing or bid is signed against (`utils/trades.ts`)
- the **spender** the UI asks the user to approve (`BidModal`, both `SellModal`s)

Changing only the signing would have had users approve V2 while the trade settles on V3, so the allowance sits on the wrong contract and `accept()` fails. Both move here.

Note the webapp does **not** use dapps' trades lib — it has its own copy of `getTradeSignature`, `OFFCHAIN_MARKETPLACE_TYPES` and friends. So this needs no dapps bump, only `decentraland-transactions` `^3.0.3` → `^3.1.1` for `ContractName.OffChainMarketplaceV3`.

## Resolution is a fallback, not a switch

`getContract` **throws** for a version a chain does not have, and V3 is testnet-only. `getLatestOffChainMarketplaceContract` tries V3 then V2, so testnets move while mainnet keeps V2 until V3 ships there.

The Settings page now names an approval granted to *any* version, guarded the same way — without it a V3 approval falls through to the generic lookup, which does not know these addresses.

## Deliberately left alone

`getAcceptBidAuthorizationOptions` still defaults to V2 when a bid carries no `tradeContractAddress`. That fallback only fires for legacy rows predating the column, which were signed against V1/V2; pointing it at V3 would mis-authorize them. New bids carry the address and resolve through `getContractName`, which knows V3 as of 3.1.1.

## Verification

`tsc` clean, `eslint` clean, full suite **1841 passed / 21 skipped, 160 suites**. The one failure in the full run (`RentalPeriodFilter`, a 5s timeout) passes in isolation in 19ms and touches nothing here.

The existing signature test derives its expectation from the resolved contract rather than a hardcoded string, so it now proves the signature is bound to V3 on Sepolia. Three cases added for the resolver: V3 chain, V2 fallback, no marketplace at all.

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -28,7 +28,7 @@
         "decentraland-connect": "^12.0.0",
         "decentraland-crypto-fetch": "^2.0.1",
         "decentraland-dapps": "^29.7.0",
-        "decentraland-transactions": "^3.0.3",
+        "decentraland-transactions": "^3.1.1",
         "decentraland-ui": "^7.1.0",
         "decentraland-ui2": "^3.22.0",
         "ethers": "^5.6.8",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -23,7 +23,7 @@
     "decentraland-connect": "^12.0.0",
     "decentraland-crypto-fetch": "^2.0.1",
     "decentraland-dapps": "^29.7.0",
-    "decentraland-transactions": "^3.0.3",
+    "decentraland-transactions": "^3.1.1",
     "decentraland-ui": "^7.1.0",
     "decentraland-ui2": "^3.22.0",
     "ethers": "^5.6.8",
--- a/webapp/src/components/AssetCard/AssetCard.spec.tsx
+++ b/webapp/src/components/AssetCard/AssetCard.spec.tsx
@@ -4,7 +4,7 @@ import { BodyShape, ChainId, Network, NFTCategory, Rarity, WearableCategory } fr
 import { Asset } from '../../modules/asset/types'
 import { INITIAL_STATE } from '../../modules/favorites/reducer'
 import { PageName, SortBy } from '../../modules/routing/types'
-import { useTradePriceDenomination } from '../../modules/trade/hooks'
+import { useTradePricing } from '../../modules/trade/hooks'
 import { renderWithProviders } from '../../utils/test'
 import { HoverPreviewProvider } from '../HoverPreview'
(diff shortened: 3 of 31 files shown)


LautaroPetaccio:
Fixed #2 and #3 in 291fbfc2, both pinned by tests that fail when the change is reverted.

**#2 gas estimation** — falls back to `trade.contract` and throws when neither that nor the caller's address is available, so it can no longer measure a call against a contract the signature does not authorise.

**#3 credits route** — correct catch, and a regression I introduced. The registration lookup ran before the route split, so a version with no *direct* registration also blocked the credits path, which goes through the CreditsManager and never needed one. The check now lives in the direct-`accept` branch only.

**#1 I don't think is actionable, and I'd rather not "fix" it — happy to be shown wrong.**

Two reasons:

1. `TradeCreation` in `@dcl/schemas` has no `contract` field. Only the persisted `Trade` does:

```ts
export type Trade = { ...; contract: string }          // has it
export type TradeCreation = { signer; signature; network; chainId; type; checks; sent; received }   // does not
```

So the webapp cannot include it without a schemas change.

2. More importantly, the server does not want it. `marketplace-server` derives the contract by **signature recovery** rather than trusting the client (`src/ports/trades/component.ts`, merged in #395):

```ts
// Resolving it also reports WHICH marketplace version signed it, which is
// what the trade's contract and EIP-712 digest have to be recorded from
const signatureMatch = resolveTradeSignature(trade, signer)
if (!signatureMatch) throw new InvalidTradeSignatureError()
...
getInsertTradeQuery({ ...trade, contract: signatureMatch.contract.address, tradeDigest: signatureMatch.cancellationDigest }, signer)
```

`resolveTradeSignature` tries each candidate marketplace's EIP-712 domain and returns the one the signature actually verifies against. Persisting a client-supplied `contract` would be strictly weaker: a client could assert a version the signature does not correspond to, and every downstream consumer this PR added — the Transak registry, the ABI, the oracle, `TradeService.accept`/`cancel` — would then trust it.

The premise that "the rest of this PR relies on persisted `trade.contract`" is right; the field is just already populated, and from the signature rather than from the caller.

PR #2709, opened 2026-09-11
fix: bind the mobile-iap checkout to the payment source it presents

juanmahidalgo:
## Changes

- Recognise the `/marketplace` basename the way `history` does before the router sees the path — case-insensitively, and only at a path boundary. The redirect in `shopRedirect` and the router disagreed, so a case variant of the basename resolved to the route while skipping the hand-off to the shop. Fixed paths now also tolerate a trailing slash, which `ITEM_PATH` already did.
- Open the checkout a `buyWithCrypto=true` deep link asks for from an effect, once per mount, through the same handler the button uses. It ran from the render body and called the dispatch prop directly, so it re-dispatched on every render and dropped the credits selection the handler carries.
- Pass the resolved pre-credits MANA amount down as `priceBeforeCredits`. Mobile-IAP mode read the asset's own `price` field instead, which carries no unit — on a USD-denominated listing it is USD wei — and reconstructed the pre-credits figure by adding credits back to an amount the caller already knew.
- Present a purchase as credits only when credits are the payment source. The mobile-IAP marker rides on the URL and says nothing about what will be spent, so keying the branding off it alone labelled a MANA purchase as credits on any flow that does not set the flag.
- Pick the mobile-IAP action from the payment source rather than from whichever callback was supplied. `onBuyWithCredits` is passed on the ENS claim path regardless of what the buyer selected, so the old check settled through credits even when they chose MANA. This matches what `renderBuyNowButton` already did.
- Withhold the mobile-IAP confirmation for an item when the credits selection did not arrive, rather than falling through to MANA.

## Test plan

- [x] `shopRedirect` covered for every case variant of the basename, for a longer first segment that is not the basename, and for trailing slashes
- [x] The deep-link opening covered for firing once per mount rather than per render, and for carrying the credits selection in and out of mobile-IAP mode
- [x] The shared checkout covered for the branding following the payment source, for the action following it, and for showing the resolved MANA amount rather than the listed figure on a USD-denominated item
- [x] The item checkout covered for withholding the confirmation without the credits selection, and for presenting it once the selection is there
- [x] `npm run build`, `npm run check:code` and `npx jest` (164 suites, 1899 tests)

### DIFF
--- a/webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyNFTButtons.spec.tsx
+++ b/webapp/src/components/AssetPage/SaleActionBox/BuyNFTButtons/BuyNFTButtons.spec.tsx
@@ -1,0 +1,92 @@
+import React from 'react'
+import { screen, waitFor } from '@testing-library/react'
+import { ChainId, NFTCategory, Network } from '@dcl/schemas'
+import { Asset, AssetType } from '../../../../modules/asset/types'
+import { renderWithProviders } from '../../../../utils/test'
+import BuyNFTButtons from './BuyNFTButtons'
+import { Props } from './BuyNFTButtons.types'
+
+const ITEM = {
+  id: 'an-item',
+  itemId: '0',
+  contractAddress: '0xcollection',
+  category: NFTCategory.WEARABLE,
+  network: Network.MATIC,
+  chainId: ChainId.MATIC_MAINNET,
+  price: '100000000000000000000',
+  available: 1,
+  data: {},
+  isOnSale: true
+} as unknown as Asset
+
+// Supplies the loaded asset the real provider would resolve, so the render callback under test runs
+// without the store and network machinery behind it.
+jest.mock('../../../AssetProvider', () => ({
+  AssetProvider: ({ children }: { children: (asset: Asset, order: null) => React.ReactNode }) => children(ITEM, null)
+}))
+
+jest.mock('../UseCreditsToggle', () => ({ __esModule: true, default: () => null }))
+jest.mock('./BuyWithCryptoButton', () => ({ BuyWithCryptoButton: () => <button>buy</button> }))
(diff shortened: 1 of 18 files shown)


juanmahidalgo:
Applied the first finding, with one change of approach: rather than adding the missing `priceBeforeCredits` to `MintNameWithCryptoModal`, I made the prop **required** (`1786a640`). The defect is not that one caller forgot it — it is that it could be omitted silently. Required, the compiler named all three callers at once and a future one cannot leave it out.

Not applying the `isIAP && useCredits !== true` guard, on either modal. Two reasons.

**It no longer corrects anything.** `eb6e6c9f` binds the credits presentation to the payment source (`presentsCredits = isIAP && useCredits === true`). The inconsistency the guard was proposed for — a screen labelled as credits while settling in MANA — cannot occur once the label follows the flag. What remains is a policy choice: refuse to proceed in mobile-IAP mode unless credits are selected.

**And it blocks flows that legitimately settle in MANA.**

- `BuyNftWithCryptoModal` is also opened by `SaleRentActionBox.container.ts:25` and `Actions.container.ts:19`, neither of which passes `useCredits`; it defaults to `false` at `BuyNftWithCryptoModal.tsx:38`. With the guard, LAND/Estate orders and NAME resales are blocked in mobile-IAP mode. That is why it was removed from this modal in `3c7f6556`. The review notes the gap as "currently unreachable (`BuyNFTButtons` initialises `useCredits` to `true` in IAP mode)", which holds for that entry point but not for these two.
- `MintNameWithCryptoModal`: `ClaimNameFatFingerModal.tsx:47` initialises the toggle to `false`, so paying for a NAME in MANA is both supported and the default. The guard would refuse the checkout to anyone who leaves it off.

If defence in depth is still wanted at that layer, the version that does not cost availability is the one already in place: the presentation follows the payment source, so a flow without credits is both shown and charged in MANA.

PR #2716, opened 2026-09-15
fix: bind the estate shown at review to the fingerprint that gets signed

juanmahidalgo:
## Changes

- Read an Estate's composition from the registry once when a bid, purchase, or listing flow opens — its size, its LAND ids, and its `getFingerprintV2`, all pinned to a single block — and recompute the fingerprint from the ids that came back, rejecting a composition that moved mid-read.
- Freeze that reading for the life of the flow, and use its fingerprint for the listing, bid or order, so what is signed describes the LANDs that were on screen. Re-read the fingerprint immediately before submitting and refuse if it no longer matches. Applied to placing a bid, buying, and both listing modals.
- Draw the Estate on those screens from the registry reading rather than from the indexed copy, so the composition shown is the one the action binds to even while the indexer is behind.
- Do not let the user continue while the reading is pending, when it failed, or when the registry holds a different set of LANDs than the page is showing — and say which of those it is, with a retry for a failed read.
- Keep an Estate purchase on its own chain and currency: the cross-chain route settles through a marketplace call that carries no fingerprint, which an Estate transfer requires, so that path cannot complete.
- Read the LAND ids through Multicall3 where available (the largest Estate on mainnet holds 9,639), falling back to individual reads otherwise.

## Test plan

- [x] `readEstateSnapshot`, `decodeLandTokenId`, `computeEstateFingerprint`, `isSameEstateComposition` and `applyEstateSnapshot` covered against a real mainnet Estate's land ids, coordinates and fingerprint, plus the four coordinate quadrants; both read paths, the single-block pinning, the multi-batch multicall, and rejection of ids that do not add up to the fingerprint
- [x] `useEstateSnapshot` covered for ready / pending / unreadable / different-composition, that it withholds the fingerprint in every state but ready, and that it does not re-read when the store hands over a new object for the same Estate
- [x] The bid and buy flows covered end to end: the confirmation is withheld while the composition is blocking, `confirm()` runs before signing, and the frozen fingerprint is what gets sent
- [x] `npm run build`, `npm run check:code` and `npx jest` (169 suites, 1964 tests)


juanmahidalgo:
Both P1s fixed in `23b5cdb5`.

**Legacy credits path dropping the fingerprint** — `useCreditsLegacyMarketplace` settles through `executeOrder`, which carries no fingerprint, so it cannot bind an Estate's composition. The legacy branch now refuses credits for an Estate rather than settling unbound; the fingerprint-carrying `safeExecuteOrder` path is the only Estate-safe legacy route. (In the UI this path is currently unreachable — the credits toggle is gated to Polygon assets and Estates are on Ethereum L1 — so this is a fail-closed backstop for that code path.)

**Superseded reads publishing a stale snapshot** — added a per-read generation counter to `useEstateSnapshot`. A resolved read now publishes only when its generation is still the latest, so on A → B → A the earlier A read can no longer overwrite the newer one. Covered with a deferred-promise A → B → A test that resolves the later read first and asserts the fresh snapshot survives.

Both changes have a negative-control check confirming the regression test fails without the fix. `npm run build`, `npm run check:code` and `npx jest` (170 suites, 1975 tests) pass.

